Issue #55, opened 2022-06-27
rename all options to be `camelCase` and not `snake_case`


joshspicer:
To follow with JSON naming conventions, we should make all the options in the `devcontainer-features.json` camelCase.

Right now, options are converted to environment variables with [this logic in the CLI](https://github.com/devcontainers/cli/blob/main/src/spec-node/containerFeatures.ts#L25-L28).  

["v2" features]() require environment variables [of this form](https://github.com/devcontainers/features/blob/main/src/docker-in-docker/install.sh#L11-L16), which will get emitted by the tooling.  The tranformation linked above converts what the user types in their `devcontainer-feature.json` to the env variable that is actually emitted by the tooling.

One solution is to sanitize and transform camelcase to CAPITAL_SNAKE_CASE at the capitalization boundaries.

`dockerDashComposeVersion` -> `DOCKER_DASH_COMPOSE_VERSION`.

This change alone would may not automatically be backwards compatible, thus we'll have to be careful with how it is implemented.

Chuxel:
This appears to be fixed, Should we close the issue?

joshspicer:
@samruddhikhandale I think you resolved all these - but please reopen if that's not the case.

PR #80, opened 2022-08-04
Add new NVIDIA CUDA feature

jungaretti:
Adds a new feature that installs Nvidia's CUDA and cuDNN libraries. These libraries enable tools such as TensorFlow to leverage an Nvidia GPU to perform highly-parallelizable tasks.

### Motivation

This new feature would dramatically decrease friction for developers who use Codespaces to work on machine learning projects. Instead of copying, pasting, and executing [this script](https://gist.github.com/jungaretti/d19410cb8216aea94ce570f5bc310217#file-install-cuda-libraries-oneliner-sh), developers could simply add the new `nvidia` feature to their `devcontainer.json` and rebuild their codespace.

### Changes

- Added a new Nvidia feature
- Added a test for the new feature
- Added two scenario tests for the new feature


src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "nvidia-cuda",
+  "name": "NVIDIA CUDA",
+  "description": "Installs NVIDIA CUDA libraries.",
+  "version": "1.0.0",
+  "options": {
+    "install_cuda": {
```
jkeech, line 7: What's the scenario where you wouldn't want to install the base CUDA libraries? Do we need to make this an option or should it just always happen?

src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -0,0 +1,18 @@
...
+      "type": "boolean",
+      "default": true,
+      "description": "Install essential CUDA libraries"
+    },
+    "install_cudnn": {
```
jkeech, line 12: Given the discussion on https://github.com/devcontainers/features/issues/55, can we switch these options to use camelCase? We need to handle back-compat for the existing features, but since this is a new feature, we could just start with camelCase from the beginning.

src/nvidia-cuda/install.sh
```diff
@@ -0,0 +1,36 @@
...
+install_cudnn=${INSTALL_CUDNN:-"true"}
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
+# Download NVIDIA's keyring package
+ubuntu_version="$(lsb_release -sr | sed 's/\.//g')"
+keyring_repo="https://developer.download.nvidia.com/compute/cuda/repos/ubuntu$ubuntu_version/x86_64"
```
jkeech, line 15: Do we need to handle different architectures for ARM or anything else?

src/nvidia-cuda/install.sh
```diff
@@ -0,0 +1,36 @@
...
+keyring_package_file="$keyring_package_path/$keyring_package"
+wget -O "$keyring_package_file" "$keyring_package_url"
+
+# Install NVIDIA's keyring package
+apt-get install -yq "$keyring_package_file"
+apt-get update -yq
+
+if [ "$install_cuda" = "true" ]; then
+    echo "Installing CUDA libraries..."
+    apt-get install -yq cuda-libraries-11-7
```
jkeech, line 28: Do we need to expose the version of these libraries as an option for users to pick?

src/nvidia-cuda/devcontainer-feature.json
jungaretti: I can't think of a scenario, as this is the point of this feature 😄 I'll remove the option

src/nvidia-cuda/devcontainer-feature.json
jungaretti: Absolutely

Chuxel:
There's a bit of a gap in the dev container spec for features that is needed here.  Specifically, there's no way that I'm aware of for a feature to specify GPU usage - which I'd expect you'd want for this feature, correct?  Is there any reason to use this feature without it?

src/nvidia-cuda/install.sh
Chuxel: Features here should work on both Ubuntu and Debian - most bases are actually debian rather than ubuntu in the wild.  Does the apt repo need to be different?

I'd also recommend doing `. /etc/release` and referencing $ID and $VERSION_COODENAME here to avoid the dependency on `lsb_release`.  That's not in most images by default.

src/nvidia-cuda/install.sh
```diff
@@ -0,0 +1,36 @@
...
+set -e
+
+install_cuda=${INSTALL_CUDA:-"true"}
+install_cudnn=${INSTALL_CUDNN:-"true"}
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
```
Chuxel, line 12: You are missing logic to check for the commands you are using and then installing them if not present.  It looks like you'd need `wget` and `apt-transport-https` at a minimum which may not be in base images. You may also need `gnupg2`. 

Try using this feature with the straight `ubuntu` and `debian` images to verify what you can count on being present.

Codespces' kitchen sink is massive partly because all of this is already there.

src/nvidia-cuda/install.sh
Chuxel: Is this alone enough? 

e.g., this is how Nvidia's base image installs: https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.7.0/ubuntu2004/runtime/Dockerfile#L34

It also installs `libcublas-11-7`, `cuda-nvtx-11-7`, and `libcusparse-11-7`.  I ask since I've used the Dockerfile logic above myself so I know that works.

src/nvidia/devcontainer-feature.json
```diff
@@ -0,0 +1,17 @@
+{
+  "id": "nvidia",
```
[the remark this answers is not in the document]
Chuxel: Yeah, +1 on this being a gap we should address. This feature won't work at all without some way to set that flag.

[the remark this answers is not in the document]
jkeech: The workaround today for end users would be to add the following to their devcontainer:

```
"runArgs": ["--gpus", "all"],
"features": {
    "ghcr.io/devcontainers/features/nvidia-cuda:1": {}
}
```

Ideally the feature itself could contribute the `--gpus all` run args so that the end user doesn't need to include it in the devcontainer.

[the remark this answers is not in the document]
Chuxel: Raised https://github.com/devcontainers/spec/issues/82

jungaretti:
> Is there any reason to use this feature without it?

Not really. These libraries enable tools to leverage NVIDIA GPUs. That said, you _can_ successfully install this feature without a GPU, but it won't be too useful.

src/nvidia-cuda/install.sh
jungaretti: Thanks! I added the `apt_get_update_if_needed` and `check_packages` functions from the Python feature. Adding `check_packages wget ca-certificates` fixes things on `ubuntu` and `debian`.

src/nvidia-cuda/install.sh
jungaretti: Thanks, done!

src/nvidia-cuda/install.sh
jungaretti: `cuda-libraries-11-7` depends on a handful of library packages including `libcublas-11-7` and `libcusparse-11-7`, and the test I added specifically checks for these libraries: https://github.com/devcontainers/features/blob/62bcebaeabae49ebab62672747d4f1677399b4f3/test/nvidia-cuda/test.sh#L9-L15

Screen Shot 2022-08-15 at 3 34 27 PM

jungaretti: We could add an option for NVTX, but I don't think it's needed by default: https://docs.nvidia.com/nsight-visual-studio-edition/nvtx/index.html

jungaretti:
It looks like NVIDIA only releases cuDNN for Ubuntu. It's nowhere to be found in their Debian repo: https://developer.download.nvidia.com/compute/cuda/repos/debian11/x86_64/

This means that the feature doesn't work on Debian when `installCudnn` is true. Do we have a precedent for situations like this?

src/nvidia-cuda/install.sh
jungaretti: I added an option for NVTX

src/nvidia/devcontainer-feature.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "Nvidia",
+  "description": "Installs Nvidia's CUDA libraries.",
+  "options": {
+    "install_cuda": {
+      "type": "boolean",
+      "default": true,
+      "description": "Install CUDA libraries"
```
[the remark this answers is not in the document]
jungaretti: We're only installing the libraries that software needs to use CUDA, not the whole toolkit. The toolkit is much heavier and takes longer to install:
- `cuda-libraries-11-7`: 57s to install, 1.6 GB added
- `cuda-toolkit-11-7`: 246s to install, 6.2 GB added

According to NVIDIA, `cuda-libraries-*` is a meta package that includes "all available library [CUDA] runtime packages."

We could add an option to install the full toolkit, but it includes many graphics-oriented tools such as [Nsight graphics](https://developer.nvidia.com/nsight-graphics).

Chuxel:
> It looks like NVIDIA only releases cuDNN for Ubuntu. It's nowhere to be found in their Debian repo: https://developer.download.nvidia.com/compute/cuda/repos/debian11/x86_64/
> 
> This means that the feature doesn't work on Debian when `installCudnn` is true. Do we have a precedent for situations like this?

We do for other distributions - generally the model is to detect the situation, provide a good error, and then document. 

That said, that's super odd considering Debian is called out in the docs. https://docs.nvidia.com/deeplearning/cudnn/install-guide/index.html#installlinux-deb.  Wonder if the fallback should just be to curl the deb in this case.

jungaretti:
This PR is ready for another set of eyes! Here are my notable changes:

- **Added more helpful error messages when package installation fails.** This feature prints an error message with the failed package (for example, `libcudnn8=8.5.0.96-1+cuda11.7`) and a link to NVIDIA's package repo.
- **Added support for installing a certain version of CUDA and cuDNN.** Unfortunately, NVIDIA's package names include their minor version number (for example, `cuda-libraries-11-7`). I've hardcoded the latest CUDA and cuDNN versions, but I'm open to more elegant solutions. This problem has existed since my first commit.
- **Pivoted to always using NVIDIA's Ubuntu 20.04 package repo.** I've found a lot of inconsistency among NVIDIA's Debian and Ubuntu repos. The Ubuntu 20.04 packages work perfectly on every image I've tested (Debian 10, Debian 11, Ubuntu 18, Ubuntu 20, Ubuntu 22). **Another approach here would be to fail package installation with the more helpful error message.**

Failed to install cuda-libraries-11-9. See https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64 for all available packages and versions.

src/nvidia-cuda/install.sh
```diff
@@ -0,0 +1,75 @@
...
+# Function to run apt-get if needed
+apt_get_update_if_needed() {
+    if [ ! -d "/var/lib/apt/lists" ] || [ "$(ls /var/lib/apt/lists/ | wc -l)" = "0" ]; then
+        echo "Running apt-get update..."
+        apt-get update
```
joshspicer, line 25: We've actually had a few issues recently (https://github.com/devcontainers/features/pull/98) and are removing this optimization until we're confident it doesn't cause breaks for people.

joshspicer (dismissed):

src/nvidia-cuda/install.sh
```diff
@@ -0,0 +1,75 @@
...
+set -e
+
+INSTALL_CUDNN=${INSTALLCUDNN:-"false"}
+INSTALL_NVTX=${INSTALLNVTX:-"false"}
+CUDA_VERSION=${VERSION:-"latest"}
+CUDNN_VERSION=${CUDNNVERSION:-"latest"}
+
+# NVIDIA's package names include this information
+LATEST_CUDA_VERSION="11.7"
+LATEST_CUDNN_VERSION="8.5.0.96-1"
```
joshspicer, line 12: Discussed offline about moving these values to the `devcontainer-feature.json`. 🚀

joshspicer (approved):

PR #89, opened 2022-08-11
Options: Follow camelCase

samruddhikhandale:
- Options are renamed to follow `camelCase`
- Options match with https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/container-features/src/devcontainer-features.json


src/desktop-lite/install.sh
```diff
@@ -7,10 +7,10 @@
...
-NOVNC_VERSION=${NOVNC_VERSION:-"1.2.0"} # TODO: Add in a 'latest' auto-detect and swap name to 'version'
-VNC_PASSWORD=${VNC_PASSWORD:-"vscode"}
-NOVNC_PORT="${NOVNC_PORT:-6080}"
-VNC_PORT="${VNC_PORT:-5901}"
+NOVNC_VERSION=${NOVNCVERSION:-"1.2.0"} # TODO: Add in a 'latest' auto-detect and swap name to 'version'
+VNC_PASSWORD=${PASSWORD:-"vscode"}
+NOVNC_PORT="${WEBPORT:-6080}"
+VNC_PORT="${VNCPORT:-5901}"
 
 INSTALL_NOVNC=${INSTALL_NOVNC:-"true"}
```
samruddhikhandale, line 15: > INSTALL_NOVNC=${INSTALL_NOVNC:-"true"}

@joshspicer `installNoVnc` is not exposed as one of the `options`. Then, what is the purpose of `${INSTALL_NOVNC:-"true"}` ? 

eg. We have `additionalVersions` which is not exposed in `options`, but that's needed by `images` repo. 
https://github.com/devcontainers/images/blob/main/src/codespaces/.devcontainer.json#L15

So I did convert it to follow camelCase in [here](https://github.com/devcontainers/features/blob/c5877b0138be1c0cf81fbcd1e236a914acbbabbe/src/dotnet/install.sh#L31). Hence, wondering if we'd like to convert all the `${INSTALL_NOVNC:-"true"}` to `${INSTALLNOVNC:-"true"}`

joshspicer: > installNoVnc

That was probably an existing option from when the script was in vscode-dev-containers, and just never exposed as a 'feature option'.  

I suppose for consistency it would be nice for everything to follow the same pattern, although if it isn't relevant to the back-compat story I don't have a strong opinion.

joshspicer (approved):
👏

jungaretti (approved):

src/desktop-lite/install.sh
samruddhikhandale: As it's not relevant to back-compat, will target this issue for future.

PR #91, opened 2022-08-11
update `dotnet`/`oryx` features to use recommended install location for dotnet6

joshspicer:
Update oryx feature to auto-install dotnet 6 if needed for compilation.  Uses the latest [guidance to install dotnet6 on jammy (as of yesterday)](https://github.com/dotnet/core/issues/7699)

Additionally,

- Adds the devcontainer/cli to install automatically as a postCreateCommand trigger
- Test CI will ignore automated documentation updates


src/dotnet/install.sh
```diff
@@ -230,6 +230,21 @@ install_using_apt() {
     fi
 }
 
+install_using_default_apt_repo() {    
+    apt_get_update_if_needed  
+    DOTNET_PACKAGE="dotnet6"
+
+    if [ "${DOTNET_VERSION}" = "latest" ] || [ "${DOTNET_VERSION}" = "lts" ] || "${DOTNET_VERSION}" = "6.0"; then
```
samruddhikhandale, line 237: > "${DOTNET_VERSION}" = "6.0"

Should we instead check for the major version (`6`)?

joshspicer: I was mirroring what we had in the `devcontainer-feature.json`, but I agree it makes more sense to present major versions, so changed that.

samruddhikhandale (approved):

PR #96, opened 2022-08-15
sshd: always run initial apt-get

### LINKED ISSUE #94: `sshd` fails to install intermittently
From an internal report:

>We recently add `"features": { "sshd": "latest" }` to our devcontainer and have since run into periodic build failures due its related packages not being found. It looks like the sshd_debian.sh script has some logic to [conditionally skip apt-update](https://github.com/microsoft/vscode-dev-containers/blob/faae2d13a166cf51473f2613fada78e8f9e85d0c/script-library/sshd-debian.sh#L45-L53), which I think may be the issue.

Logs:
```
#12 [stage-0 2/3] COPY --from=dev_containers_feature_content_source . /tmp/build-features/
#12 DONE 0.2s
#13 [stage-0 3/3] RUN cd /tmp/build-features/sshd_1 && chmod +x ./install.sh && ./install.sh
#13 1.211 (*) Script: sshd-debian.sh
#13 1.270 Skipping apt-get update.
#13 1.326 Reading package lists...
#13 1.964 Building dependency tree...
#13 2.078 Reading state information...
#13 2.238 openssh-client is already the newest version (1:8.4p1-5).
#13 2.238 The following packages were automatically installed and are no longer required:
#13 2.239   libconfig-inifiles-perl libdbd-mariadb-perl libdbi-perl libmariadb3
#13 2.239   libterm-readkey-perl mariadb-common
#13 2.239 Use 'apt autoremove' to remove them.
#13 2.239 The following additional packages will be installed:
#13 2.240   libwrap0 openssh-sftp-server runit-helper
#13 2.240 Suggested packages:
#13 2.240   molly-guard monkeysphere ssh-askpass ufw
#13 2.240 Recommended packages:
#13 2.240   default-logind | logind | libpam-systemd ncurses-term xauth
#13 2.343 The following NEW packages will be installed:
#13 2.344   libwrap0 lsof openssh-server openssh-sftp-server runit-helper
#13 2.408 0 upgraded, 5 newly installed, 0 to remove and 8 not upgraded.
#13 2.408 Need to get 824 kB of archives.
#13 2.408 After this operation, 2339 kB of additional disk space will be used.
#13 2.408 Get:1 http://deb.debian.org/debian bullseye/main amd64 lsof amd64 4.93.2+dfsg-1.1 [319 kB]
#13 2.431 Get:2 http://deb.debian.org/debian bullseye/main amd64 libwrap0 amd64 7.6.q-31 [59.0 kB]
#13 2.625 Err:3 http://deb.debian.org/debian bullseye/main amd64 openssh-sftp-server amd64 1:8.4p1-5
#13 2.625   404  Not Found [IP: 146.75.30.132 80]
#13 2.625 Get:4 http://deb.debian.org/debian bullseye/main amd64 runit-helper all 2.10.3 [7808 B]
#13 2.626 Get:5 http://deb.debian.org/debian bullseye/main amd64 openssh-server amd64 1:8.4p1-5 [385 kB]
#13 2.651 E: Failed to fetch http://deb.debian.org/debian/pool/main/o/openssh/openssh-sftp-server_8.4p1-5_amd64.deb  404  Not Found [IP: 146.75.30.132 80]
#13 2.651 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
#13 2.651 Fetched 771 kB in 0s (2622 kB/s)
Error: Process completed with exit code 1.
```

joshspicer:
closes #94

### DIFF
--- a/src/sshd/install.sh
+++ b/src/sshd/install.sh
@@ -50,8 +50,6 @@ check_packages() {
 # Ensure apt is in non-interactive to avoid prompts
 export DEBIAN_FRONTEND=noninteractive
 
-
-
 # Install openssh-server openssh-client
 check_packages openssh-server openssh-client lsof
 
--- a/src/sshd/install.sh
+++ b/src/sshd/install.sh
@@ -44,7 +44,7 @@ apt_get_update_if_needed()
 {
     if [ ! -d "/var/lib/apt/lists" ] || [ "$(ls /var/lib/apt/lists/ | wc -l)" = "0" ]; then
         echo "Running apt-get update..."
-        apt-get update
+        apt-get update -y
     else
         echo "Skipping apt-get update."
     fi
@@ -60,6 +60,9 @@ check_packages() {
 
 # Ensure apt is in non-interactive to avoid prompts
 export DEBIAN_FRONTEND=noninteractive
+
+# Initial apt get
+apt-get update -y
 
 # Install openssh-server openssh-client
 check_packages openssh-server openssh-client lsof
--- a/src/sshd/install.sh
+++ b/src/sshd/install.sh
@@ -39,21 +39,9 @@ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then
     USERNAME=root
 fi
 
-# Function to run apt-get if needed
-apt_get_update_if_needed()
-{
-    if [ ! -d "/var/lib/apt/lists" ] || [ "$(ls /var/lib/apt/lists/ | wc -l)" = "0" ]; then
-        echo "Running apt-get update..."
-        apt-get update -y
-    else
-        echo "Skipping apt-get update."
-    fi
-}
-
 # Checks if packages are installed and installs them if not
 check_packages() {
     if ! dpkg -s "$@" > /dev/null 2>&1; then
-        apt_get_update_if_needed
         apt-get -y install --no-install-recommends "$@"
     fi
 }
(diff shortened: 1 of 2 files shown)


src/sshd/install.sh
```diff
@@ -44,7 +44,7 @@ apt_get_update_if_needed()
 {
     if [ ! -d "/var/lib/apt/lists" ] || [ "$(ls /var/lib/apt/lists/ | wc -l)" = "0" ]; then
         echo "Running apt-get update..."
-        apt-get update
+        apt-get update -y
```
jkeech, line 47: Instead of running the check twice (initial one and here), can we just remove this function entirely and replace `apt_get_update_if_needed` to be `apt-get update -y` on line 56? That would just remove the conditional part of the update and always run it once.

Do you know the reason why we previously had this behind a conditional? I want to make sure we aren't breaking anything by always running this. Perhaps it was just meant to be a performance optimization, in which case, I think it's fine to always run to prefer correctness over performance.

jkeech: Another reason to move all of this to line 56 is to skip the update if we aren't going to install any packages. The root cause of this bug seems to be that we were skipping the update on line 56 and then trying to install packages on line 57. But we only want to run the update and install if we don't have those packages installed already. So we shouldn't run `apt update` until we know we need to install the missing packages.

joshspicer: The cached apt lists are what tell apt what to install, so unless we update them first, apt is not going to know whether or not the packages need to be updated.  

I suppose this could check if _any_ version of `openssh-server openssh-client lsof`, but I think it is just easier to let apt figure that out for us and install the latest compatible versions of each

jkeech (approved):

twisted54 (commented):
src/sshd/install.sh

Issue #92, opened 2022-08-11
Extensions and settings for vscode in old locations


Chuxel:
It looks like the extensions and settings for VS Code are at the top level rather than under `customizations.vscode`. 

Running...

```bash
devcontainer read-configuration --include-features-configuration --log-format json --workspace-folder .
```

...does not show the extensions or settings, which implies they may not be getting picked up.

jkeech:
@samruddhikhandale, can you move all of the `settings` and `extensions` under `customizations.vscode`? Example: https://github.com/devcontainers/features/blob/main/src/java/devcontainer-feature.json#L31-L33

samruddhikhandale:
> Running...
`devcontainer read-configuration --include-features-configuration --log-format json --workspace-folder` .
...does not show the extensions or settings, which implies they may not be getting picked up.

Fix for this 🐛  - https://github.com/devcontainers/cli/pull/126

> can you move all of the settings and extensions under customizations.vscode?
https://github.com/devcontainers/features/pull/97

samruddhikhandale:
Released features with the changes in https://github.com/devcontainers/features/pull/97 🚀 
https://github.com/devcontainers/features/runs/7886497114?check_suite_focus=true

Closing as fixed.

PR #97, opened 2022-08-16
Move extensions and settings under customizations.vscode

samruddhikhandale:
Addresses - https://github.com/devcontainers/features/issues/92#issuecomment-1212687080


joshspicer (approved):
Thank you!  We'll also probably want to bump versions

samruddhikhandale:
@joshspicer For bumping versions, are we following `bump version (according to semver) if you change something in the feature` logic? 

I am worried if that won't always work. For example, for [ruby](https://github.com/devcontainers/features/pkgs/container/features%2Fruby) we currently have `1.0.0` as `latest` but [this PR](https://github.com/devcontainers/features/pull/95/files) did bump it up to `1.0.1` so ideally I shouldn't bump ruby version in my PR, correct? but what if there's a race condition in my PR merging and features publishing? 

I wonder if we should always create a separate PR which bumps versions before every release, that way we won't have inconsistencies. (We could decide on PR title which will make it simpler to decide whether or not to bump version of a particular feature by looking at the commit history.)

joshspicer:
I think `main` should always be the source of truth, and when we merge a feature with an updated version, the action should do the proper publishing and retagging.  I think this would solve any problems related to versioning.  We haven't discussed release cadence, but I don't see why a feature in main shouldn't be published.

We may want the `devcontainer feature publish` command to refuse to publish an older breakfix version than what it has already seen, maybe that will also help mitigate any problems you foresee?

joshspicer:
Perhaps we have the github action run on PRs and scan all features in `main`, and fail if there is an inconsistency in the versioning.

samruddhikhandale:
> We haven't discussed release cadence, but I don't see why a feature in main shouldn't be published.

I somehow don’t like the idea of publishing features as soon as the PR merges, but this could be just me.

Cons

- The PR changes will be immediately available to the public, it seems scary and having some gap might help catch more bugs
- We would have plenty of new patch versions, as we will end up creating new releases which might not even be user visible (eg - doc updates, nits, code shuffling)

samruddhikhandale:
> We may want the devcontainer feature publish command to refuse to publish an older breakfix version than what it has already seen, maybe that will also help mitigate any problems you foresee?

But we'd like to support that scenario to be able to publish a hotfix for minor/patch scenario. [example](https://github.com/devcontainers/cli/blob/07bc66f10a26fcc1648ea9ead27a48e695d93f0c/src/test/container-features/featuresCLICommands.test.ts#L128-L144)

samruddhikhandale:
> Perhaps we have the github action run on PRs and scan all features in main, and fail if there is an inconsistency in the versioning.
Liked this 💡

samruddhikhandale:
> We should make sure that the PR can only be merged when it is in sync with main, this will make sure that the action checks the right contents.

image

Enabling this branch protection policy should help.

samruddhikhandale:
Discussed the version issue offline, @joshspicer  could you help re-approve? 
Bumped versions that are needed.

joshspicer (approved):

PR #103, opened 2022-08-17
Skip CI on Documentation PRs

joshspicer:
Attempts to use https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/ to not run tests on Documentation changes, as this is wasteful and takes times for no reason.


.github/workflows/test-all.yaml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, 'no-ci') && !contains(github.event.head_commit.message, 'CI ignore') && !contains(github.event.head_commit.message, 'Automated documentation update')"
+    if: "!contains(github.event.head_commit.message, 'no-ci') && !contains(github.event.head_commit.message, 'CI ignore')"
```
samruddhikhandale, line 11: Curious, do we need this condition now? 
I believe we can achieve the same thing if we use keywords listed in https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

.github/workflows/test-pr.yaml
```diff
@@ -5,7 +5,7 @@ on:
 jobs:
   detect-changes:
     runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, 'no-ci') && !contains(github.event.head_commit.message, 'CI ignore') && !contains(github.event.head_commit.message, 'Automated documentation update')"
+    if: "!contains(github.event.head_commit.message, 'no-ci') && !contains(github.event.head_commit.message, 'CI ignore')"
```
samruddhikhandale, line 8: Same here and for other workflow files as well :)

.github/workflows/test-all.yaml
joshspicer: yea good point. Reading online I am not sure if this is going to work yet which is why I kept these around :D

samruddhikhandale (approved):

Issue #110, opened 2022-08-22
Error installing os-provided python


chrmarti:
@samruddhikhandale @joshspicer 

`find` passes its results on to `xargs` in a way that breaks when the paths contain spaces:
https://github.com/devcontainers/features/blob/ec1deb82808c14de9ce601838e255357718b49d4/src/python/install.sh#L391

I see this fail using the following with 0.248.0:
```json
{
    "image": "mcr.microsoft.com/vscode/devcontainers/base:0-bullseye",
    "features": {
       "node": "latest",
       "python": "os-provided"
    } 
}
```

https://unix.stackexchange.com/questions/172481/how-to-quote-arguments-with-xargs suggests using `find` with `-print0` and `xargs` with `-0` to use `\0` as the delimiter. (This seems to work for our case too based on a quick try.)

Similarly:
https://github.com/devcontainers/features/blob/ec1deb82808c14de9ce601838e255357718b49d4/src/python/install.sh#L411

```
[12 ms] Remote-Containers 0.248.0 in VS Code 1.71.0-insider (235a92a3693176fbf9af16ec5b829c458a1d682f).
[...]
#0 11.32 Skipping apt-get update.
#0 11.33 Reading package lists...
#0 11.60 Building dependency tree...
#0 11.66 Reading state information...
#0 11.73 python3 is already the newest version (3.9.2-3).
#0 11.73 python3 set to manually installed.
#0 11.73 The following additional packages will be installed:
#0 11.73   blt libjs-jquery libjs-sphinxdoc libjs-underscore libpython3-dev
#0 11.73   libpython3.9 libpython3.9-dev python-pip-whl python3-distutils
#0 11.73   python3-lib2to3 python3-pkg-resources python3-setuptools python3-wheel
#0 11.73   python3.9-dev python3.9-doc python3.9-venv tk8.6-blt2.5
#0 11.74 Suggested packages:
#0 11.74   blt-demo python3-examples python-setuptools-doc tix python3-tk-dbg
#0 11.74 Recommended packages:
#0 11.74   javascript-common build-essential
#0 11.87 The following NEW packages will be installed:
#0 11.87   blt libjs-jquery libjs-sphinxdoc libjs-underscore libpython3-dev
#0 11.87   libpython3.9 libpython3.9-dev python-pip-whl python3-dev python3-distutils
#0 11.87   python3-doc python3-lib2to3 python3-pip python3-pkg-resources
[...]

PR #112, opened 2022-08-22
Fix bug - Python: Error installing os-provided python 

samruddhikhandale:
Addresses - https://github.com/devcontainers/features/issues/110

- Fixes bug
- Added test!


src/python/install.sh
```diff
@@ -386,11 +390,14 @@ if [ "${PYTHON_VERSION}" != "none" ]; then
         IFS=$OLDIFS
     fi
 
-    chown -R "${USERNAME}:python" "${PYTHON_INSTALL_PATH}"
-    chmod -R g+r+w "${PYTHON_INSTALL_PATH}"
-    find "${PYTHON_INSTALL_PATH}" -type d | xargs -n 1 chmod g+s
+    if [ ${PYTHON_VERSION} != "os-provided" ] && [ ${PYTHON_VERSION} != "system" ]; then
+        updaterc "if [[ \"\${PATH}\" != *\"${CURRENT_PATH}/bin\"* ]]; then export PATH=${CURRENT_PATH}/bin:\${PATH}; fi"
```
joshspicer, line 394: We likely don't need to `updaterc` (and should update other scripts that use it) if the scripts are used as features, since the `devcontainer-feature.json` already defines these as `containerEnv`.

We need to make sure we are standardizing our scripts to not use be setting variables like $PATH  in multiple ways

joshspicer: Can be handled in another PR since that is irrelevant to this change

joshspicer (approved):

joshspicer (approved):

PR #114, opened 2022-08-23
test updates to adopt https://github.com/devcontainers/cli/pull/124

joshspicer:


.github/workflows/test-all.yaml
```diff
@@ -37,13 +37,73 @@ jobs:
             "rust",
             "sshd",
             "terraform",
+        ]
+        baseImage:
+          [
+            "ubuntu:focal",
+            "ubuntu:jammy",
+            "debian:11",
+            "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",
```
samruddhikhandale, line 46:
```suggestion
            "mcr.microsoft.com/devcontainers/base:ubuntu",
```

.github/workflows/test-all.yaml
```diff
@@ -37,13 +37,73 @@ jobs:
...
+            "ruby",
+            "rust",
+            "sshd",
+            "terraform",
+        ]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: "Install latest devcontainer CLI"
+        run: npm install -g @devcontainers/cli
```
samruddhikhandale, line 94: Wondering if we could install `@devcontainers/cli` only once at some top level which can then be consumed by all jobs.

.github/workflows/test-pr.yaml
```diff
@@ -1,10 +1,11 @@
-name: "PR -Test Updated Features"
+name: "PR - Test Updated Features"
 on:
   pull_request:
 
 jobs:
   detect-changes:
     runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'no-ci') && !contains(github.event.head_commit.message, 'CI ignore')"
```
samruddhikhandale, line 8: Wondering, why have we added this again ? 
Can't we use `[skip ci]` instead?

.github/workflows/test-pr.yaml
```diff
@@ -51,13 +52,29 @@ jobs:
             "ubuntu:jammy",
             "debian:11",
             "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",
```
samruddhikhandale, line 54:
```suggestion
            "mcr.microsoft.com/devcontainers/base:ubuntu",
```

We do have this image --> https://mcr.microsoft.com/v2/devcontainers/base/tags/list

src/python/install.sh
```diff
@@ -10,7 +10,7 @@
 PYTHON_VERSION=${VERSION:-"latest"} # 'system' checks the base image first, else installs 'latest'
 INSTALL_PYTHON_TOOLS=${INSTALLTOOLS:-"true"}
 OPTIMIZE_BUILD_FROM_SOURCE=${OPTIMIZE:-"false"}
-PYTHON_INSTALL_PATH=${INSTALL_PATH:-"/usr/local/python"}
+PYTHON_INSTALL_PATH=${INSTALLPATH}
```
samruddhikhandale, line 13: Is this same as 👇  ? 

> PYTHON_INSTALL_PATH=${INSTALLPATH:-"/usr/local/python"}

test/python/scenarios.json
```diff
@@ -0,0 +1,24 @@
...
+        "image": "ubuntu:focal",
+        "features": {
+            "python": {
+                "version": "3.10.5",
+                "additionalVersions": "3.8,3.9.13"
+            }
+        }
+    },
+    "install_jupyterlab": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
```
samruddhikhandale, line 12:
```suggestion
        "image": "mcr.microsoft.com/devcontainers/base:focal",
```

src/python/install.sh
joshspicer: We get defaults from the `devcontainer-feature.json` (set to `/usr/local/python` by default).  It seems confusing to me that in a lot of the scripts we have two defaults

.github/workflows/test-all.yaml
joshspicer: These run on different vms in parallel. Not sure if there is a "beforeAll().." hook for actions.

test/dotnet/scenarios.json
```diff
@@ -0,0 +1,27 @@
+{
```
samruddhikhandale, line 1: Remove file https://github.com/devcontainers/features/blob/jospicer/features-test-updates/test/dotnet/install_os_provided_python.sh ?

test/python/scenarios.json
```diff
@@ -0,0 +1,31 @@
...
+            "python": {
+                "installJupyterlab": true,
+                "configureJupyterlabAllowOrigin": "*"
+            }
+        }
+    },
+    "install_os_provided_python": {
+        "image": "mcr.microsoft.com/devcontainers/base:0-bullseye",
+        "features": {
+            "node": "latest",
```
samruddhikhandale, line 27: Curious, should this be in `global` tests as this scenario is referencing `node` and `python` Features ?

test/python/scenarios.json
```diff
@@ -0,0 +1,31 @@
...
+                "version": "3.10.5",
+                "additionalVersions": "3.8,3.9.13"
+            }
+        }
+    },
+    "install_jupyterlab": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
+        "remoteUser": "vscode",
+        "features": {
+            "common-utils": {
```
samruddhikhandale, line 15: Curious, should this be in global tests as this scenario is referencing `common-utils` and `python` Features ?

samruddhikhandale (changes requested):

test/python/scenarios.json
joshspicer: IMO it's a judgement call.  If the primary point of the test is testing `python` _with_  another feature, it is cool to be placed under `test/python`. 

Right now the global tests don't run on PRs (might wanna do that tho). I'll leave it here for at least the time being.

test/dotnet/scenarios.json
joshspicer: :D

test/python/scenarios.json
samruddhikhandale: > IMO it's a judgement call. If the primary point of the test is testing python with another feature, it is cool to be placed under test/python.

Good to know 👍

samruddhikhandale (approved):
This change is amazing!!!!!!! 
Thank you! 👏 👏

PR #116, opened 2022-08-24
Oryx - Clean up and shrink size

samruddhikhandale:
With this cleanup, the size of the docker layer for the oryx feature reduced from `2.6 GB` to `205 MB`

Before these changes

image

After

image

### DIFF
--- a/src/oryx/devcontainer-feature.json
+++ b/src/oryx/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.0.1",
+    "version": "1.0.2",
     "name": "Oryx",
     "description": "Installs the oryx CLI",
     "containerEnv": {
--- a/src/oryx/devcontainer-feature.json
+++ b/src/oryx/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.0.2",
+    "version": "1.0.3",
     "name": "Oryx",
     "description": "Installs the oryx CLI",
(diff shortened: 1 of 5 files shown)


src/oryx/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.0.1",
+    "version": "1.0.2",
```
joshspicer, line 3: This is going to have to get bumped one higher with the change I just merged

joshspicer:
Re: the oryx/jammy test missing libssl:

I spent (too much) time debugging this in the `dotnet` feature a month or so ago.  I left some comments in the script that might be useful

https://github.com/devcontainers/features/blob/main/src/dotnet/install.sh#L328

samruddhikhandale:
> Re: the oryx/jammy test missing libssl:
> 

Even with installed `libssl` packages, the `oryx build` and `oryx prep` cmd fails. It looks like a upstream oryx issue. I have asked them about it in chat if they are aware, else I'll create issues on `oryx` and this repo for tracking. 

We are realizing this 🐛  now because I added `oryx prep` in the `test.sh` in this PR. 

image

samruddhikhandale:
>  It looks like a upstream oryx issue. I have asked them about it in chat if they are aware.

Spoke with the team, oryx is only currently supported `ubuntu focal scm` based build image

joshspicer (approved):
Chatted offline.


After adding a warning in the oryx/install.sh noting no support for jammy, this is good to override merge with the failing oryx jammy test.

We can leave it as another action item for https://github.com/devcontainers/cli/issues/139 :)

PR #121, opened 2022-08-31
Anaconda - Clean up and shrink size

samruddhikhandale:
With this cleanup, the size of the docker layer for the Anaconda Feature reduced from **`4.0 GB`** to **`200 MB`**

Cause - Previously it was built from source where `conda/lib` was approx 2 GB. Switched to `apt-get install`

Before these changes

image

After

image

### DIFF
--- a/src/anaconda/install.sh
+++ b/src/anaconda/install.sh
@@ -11,7 +11,7 @@
 VERSION=${VERSION:-"latest"}
 USERNAME=${USERNAME:-"automatic"}
 UPDATE_RC=${UPDATE_RC:-"true"}
-CONDA_DIR=${CONDA_DIR:-"/usr/local/conda"}
+CONDA_DIR="/opt/conda"
 
 set -eux
 export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


src/anaconda/install.sh
```diff
@@ -11,7 +11,7 @@
 VERSION=${VERSION:-"latest"}
 USERNAME=${USERNAME:-"automatic"}
 UPDATE_RC=${UPDATE_RC:-"true"}
-CONDA_DIR=${CONDA_DIR:-"/usr/local/conda"}
+CONDA_DIR="/opt/conda"
```
samruddhikhandale, line 14: Changing directory as apt-get installs conda here by default, would it need backcompat? 

Should we link `"/opt/conda"` to `"/usr/local/conda"` ?

(btw CONDA_DIR was never exposed as option)

joshspicer: I believe this should be fine

joshspicer (approved):

Chuxel:
@samruddhikhandale This is not correct - that just results in "conda" being installed which is the package manager.  Anaconda is the full distro.  Please roll back this update.

PR #125, opened 2022-08-31
Oryx - Fix bug when dynamically installing dotnet in universal image

samruddhikhandale:
Fixes - https://github.com/devcontainers/images/runs/8118223037?check_suite_focus=true

image

- Oryx Feature is now `223 MB` with this change
- Regression from https://github.com/devcontainers/features/pull/116
- Added tests which replicates Oryx's behavior for universal image

### DIFF
--- a/src/oryx/devcontainer-feature.json
+++ b/src/oryx/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.0.4",
+    "version": "1.0.5",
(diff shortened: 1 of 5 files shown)


samruddhikhandale:
FYI - The test failure `PR - Test Updated Features / test (oryx, ubuntu:jammy) (pull_request)` is not new. Details here - https://github.com/devcontainers/features/pull/116#issuecomment-1226270257

PR #127, opened 2022-08-31
Adds a new Conda Feature

samruddhikhandale:
- Adds a new conda feature
- Adds tests


src/conda/NOTES.md
```diff
@@ -0,0 +1,13 @@
+## Using Conda
+
+This Feature includes [the `conda` package manager](https://docs.conda.io/projects/conda/en/latest/index.html) which is a part of the [Anaconda Distribution](https://repo.anaconda.com). Additional packages installed using Conda will be downloaded from Anaconda or another repository if you configure one. To reconfigure Conda in this container to access an alternative repository, please see information on [configuring Conda channels here](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/channels.html ).
```
samruddhikhandale, line 3: This file is exactly similar to [NOTES.md for anaconda](https://github.com/devcontainers/features/blob/main/src/anaconda/NOTES.md).

I have only added in one sentence --> This Feature includes [the `conda` package manager](https://docs.conda.io/projects/conda/en/latest/index.html) `which is a part of the [Anaconda Distribution](https://repo.anaconda.com)`

cc @Chuxel

Chuxel: Yep - as discussed, we can add an option to allow the use of something like conda-forge as well. Different people will have different preferences here, so we just need to make sure that people are aware of the nuances.

edgonmsft (dismissed):
LGTM

src/conda/NOTES.md
samruddhikhandale: Sounds good. Added the option `addCondaForge` 👍

Chuxel (approved):

edgonmsft (approved):
LGTM

eitsupi:
FYI, as an alternative to this for those who prefer to use mamba instead of conda, the following feature is now available.
https://github.com/rocker-org/devcontainer-features/tree/main/src/miniforge

PR #135, opened 2022-09-06
fix bug - Debian / Ubuntu base conda-notice.txt error 

samruddhikhandale:
Addresses - https://github.com/microsoft/vscode-dev-containers/issues/1617#issuecomment-1238177299

Before this change
image

After
image

### DIFF
--- a/src/anaconda/install.sh
+++ b/src/anaconda/install.sh
@@ -104,6 +104,7 @@ if ! conda --version &> /dev/null ; then
 fi
 
 # Display a notice on conda when not running in GitHub Codespaces
+mkdir -p /usr/local/etc/vscode-dev-containers
 cat << 'EOF' > /usr/local/etc/vscode-dev-containers/conda-notice.txt
 When using "conda" from outside of GitHub Codespaces, note the Anaconda repository contains
 restrictions on commercial use that may impact certain organizations. See https://aka.ms/ghcs-conda
@@ -119,6 +120,8 @@ fi
 EOF
 )"
 
-echo "${notice_script}" | tee -a /etc/bash.bashrc >> /etc/zsh/zshrc
+if [ -f "/etc/zsh/zshrc" ]; then
+    echo "${notice_script}" | tee -a /etc/bash.bashrc >> /etc/zsh/zshrc
+fi
 
 echo "Done!"
--- a/src/anaconda/install.sh
+++ b/src/anaconda/install.sh
@@ -103,4 +103,22 @@ if ! conda --version &> /dev/null ; then
     updaterc "export CONDA_DIR=${CONDA_DIR}/bin"
 fi
 
+# Display a notice on conda when not running in GitHub Codespaces
+cat << 'EOF' > /usr/local/etc/vscode-dev-containers/conda-notice.txt
+When using "conda" from outside of GitHub Codespaces, note the Anaconda repository contains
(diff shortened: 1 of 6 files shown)


src/common-utils/install.sh
```diff
@@ -449,15 +449,15 @@ echo -e "\
     ZSH_ALREADY_INSTALLED=${ZSH_ALREADY_INSTALLED}" > "${MARKER_FILE}"
 
 # Display a notice on conda when not running in GitHub Codespaces
-cat << 'EOF' > ${DEV_CONTAINERS_DIR}/conda-notice.txt
+cat << 'EOF' > /usr/local/etc/vscode-dev-containers/conda-notice.txt
```
jkeech, line 452: Can this config live in the `conda`/`anaconda` features, that way it's only added when you include the `conda`/`anaconda` features?

samruddhikhandale: That makes sense. Moved it.

src/anaconda/install.sh
```diff
@@ -103,4 +103,22 @@ if ! conda --version &> /dev/null ; then
     updaterc "export CONDA_DIR=${CONDA_DIR}/bin"
 fi
 
+# Display a notice on conda when not running in GitHub Codespaces
+cat << 'EOF' > /usr/local/etc/vscode-dev-containers/conda-notice.txt
+When using "conda" from outside of GitHub Codespaces, note the Anaconda repository contains
+restrictions on commercial use that may impact certain organizations. See https://aka.ms/ghcs-conda
```
jkeech, line 109: FYI, we should repoint `https://aka.ms/ghcs-conda` to `https://github.com/devcontainers/features/blob/main/src/conda/NOTES.md#using-conda` instead of `https://github.com/microsoft/vscode-dev-containers/tree/main/containers/codespaces-linux#using-conda`

jkeech (dismissed):
Thanks!

src/anaconda/install.sh
samruddhikhandale: @bamurtaugh Could you help change the target URL? I can see you listed as one of the link owners 😄

bamurtaugh: Done! ✅

src/anaconda/install.sh
```diff
@@ -103,4 +103,25 @@ if ! conda --version &> /dev/null ; then
...
+notice_script="$(cat << 'EOF'
+if [ -t 1 ] && [ "${IGNORE_NOTICE}" != "true" ] && [ "${TERM_PROGRAM}" = "vscode" ] && [ "${CODESPACES}" != "true" ] && [ ! -f "$HOME/.config/vscode-dev-containers/conda-notice-already-displayed" ]; then
+    cat "/usr/local/etc/vscode-dev-containers/conda-notice.txt"
+    mkdir -p "$HOME/.config/vscode-dev-containers"
+    ((sleep 10s; touch "$HOME/.config/vscode-dev-containers/conda-notice-already-displayed") &)
+fi
+EOF
+)"
+
+if [ -f "/etc/zsh/zshrc" ]; then
```
jkeech, line 123: Wouldn't this cause the `/etc/bash.bashrc` file to not be written if ZSH isn't installed? It seems like we should split the two out and have one conditional for bash and another conditional for ZSH -- each would only install if the respective shell is installed.

samruddhikhandale: Good catch! Addressed that.

joshspicer (approved):

Issue #132, opened 2022-09-02
github-cli : Fails with signature invalid error


samruddhikhandale:
`github-cli` Feature fails to install with the following error. 

![image](https://user-images.githubusercontent.com/24955913/188232539-e97a8d34-e06f-471c-8c57-e1bff037139e.png)

This error is due to the GPG key used to sign debian packages is expired. Upstream issue --> https://github.com/cli/cli/issues/6175 

Once the upstream GPG key is updated, it also needs to be modfied in this `github-cli` Feature.

samruddhikhandale:
Fixed with https://github.com/devcontainers/features/pull/133

jkeech:
Closing as fixed by #133

PR #139, opened 2022-09-06
Use the newly updated GPG key for GitHub CLI

exilvm:
Fixes the [`apt-key` approach](https://github.com/devcontainers/features/blob/a5fbdab3c096152dd3a318c0c23da9465815aa69/src/github-cli/install.sh#L230) (related to #132) by using the new GPG key: https://github.com/cli/cli/issues/6175#issuecomment-1238477714.


jkeech (dismissed):
Thanks! Do you mind also bumping the patch version in the `devcontainer-feature.json`?

src/github-cli/install.sh
```diff
@@ -10,7 +10,7 @@
 CLI_VERSION=${VERSION:-"latest"}
 INSTALL_DIRECTLY_FROM_GITHUB_RELEASE=${INSTALLDIRECTLYFROMGITHUBRELEASE:-"true"}
 
-GITHUB_CLI_ARCHIVE_GPG_KEY=C99B11DEB97541F0
+GITHUB_CLI_ARCHIVE_GPG_KEY=23F3D4EA75716059
```
samruddhikhandale, line 13: Can we bump the patch version here to 1.0.4? We follow semantic versioning. https://github.com/devcontainers/features/blob/main/src/github-cli/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks for making this change ✨

src/github-cli/install.sh
exilvm: Thanks for letting me know!

jkeech (approved):

samruddhikhandale (approved):
Thanks! 👏

PR #142, opened 2022-09-07
add bzip2 to common-utils (former https://github.com/microsoft/vscode-dev-containers/pull/1586)

lucasfcnunes:


src/common-utils/install.sh
```diff
@@ -85,6 +85,7 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
         rsync \
         ca-certificates \
         unzip \
+        bzip2 \
```
samruddhikhandale, line 88: Can we bump the minor version here ( to 1.1.0 ) as we are adding a new package? (We follow semantic versioning) https://github.com/devcontainers/features/blob/main/src/common-utils/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks! 

image

`bzip2` package will consume approx 196 kB of extra space.

src/common-utils/install.sh
lucasfcnunes: 👍🏽

samruddhikhandale (approved):
Thanks!

PR #145, opened 2022-09-08
fix: jupyterlab not found with docker exec bash -c

samruddhikhandale:
image

### DIFF
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -447,7 +447,7 @@ fi
 
 # Install JupyterLab if needed
 if [ "${INSTALL_JUPYTERLAB}" = "true" ]; then
-    if [[ $(python --version) != "" ]]; then
+    if [[ $(python3 --version) != "" ]]; then
         install_user_package jupyterlab
 
         # Configure JupyterLab if needed
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -315,7 +315,9 @@ sudo_if() {
 
 install_user_package() {
     PACKAGE="$1"
-    sudo_if "$CURRENT_PATH/bin/python" -m pip install --upgrade --no-cache-dir "$PACKAGE"
+    PYTHON_ROOT=$(which python3)
+
+    sudo_if "$PYTHON_ROOT" -m pip install --upgrade --no-cache-dir "$PACKAGE"
 }
 
 add_user_jupyter_config() {
@@ -391,8 +393,9 @@ if [ "${PYTHON_VERSION}" != "none" ]; then
         chmod -R g+r+w "${PYTHON_INSTALL_PATH}"
         find "${PYTHON_INSTALL_PATH}" -type d -print0 | xargs -0 -n 1 chmod g+s
 
-        PATH="${INSTALL_PATH}/bin:${PATH}"
     fi
+
+    export PATH="${CURRENT_PATH}/bin:${PATH}"
 fi
 
 # Install Python tools if needed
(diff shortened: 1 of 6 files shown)


src/python/install.sh
```diff
@@ -315,7 +315,7 @@ sudo_if() {
 
 install_user_package() {
     PACKAGE="$1"
-    sudo_if "$INSTALL_PATH/bin/python3" -m pip install --user --upgrade --no-cache-dir "$PACKAGE"
+    sudo_if "$CURRENT_PATH/bin/python" -m pip install --upgrade --no-cache-dir "$PACKAGE"
```
samruddhikhandale, line 318: image

jungaretti (dismissed):
Great! 💰

src/python/install.sh
jkeech, line 318: How do we ensure `CURRENT_PATH` is set properly when this function executes? It looks like if `"${PYTHON_VERSION}" != "none"`, `CURRENT_PATH` won't be set when jupyterlab tries to install.

If it's not a supported configuration to pass `none` for python version but also to request jupyterlab installation, then we should probably error out above and say the combination of options specified is not valid.

jkeech (commented):
Left one question about combinations of options where the fix might not work. Also, can you explain why we need to remove the `--user` scope on the pip install?

samruddhikhandale:
> Also, can you explain why we need to remove the --user scope on the pip install?

@jkeech  `--user` makes pip install packages in the home directory (ie `/home/$USER/.local/bin` ). This `PATH` is not known when we  run `docker exec $ID bash -c "jupyter-lab --version"`. Hence, it currently fails.

The cli takes care of PATH issues with [containerEnv](https://github.com/devcontainers/features/blob/main/src/python/devcontainer-feature.json#L49). But couldn't find a way to add it there as `/home/$USERNAME/.local/bin` is dynamic and depends on the $USERNAME.

src/python/install.sh
samruddhikhandale: Good point. let me address that

samruddhikhandale: @jkeech as we have removed `--user` now it installs packages within the `python/bin` ☝️ and this path is known during `docker exec bash -c`

Chuxel: Don't use `python` since that can be python 2.x and like on Debian bionic.  Use python3 and pip3 always.

Chuxel: Commented above, but please use `python3` for the command.

Also - be sure to add an export statement to the script to get the version of python that is installed into the PATH.  That won't happen by default.  Furthermore, if you fire this where it is now in the script, it will pick up these env vars: 

https://github.com/devcontainers/features/blob/main/src/python/install.sh#L422-L423

You should `unset` these first.

samruddhikhandale: > Don't use python since that can be python 2.x and is on Debian buster by default along with Ubuntu bionic. Use python3 and pip3 always. If those commands aren't found, you're likely executing against Python 2.x so its a good way to verify - and python3 installs do not always set python either.

We are now linking `python3` to `python` in all scenarios except `none` case. 
Eg - [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L280) but yeah that's a good point, switching it back to `python3`

jungaretti: We are intentionally installing JupyterLab to the user site that users can install anything else with `pip install --user` and not worry about permissions issues. When Jupyter is installed by `root` to a non-user site, we see issues with Jupyter not finding modules.

Chuxel: @samruddhikhandale 

> We are now linking python3 to python in all scenarios except none case.
Eg - [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L280) but yeah that's a good point, switching it back to python3

Yeah, it's the OS version in particular that could be python 2.x.

> We are intentionally installing JupyterLab to the user site that users can install anything else with pip install --user and not worry about permissions issues. When Jupyter is installed by root to a non-user site, we see issues with Jupyter not finding modules.

That likely means it's getting installed using the wrong version of Python... probably the OS version.  If the same version of python is used in both cases, global modules will be found in addition to user ones.

Chuxel:
> image

FYI - These commands would also work if you ran them interactively.  `docker exec -i `...  What is the use case for doing docker exec directly?

samruddhikhandale:
> 
> FYI - These commands would also work if you ran them interactively. `docker exec -i `... What is the use case for doing docker exec directly?
@Chuxel 

@jungaretti would be be able to answer it better, but it is required by codespaces for jupyterlab integration. Codespaces tries to run `docker exec -u codespace <container-id> jupyter server list`  and `docker exec -u codespace <container-id> jupyter lab --version`. 

@jungaretti could we try what Chuck suggested `docker exec -i`, if that works then we don't need to make Feature/universal image changes?

Chuxel:
Yeah...
```
$ docker run -d --privileged --rm mcr.microsoft.com/devcontainers/universal sleep infinity
$ docker exec -it -u codespace <container ID> bash -ic 'jupyter --version'    
```
...works with no errors

Part of the problem here with the universal image is that even the version of python in the path can be dynamic at startup, so there may need to be some special work done in that case verses elsewhere.  Files in the source tree can cause a different version of python to be in the path.

jungaretti:
@samruddhikhandale Can we close this? We've fixed the Python feature and this no longer repros in the Universal image.

```sh
docker run --rm -u codespace mcr.microsoft.com/devcontainers/universal:latest bash -c 'jupyter lab --version'
```
```sh
3.4.7
```

PR #153, opened 2022-09-13
Conda and Java: sync env variables

samruddhikhandale:


src/conda/devcontainer-feature.json
```diff
@@ -22,6 +22,7 @@
     },
     "containerEnv": {
         "CONDA_DIR": "/opt/conda",
+        "CONDA_SCRIPT":"${CONDA_DIR}/etc/profile.d/conda.sh",
         "PATH": "${PATH}:${CONDA_DIR}/bin:"
```
joshspicer, line 26:
```suggestion
        "PATH": "${PATH}:${CONDA_DIR}/bin"
```

I know you didn't change this, but @jungaretti's eagle eyes caught this.

src/conda/devcontainer-feature.json
```diff
@@ -22,6 +22,7 @@
     },
     "containerEnv": {
         "CONDA_DIR": "/opt/conda",
-        "PATH": "${PATH}:${CONDA_DIR}/bin:"
+        "CONDA_SCRIPT":"${CONDA_DIR}/etc/profile.d/conda.sh",
+        "PATH": "${PATH}:${CONDA_DIR}/bin"
```
joshspicer, line 26: I think this pattern that we've adopted here (using one containerEnv inside another) -_might_ be dangerous moving forward, since I think that the order this object is emitted into the intermediate Dockerfile is important. Right now the tooling ends up emitting in the provided order

image


image


image

--- 

Notably, this output is different when using docker exec directly
image

joshspicer: Changing the order to:

```
    "containerEnv": {
        "PATH": "${PATH}:${CONDA_DIR}/bin",
        "CONDA_SCRIPT":"${CONDA_DIR}/etc/profile.d/conda.sh",
        "CONDA_DIR": "/opt/conda"
    }
```

The auto-generated test fails:
image

image

image

image

image

joshspicer: It seems to make the most sense to me to encode information about the `$PATH` into the `devcontainer-feature.json`, but we might want to explore a more explicit approach.  Note that this is only showing one feature, but multiple features might alter the `$PATH` variable in different ways.

samruddhikhandale: That's a very good point. Changed all the usage of `containerEnv` to no longer depend on another `containerEnv` so that we could avoid the ordering issue.

joshspicer (approved):

PR #160, opened 2022-09-16
feat(nix): contribute nix devcontainer feature

ghuntley:
👋 @Chuxel - https://twitter.com/Chuxel/status/1570582551458369537?s=20&t=Fu1-XS51U969wNmKjP-vwQ

I have signed the Microsoft CLA and surrender all rights and copyright to Microsoft. The intention here is to transfer ownership of this feature layer back to the "Dev Container Spec Maintainers" as nix should be a 1st class citzen.

nb: I haven't tested this works with VSCode devcontainers but the steps are from knowledge I'd taken from my personal docker containers so it should "just work" in theory with minor touchups. Please rework the PR as you see fit to get it over the line.

### DIFF
--- a/src/nix/devcontainer-feature.json
+++ b/src/nix/devcontainer-feature.json
@@ -1,0 +1,26 @@
+{
+    "id": "nix",
+    "version": "1.0.0",
+    "name": "Nix",
+    "documentationURL": "https://github.com/devcontainers/features/tree/main/src/nix",
+    "description": "Installs Nix which is a tool that takes a unique approach to package management and system configuration",
+    "options": {
+        "version": {
+            "type": "string",
+            "enum": [
+                "latest"
+            ],
+            "default": "latest",
+            "description": "Currently unused."
+        }
+    },
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "bbenoist.Nix"
+            ]
+        }
+    },
+    "containerEnv": {
+    }
+}
--- a/src/nix/install.sh
+++ b/src/nix/install.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+#
+# Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/sshd.md
+# Maintainer: The VS Code and Codespaces Teams
+#
+
+set -e
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
+
+sh <(curl -L https://nixos.org/nix/install) --daemon
+
+mkdir -p $HOME/.config/nix $HOME/.config/nixpkgs
+echo 'sandbox = false' >> $HOME/.config/nix/nix.conf
+echo '{ allowUnfree = true; }' >> $HOME/.config/nixpkgs/config.nix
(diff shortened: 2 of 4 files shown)


src/nix/install.sh
```diff
@@ -0,0 +1,24 @@
...
+    exit 1
+fi
+
+
+sh <(curl -L https://nixos.org/nix/install) --daemon
+
+mkdir -p $HOME/.config/nix $HOME/.config/nixpkgs
+echo 'sandbox = false' >> $HOME/.config/nix/nix.conf
+echo '{ allowUnfree = true; }' >> $HOME/.config/nixpkgs/config.nix
+echo '. $HOME/.nix-profile/etc/profile.d/nix.sh' >> $HOME/.bashrc
```
ghuntley, line 24: If the user uses any other shell than bash then this would break. Not sure how to resolve - opening PR up for access to review hivemind knowledge.

jkeech, line 24: I expect that we'll want these setup in the `$HOME` of the `remoteUser`. This is a bit tricky today, and https://github.com/devcontainers/spec/issues/91 is tracking passing in the `remoteUser` as an environment variable to feature install scripts. It might be simplest to wait for that change to land to keep this feature script simple.

src/nix/install.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -e
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
+
+sh <(curl -L https://nixos.org/nix/install) --daemon
```
jkeech, line 19: I'm not very familiar with Nix. Do you know if it works well in daemon mode in docker containers? Do we need to add anything to the `entrypoint` to start the daemon on container startup?

src/nix/devcontainer-feature.json
```diff
@@ -0,0 +1,26 @@
+{
+    "id": "nix",
+    "version": "1.0.0",
+    "name": "Nix",
+    "documentationURL": "https://github.com/devcontainers/features/tree/main/src/nix",
+    "description": "Installs Nix which is a tool that takes a unique approach to package management and system configuration",
+    "options": {
```
jkeech, line 7: Would it make sense to have an option for a set of Nix packages to include in the installation? That way the package downloads can be done during an image prebuild to save time later on when users create their dev environment from the image.

jkeech (commented):
Thanks for opening this PR! We've heard a few requests for Nix support, and I agree it could be one of the supported ones in this feature collection.

We also have a self-publishing mechanism available if you want to publish and maintain this nix feature in your own repo: https://github.com/devcontainers/feature-template. We'd love for the community to be able to publish and use features without having to go through a central repo.

joshspicer:
+1 to exploring our "self publishing mechanism".  

Our intention is that features in this repo are not treated differently from community features (other than attribution for the author of the feature), given that the author follows the distribution steps we're shared as a specification (https://containers.dev/implementors/features-distribution/).  As @jkeech mentioned, we have a template to help bootstrap a features repository (https://github.com/devcontainers/feature-template).  Feel free to tag me for feedback if you decide to go down this route.

 As a plus, you would not need to sign the Microsoft CLA 😛

src/nix/install.sh
Chuxel: Yeah an enterpoint would be needed.  This I think would also require the container run as root - which isn't a problem for services like Codespaces or when using Docker Desktop, but is a consideration otherwise.

Most of the exploration I did for https://github.com/Chuxel/feature-library/blob/main/src/nix/install.sh was trying to figure out a way to use single user unless the UID/GID changed. I think I have a working model.  We could opt to start from here as a baseline if preferrable.

It also verifies the download using their GPG key, has support for referencing a list of packages, a flake, or a derivation.

samruddhikhandale:
Hi @ghuntley,
We have released a [Nix Feature](https://github.com/devcontainers/features/tree/main/src/nix) ✨ 

Also, we had CC'ed you in the [PR](https://github.com/devcontainers/features/pull/228) with the changes. We are happy to take any feedbacks. Thanks!

PR #161, opened 2022-09-16
Adding azure developer cli

aaronpowell:
This is a good companion with the azure-cli for local dev.

It uses a slightly adjusted version of the install docs from https://github.com/Azure/azure-dev#linuxmacos so that it can install specified versions.

Since you need the azure-cli installed, I've set the `installsAfter` for that, creating a soft dependency (as per the docs).

The definition is a clone of the `azure-cli` extension.


samruddhikhandale (commented):
Thank you for for opening the PR. Azure Dev CLI would be an exciting one to be added to the Feature's collection. 

We also have a self-publishing mechanism available if you want to publish and maintain this feature in your own repo: https://github.com/devcontainers/feature-template. We'd love for the community to be able to publish and use features without having to go through a central repo.

You could also refer to https://github.com/devcontainers/features#contributions for contributing details.

jkeech:
To me this seems like a good candidate for an option inside the `azure-cli` feature, such as `"installDeveloperCli": true`, rather than a separate feature. The default could even be `true` to make it opt-out instead of opt-in.

I think of features as more "scenario scoped" rather than individual tools. That way users don't need to include as many features in their devcontainer.json. Similar to how the "node" feature brings in "yarn".

aaronpowell:
> To me this seems like a good candidate for an option inside the `azure-cli` feature, such as `"installDeveloperCli": true`, rather than a separate feature. The default could even be `true` to make it opt-out instead of opt-in.
> 
> I think of features as more "scenario scoped" rather than individual tools. That way users don't need to include as many features in their devcontainer.json. Similar to how the "node" feature brings in "yarn".

Sure, happy to propose that as an option instead. I wasn't sure how thinly slices the scope of a feature should be, whether it should be _do a single thing_ or not. I'll create a new PR that adds it to the existing `azure-cli` feature.

jkeech:
>Sure, happy to propose that as an option instead. 

Great, please feel free to make those changes in this PR 👍. Let us know if you have any questions or hit any issues adding it to the existing `azure-cli` feature.

>I wasn't sure how thinly slices the scope of a feature should be, whether it should be do a single thing or not. I'll create a new PR that adds it to the existing azure-cli feature.

Technically features can be at any granularity. Features in this repo tend to be somewhat opinionated about pulling in all of the configuration needed to enable specific scenarios, which often includes a set of tools to install in the image as well as VS Code extensions or other customizations that are tied to the same scenario. For a typical Azure developer, it would be nice to pull in a single feature to get all of the tools they need for working with Azure, including both the `az` CLI and the new `azd` CLI.

aaronpowell:
> > Sure, happy to propose that as an option instead.
> 
> Great, please feel free to make those changes in this PR 👍. Let us know if you have any questions or hit any issues adding it to the existing `azure-cli` feature.

Merged across to the `azure-cli` feature.

src/azure-cli/README.md
```diff
@@ -1,14 +1,15 @@
...
+Installs the Azure CLI along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like gpg. It will also optionally install the Azure Developer CLI.
 
 ## Example Usage
 
 ```json
 "features": {
         "ghcr.io/devcontainers/features/azure-cli:1": {
-            "version": "latest"
+            "version": "latest",
+            "azure-dev-cli-version": "latest"
```
jkeech, line 12: Can you switch this to camelCase to be consistent with other features in this repo? Example: https://github.com/devcontainers/features/tree/main/src/desktop-lite
```suggestion
            "azureDevCliVersion": "latest"
```

src/azure-cli/install.sh
```diff
@@ -10,6 +10,7 @@
 set -e
 
 AZ_VERSION=${VERSION:-"latest"}
+AZ_DEV_VERSION=${VERSION:-"latest"}
```
jkeech, line 13: The right hand side of this will need to match the name of the option (uppercased). If the option name is `azureDevCliVersion`, the environment variable that will be set will be `AZUREDEVCLIVERSION`, so you could have the following
```suggestion
AZ_DEV_VERSION=${AZUREDEVCLIVERSION:-"latest"}
```

src/azure-cli/install.sh
```diff
@@ -179,4 +180,15 @@ if [ "${use_pip}" = "true" ]; then
     fi
 fi
 
+if [[ "${AZ_DEV_VERSION}" -ne "none" ]]; then
+    echo "(*) Installing Azure Dev CLI"
+
+    curl -fsSLO https://aka.ms/install-azd.sh
+    chmod +x ./install-azd.sh
+
+    ./install-azd.sh --version $AZ_DEV_VERSION
```
jkeech, line 189: Can you add a test in https://github.com/devcontainers/features/blob/main/test/azure-cli/test.sh to verify that the `azd` CLI is installed properly? If you want to add multiple test cases to verify with and without `azd` installed, you can set up multiple scenarios with different sets of feature options following the example in https://github.com/devcontainers/features/tree/main/test/dotnet.

jkeech (changes requested):
Thanks for making this change! This is looking good, but there are a couple tweaks that would be good to make before merging.

src/azure-cli/install.sh
aaronpowell: Added a single case to the main test for `azure-cli`.

I could break it to multiple scenarios if you prefer (although I'm not sure how to run the tests)

jkeech: Thanks for adding the test case! Looks like the test is failing and it's not installed `azd` as part of the feature execution.

You should be able to reproduce locally by running the test with the `devcontainer features test` command like the CI does: https://github.com/devcontainers/features/blob/main/.github/workflows/test-all.yaml#L53-L57

jkeech: I'm not a bash expert, but I saw somewhere that you might not be able to use `-ne` with string comparisons. Try changing `-ne` to `!=` and see if the installation works / tests pass.

aaronpowell: Ah, should have looked for a workflow file to see how the tests work run.

I'll work on getting the devcontainer cli installed (was fighting that the other day due to the cascading dependencies, maybe I should try a devcontainer to run it in 😜)

Aztekcito (approved):

joshspicer (approved):
Thanks!

jongio:
Please do not merge this. Discussing...

src/azure-cli/README.md
```diff
@@ -1,14 +1,15 @@
 
 # Azure CLI (azure-cli)
 
-Installs the Azure CLI along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like gpg.
+Installs the Azure CLI along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like gpg. It will also optionally install the Azure Developer CLI.
```
jongio, line 4: Adding a comment here to make sure this is not merged.

Chuxel:
I'll close this PR given https://github.com/devcontainers/features/pull/161#issuecomment-1309215124. We can reopen if things change.

Issue #159, opened 2022-09-14
Go feature does not add `$GOPATH/bin` to the `PATH`


davidwallacejackson:
Repro: https://github.com/davidwallacejackson/go-path-repro

I think this issue happens because `$GOROOT/bin` has already been added to `PATH` somehow before [these injected rc lines](https://github.com/devcontainers/features/blob/main/src/go/install.sh#L236) are run. Because the default `GOROOT` is /usr/local/go`, the path to its `bin` directory is `/usr/local/go/bin`, which contains `/go/bin` -- which then causes the match `*"\${GOPATH}/bin"*` to pass erroneously (since `/go/bin` can be found in the `PATH` string). This might be fixable just by using a regex for the match instead.

This is another example of something that would be solved by https://github.com/devcontainers/features/issues/155

samruddhikhandale:
Fixed the issue with https://github.com/devcontainers/features/pull/162

@davidwallacejackson if you could rebuild the container (without cache), you could see the `PATH` issue resolved. 
I tested with https://github.com/davidwallacejackson/go-path-repro and the fix worked fine. 

<img width="956" alt="image" src="https://user-images.githubusercontent.com/24955913/190791300-985bf915-ae45-4a37-a74f-40878002ab28.png">

PR #162, opened 2022-09-16
Go - Adds `/go/bin` to PATH

samruddhikhandale:
Fixes - https://github.com/devcontainers/features/issues/159

`go` is installed at `/usr/local/go/bin` and the go tools are installed at `/go/bin`. Updated `containerEnv` to reflect the same. 

Also, removed `updateRc` which updated `GOROOT`, `GOPATH` and `PATH` (`containerEnv` already updates that)

### DIFF
--- a/src/go/devcontainer-feature.json
+++ b/src/go/devcontainer-feature.json
@@ -27,8 +27,8 @@
     },
     "containerEnv": {
         "GOROOT": "/usr/local/go",
-        "GOPATH": "/usr/local/go",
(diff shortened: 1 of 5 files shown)


joshspicer (approved):
Looks good to me - I especially like the scenarios added here to exercise slightly more complex cases 🚀

@yuku - I believe you shared some go insights with us over on vscode-dev-containers. Would love your opinion on if these container envs all look good :)

PR #174, opened 2022-09-20
Rename JupyterLab config file

jungaretti:
Jupyter recently changed the name of the file used to configure JupyterLab. It's now `jupyter_server_config.py` instead of `jupyter_notebook_config.py`. This PR updates the name of the file authored by the Python feature.

https://jupyter-server.readthedocs.io/en/latest/users/configuration.html

I'm not the biggest fan of the Python feature managing this config file, but I think we should correct this filename ASAP while we work on refactoring the feature.

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.0.6",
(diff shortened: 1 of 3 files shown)


samruddhikhandale (approved):
LGTM 🚀

Issue #177, opened 2022-09-22
No xdg-open command, so host Browser cannot be started


nmemoto:
When starting a browser by executing cli on a terminal in devcontainer (VSCode Remote - Container), if the browser startup depends on xdg-open, it will fail because the xdg-open command does not exist.

For example, the following is obtained by executing cdktf init.

```
$ cdktf init --template="typescript"
Welcome to CDK for Terraform!

By default, cdktf allows you to manage the state of your stacks using Terraform Cloud for free.
cdktf will request an API token for app.terraform.io using your browser.

If login is successful, cdktf will store the token in plain text in
the following file for use by subsequent Terraform commands:
    /home/node/.terraform.d/credentials.tfrc.json

Note: The local storage mode isn't recommended for storing the state of your stacks.

? Do you want to continue with Terraform Cloud remote state management? Yes

opening webpage using your browser.....

If the web browser didn't open the window automatically, you can go to the following url:
        https://app.terraform.io/app/settings/tokens?source=terraform-login

? Token for app.terraform.io 🔑 [input is hidden] /usr/local/share/npm-global/lib/node_modules/cdktf-cli/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

Error: spawn xdg-open ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (node:internal/child_process:289:12)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn xdg-open',
  path: 'xdg-open',
  spawnargs: [
    'https://app.terraform.io/app/settings/tokens?source=terraform-login'
  ]
}
```

CLI tools that use the environment variable BROWSER, such as github-cli, are not affected, but there must be many CLI tools that use xdg-open.

I originally submitted the form below, but I thought it was a devcontainer issue, so I submitted it here.

https://github.com/hashicorp/terraform-cdk/issues/2132

nmemoto:
#178 is my suggestion. Could you please confirm it?

samruddhikhandale:
Hi @nmemoto,
[...]

Issue #67, opened 2022-07-19
Consider splitting up the `common` feature


alexdima:
We could consider deprecating the common feature (similar to some of the ones mentioned in https://github.com/devcontainers/features/issues/64) and creating multiple features out of it.

The current `common` feature contains (text from @Chuxel):

---

### 1. [Common OS Utilities](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/common-debian.sh#L74-L165), [set a locale if one not there](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/common-debian.sh#L174-L180), [systemctl shim](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/common-debian.sh#L285-L295)

- [ ] A “common utilities” feature w/o sudo.  To recap, its goal is to install: 

* Common command line utilities used in shell scripts (e.g., ps, lsof, wget, curl + ca-certificates zip, unzip, tar, net-tools) 
* Utilities frequently referenced in Linux install instructions (apt-transport-https, lsb-release, jq, gnupg2 + dirmngr) 
* Man pages (man-db, manpages, manpages-dev, manpages-posix manpages-posix-dev) 
* Libraries frequently required by native code that is not statically compiled (libc6, libgcc1, libkrb5-3, libgssapi-krb5-2, libicu[0-9][0-9], liblttng-ust[0-9], libstdc++6, zlib1g)  
* Give people a hint for what to do w/o systemctl since systemd is rarely in container images, but pass through if it’s there. Lots of Linux instructions reference it so it was added in time because of customer confusion. 

> Many of these are already in images like “node”, but they’re checked for coverage since they can be missing from minimal images like “debian”. Adding a default locale is a good idea given the number of developer tools that can bomb without it, but we could give the option to pick a locale. Right now, it’s always en_US.UTF-8. In theory we could pull out git in favor of the git feature but given how central it is to developer images and it only installs if not present, it might be worth keeping.

---

### 2. [Default bash theme](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/common-debian.sh#L301-L322)

- [ ] Include in “common utilities” but change the default so the source code portion of it is off by default since this can be slow in massive repos. Also disabling source control is mandatory if git is removed from common utilities. Rename it to “devcontainers-default” instead of “codespaces.”

---

### 3. [apt-get upgrade option](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/common-debian.sh#L167-L172)

[...]

PR #178, opened 2022-09-22
common-utils: add xdg-utils package

nmemoto:
One of the #177 solutions


bamurtaugh:
Thanks for opening! I believe https://github.com/devcontainers/features/issues/67 would be relevant, as there is discussing on splitting up the common Feature.

nmemoto:
@bamurtaugh I was not aware of this discussion #67. Thank you for telling me about it.

My suggestion is a change that (as far as I know) should only apply when using vscode.
I am going to close this pull request. And I hope I can suggest something in this discussion #67 .

PR #182, opened 2022-09-23
Fix additional Python and JupyterLab edge case

jungaretti:
Related: https://github.com/devcontainers/images/issues/125

Fixes an edge case in the Python feature that installs JupyterLab to the `additionalVersions` of Python instead of the primary `version` of Python.

This PR also removes junk from our JupyterLab test and adds a new test that would have caught this bug.

### Edge Case Details

`install_user_package` installs packages to whatever Python installation is specified by `$INSTALL_PATH`. Both `install_from_source` and `install_using_oryx` overwrite this variable. When we use `install_python` to install an extra version of Python, we overwrite `$INSTALL_PATH` to that installation.

When we install JupyterLab at the end of the script, it uses the `$INSTALL_PATH` set by the most recent `additionalVersions` instead of `version`.

### DIFF
--- a/test/python/scenarios.json
+++ b/test/python/scenarios.json
@@ -12,9 +12,6 @@
         "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
         "remoteUser": "vscode",
         "features": {
-            "common-utils": {
-                "username": "vscode"
-            },
             "python": {
                 "installJupyterlab": true,
                 "configureJupyterlabAllowOrigin": "*"
--- a/test/python/scenarios.json
+++ b/test/python/scenarios.json
@@ -18,6 +18,18 @@
             }
         }
     },
+    "install_additional_jupyterlab": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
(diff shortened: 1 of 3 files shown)


test/python/scenarios.json
```diff
@@ -12,15 +12,24 @@
...
-                "username": "vscode"
-            },
             "python": {
                 "installJupyterlab": true,
                 "configureJupyterlabAllowOrigin": "*"
             }
         }
     },
+    "install_additional_jupyterlab": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base:focal",
```
samruddhikhandale, line 22:
```suggestion
        "image": "mcr.microsoft.com/devcontainers/base:focal",
```

samruddhikhandale: Can we use the new image paths?

samruddhikhandale (approved):
LGTM ✨ 
Can we make the image path change before merging this PR?

samruddhikhandale (dismissed):
Thanks!

samruddhikhandale (approved):

PR #187, opened 2022-09-28
feat(python): add poetry option

### LINKED ISSUE #167: Add option to install poetry in python feature
[Poetry](https://python-poetry.org/) is a great dependency management tool for python, and it will be great if we could install it as a feature.

solismaam:
Suggestion to resolve #167


solismaam:

src/python/install.sh
```diff
@@ -454,4 +456,9 @@ if [ "${INSTALL_JUPYTERLAB}" = "true" ]; then
     fi
 fi
 
+# Install Poetry if needed
+if [ "${INSTALL_POETRY}" = "true" ]; then
+    install_user_package poetry
```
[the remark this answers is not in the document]
solismaam: Installation is now done with pipx instead of pip.
Ref: https://python-poetry.org/docs/#installing-with-pipx

joshspicer:
Hi @solismaam - thanks for taking the time to contribute!

I'd like to offer 2 alternative solution to this.  We're hoping to remove the `jupyter-labs` option in the Python Feature, as adding individual options for each pip package X does not seem sustainable.

## A: Add an option `installPipPackages`.

Then an example `devcontainer.json` would look something like this.

```
{
   "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
   "features" : {
      "ghcr.io/devcontainers/features/python:1": {
          "installPipPackages": "poetry requests"
       }
   }
}
```

## B: Self-author a `poetry` Feature

The trick here is to use the [`installsAfter`] property from the spec (https://containers.dev/implementors/features/#installsAfter).

Using this [template repo](https://github.com/devcontainers/feature-template), you'd be able to author an additional Feature to be referenced side by side with the existing python Feature here, but hosted in a repo of your own. 

I'd like this approach if you think that in the future the amount of configuration for the poetry tool may grow; otherwise, option A might be more simple.

##### src/poerty/devcontainer-feature.json

```json
{
   "id" "poetry",
   "installsAfter": [
      "ghcr.io/devcontainers/features/python"
   ]
}
```

An example `devcontainer.json` would look like:

```
{
   "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
   "features" : {
      "ghcr.io/devcontainers/features/python:1": {},
      "ghcr.io/solismaam/features/poetry:1": {},

   }
}

```

The `installsAfter` will enforce that `python` is installed before a `poetry`.

solismaam:
Thanks for your feedback @joshspicer !

I agree with you that adding individual options to every pip package is not sustainable. I originally thought I would add `poetry` to the `DEFAULT_UTILS` list, but then I decided to follow how `juputer-labs` has been implemented.

I think Python Feature users should be able to add other utilities (cli tools like `poetry`) via pipx. What do you think about if the additional option for Python Feature would be `installAdditionalUtils`?

solismaam:
Refactored proposal done - any comments @joshspicer ?

src/python/devcontainer-feature.json
```diff
@@ -25,6 +25,11 @@
       "default": true,
       "description": "Install common Python tools like pylint"
     },
+    "installToolsExtra": {
```
joshspicer, line 28: nit: I think `installExtraTools` sounds a bit better?

src/python/devcontainer-feature.json
```diff
@@ -25,6 +25,11 @@
...
+      "type": "string",
+      "default": "",
```
joshspicer, line 30: Can you add an empty `proposals` array here? The schema can be seen in Typescript here: https://github.com/devcontainers/cli/blob/main/src/spec-configuration/containerFeaturesConfiguration.ts#L52-L72

"Proposals" are meant to act as an array of suggestions, but also indicates that the user can pass whatever argument they would like.  The "enum" variant, on the other hand, represents a the only set of permitted arguments.  (This isn't enforced in the CLI, but may be someday).

src/python/install.sh
```diff
@@ -429,6 +430,7 @@ if [[ "${INSTALL_PYTHON_TOOLS}" = "true" ]] && [[ $(python --version) != "" ]];
         /tmp/pip-tmp/bin/pipx install --pip-args=--no-cache-dir pipx
         PIPX_DIR="/tmp/pip-tmp/bin"
     fi
+    if [[ "${INSTALL_PYTHON_TOOLS_EXTRA}" != "" ]]; then DEFAULT_UTILS=(${DEFAULT_UTILS[@]} ${INSTALL_PYTHON_TOOLS_EXTRA}); fi
```
joshspicer, line 433: style nit

```suggestion
    if [[ "${INSTALL_PYTHON_TOOLS_EXTRA}" != "" ]]; then 
       DEFAULT_UTILS=(${DEFAULT_UTILS[@]} ${INSTALL_PYTHON_TOOLS_EXTRA})
    fi
```

joshspicer (commented):
Thank you for the contribution! Some minor nits but otherwise looks great. Thanks for adding the test.

cc/ @jungaretti if you have any final python Feature thoughts.

src/python/install.sh
solismaam: Done.

src/python/devcontainer-feature.json
solismaam: Added empty `proposals`. 

Should I move the example in the description of the option to the `proposals` attribute?

```
proposals: [
   "poetry",
   "isort poetry sphinx"
]
```

jungaretti (commented):
Thanks for contributing. Our long-term goal for this Feature is to remove the extra package installations (including JupyterLab) and move them to brand-new Features that take advantage of `installsAfter`.

The [Python feature installs `pipx`](https://github.com/devcontainers/features/blob/4dae930c16a0b9ec2494ea11d74e26b7a0b56689/src/python/install.sh#L432) when its `installTools` option is `true` ([its default value](https://github.com/devcontainers/features/blob/4dae930c16a0b9ec2494ea11d74e26b7a0b56689/src/python/devcontainer-feature.json#L23-L27)).

I strongly prefer @joshspicer's idea of a new Feature that installs after the Python feature and uses `pipx` to install your desired tools. Here's what that new Feature could do:
1. Check for a valid installation of `pipx`
    ```bash
    if ! (pipx --version); then
        echo 'pipx is not available'
        exit 1
    fi
    ```
2. Install `isort`, `poetry`, `sphinx`, etc. using `pipx`
    ```bash
    pipx install isort
    ```

Adding this functionality to a new Feature will help us keep the Python feature light and maintainable.

solismaam:
This PR doesn't seem to be going anywhere, so I've closed it.

PR #196, opened 2022-10-05
Move conda description to description property

joshspicer:


src/conda/devcontainer-feature.json
```diff
@@ -1,7 +1,8 @@
 {
     "id": "conda",
     "version": "1.0.2",
```
samruddhikhandale, line 3: version bump? 😄

samruddhikhandale: I believe that's still needed to reflect this change in OCI for UX changes? or no?

joshspicer: yes good point!

joshspicer: Although...not strictly needed in this case since the index file will end up being updated after concatenating this new config.  But nonetheless good practice.

samruddhikhandale (approved):

PR #199, opened 2022-10-06
Node: Fix nvm ownership

samruddhikhandale:
Adds correct ownership to `nvm` group for `$NVM_DIR`

### DIFF
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -107,7 +107,7 @@ export NVM_SYMLINK_CURRENT=true
 if [ -d "${NVM_DIR}" ]; then
     echo "NVM already installed."
     if [ "${NODE_VERSION}" != "" ]; then
-       su ${USERNAME} -c "umask 0002 && . $NVM_DIR/nvm.sh && nvm install ${NODE_VERSION} && nvm clear-cache"
+       su ${USERNAME} -c ". $NVM_DIR/nvm.sh && nvm install ${NODE_VERSION} && nvm clear-cache"
     fi
     exit 0
 fi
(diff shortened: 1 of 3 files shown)


johnnyreilly:
oh is this going to resolve the weird permissioning issues I've been getting with Node I wonder?

src/node/install.sh
```diff
@@ -107,7 +107,7 @@ export NVM_SYMLINK_CURRENT=true
 if [ -d "${NVM_DIR}" ]; then
     echo "NVM already installed."
     if [ "${NODE_VERSION}" != "" ]; then
-       su ${USERNAME} -c ". $NVM_DIR/nvm.sh && nvm install ${NODE_VERSION} && nvm clear-cache"
+       su ${USERNAME} -c "umask 0002 && . $NVM_DIR/nvm.sh && nvm install ${NODE_VERSION} && nvm clear-cache"
```
joshspicer, line 110: What is the `2` bit doing here in `umask 0002`

src/node/install.sh
```diff
@@ -185,4 +185,6 @@ if [ "${INSTALL_TOOLS_FOR_NODE_GYP}" = "true" ]; then
     fi
 fi
 
+find "${NVM_DIR}" -type d -print0 | xargs -n 1 -0 chmod g+s
```
joshspicer, line 188: Why is the `chmod` on line 123 not enough?

src/node/install.sh
samruddhikhandale: Removed that, wasn't needed to fix the issue.

src/node/install.sh
samruddhikhandale: if you use a package manager and you are running as root, then the files/folders will not inherit those privs, for example. The `find` command helps address that.

This scenario fails when running tests in action

image

joshspicer (approved):

samruddhikhandale:
@johnnyreilly The `node` Feature should now contain this change. 🚀 
Feel free to open an issue if it doesn't fix your permissions issue.

PR #200, opened 2022-10-06
Add docker-in-docker azure_dns_override flag

kklopfenstein:
Add a flag (defaulted to `true`) to toggle passing the Azure DNS ip address to dockerd at startup.

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -11,7 +11,7 @@
 DOCKER_VERSION=${VERSION:-"latest"} # The Docker/Moby Engine + CLI should match in version
 USE_MOBY=${MOBY:-"true"}
 DOCKER_DASH_COMPOSE_VERSION=${DOCKERDASHCOMPOSEVERSION:-"v1"} # v1 or v2
-AZURE_DNS_OVERRIDE=${AZUREDNSOVERRIDE:-"true"}
+AZURE_DNS_AUTO_DETECTION=${AZUREDNSAUTODETECTION:-"true"}
 
 ENABLE_NONROOT_DOCKER=${ENABLE_NONROOT_DOCKER:-"true"}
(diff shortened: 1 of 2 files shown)


joshspicer:
related to: https://github.com/microsoft/vscode-dev-containers/pull/1650

src/docker-in-docker/install.sh
```diff
@@ -11,7 +11,7 @@
 DOCKER_VERSION=${VERSION:-"latest"} # The Docker/Moby Engine + CLI should match in version
 USE_MOBY=${MOBY:-"true"}
 DOCKER_DASH_COMPOSE_VERSION=${DOCKERDASHCOMPOSEVERSION:-"v1"} # v1 or v2
-AZURE_DNS_OVERRIDE=${AZUREDNSOVERRIDE:-"true"}
+AZURE_DNS_AUTO_DETECTION=${AZUREDNSAUTODETECTION:-"true"}
```
kklopfenstein, line 14: This might be my own lack of understanding, but I'm a little confused by the "auto detection" part of the name.

joshspicer: you change here (flipping this flag to `false`) is turning off that auto-detection, and using the default `dockerd` DNS config.

joshspicer: I changed it from `AZURE_DNS_OVERRIDE` since turning this flag on will _not necessarily_ fire these lines

https://github.com/devcontainers/features/pull/200/files#diff-a72a4e49c44e54f09feb84b7767d355d68d34fbee561d8f96a99d12d2488b9c5R371-R372

kklopfenstein: Oh, good point! The name makes sense to me now. 😄

joshspicer: Ok perfect :)

joshspicer (approved):

PR #201, opened 2022-10-06
Common-utils: Fix bug - optional first run image specific notice

samruddhikhandale:


src/common-utils/devcontainer-feature.json
```diff
@@ -54,5 +54,8 @@
             "default": false,
             "description": "Add packages from non-free Debian repository?"
         }
+    },
+    "containerEnv": {
+        "DEV_CONTAINERS_DIR": "/usr/local/etc/vscode-dev-containers"
```
joshspicer, line 59: For future context, why do we need this?

joshspicer (approved):

src/common-utils/devcontainer-feature.json
samruddhikhandale: removed that.

joshspicer (approved):

PR #210, opened 2022-10-09
Ensure remove apt-update cache at the beginning and end of the scripts

### LINKED ISSUE #191: Remove apt lists (`/var/lib/apt/lists/*`) the end of each scripts?
Previously, a mechanism was implemented in each script to reduce the number of `apt-get update`, so that only one `apt-get update` was needed to install multiple features.
However, that mechanism has now been removed, so it may make more sense to remove the cache by `rm -rf /var/lib/apt/lists/*` at the end of each script to reduce the layer size.
(especially if you pre-build and register the image in the registry)

eitsupi:
Fix #191

This PR re-implement the mechanism to skip `apt-get update` if it is already cached.
Deleting the cache at the beginning as well as at the end of a script avoids old caches.

I thought I checked all the scripts, but I could be missing something!

### DIFF
--- a/src/anaconda/devcontainer-feature.json
+++ b/src/anaconda/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "anaconda",
-    "version": "1.0.7",
+    "version": "1.0.8",
     "name": "Anaconda",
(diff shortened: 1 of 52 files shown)


joshspicer (approved):
Thank you @eitsupi! Much appreciated :)

PR #223, opened 2022-10-17
Add default-address-pool setting

kklopfenstein:
Add the ability to pass `--default-address-pool` settings to dockerd.

### DIFF
--- a/src/docker-in-docker/devcontainer-feature.json
+++ b/src/docker-in-docker/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "docker-in-docker",
-    "version": "1.0.7",
+    "version": "1.0.8",
     "name": "Docker (Docker-in-Docker)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/docker-in-docker",
     "description": "Create child containers *inside* a container, independent from the host's docker instance. Installs Docker extension in the container along with needed CLIs.",
@@ -33,6 +33,12 @@
             "type": "boolean",
             "default": true,
             "description": "Allow automatically setting the dockerd DNS server when the installation script detects it is running in Azure"
+        },
+        "dockerDefaultAddressPool": {
+            "type": "string",
+            "default": "",
+            "proposals": [],
+            "description": "Define default address pools for Docker networks. e.g. base=192.168.0.0/16,size=24"
         }
     },
     "entrypoint": "/usr/local/share/docker-init.sh",
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -12,6 +12,7 @@ DOCKER_VERSION=${VERSION:-"latest"} # The Docker/Moby Engine + CLI should match 
 USE_MOBY=${MOBY:-"true"}
 DOCKER_DASH_COMPOSE_VERSION=${DOCKERDASHCOMPOSEVERSION:-"v1"} # v1 or v2
 AZURE_DNS_AUTO_DETECTION=${AZUREDNSAUTODETECTION:-"true"}
+DOCKER_DEFAULT_ADDRESS_POOL=${DOCKERDEFAULTADDRESSPOOL}
 
 ENABLE_NONROOT_DOCKER=${ENABLE_NONROOT_DOCKER:-"true"}
 USERNAME=${USERNAME:-"automatic"}
@@ -325,12 +326,13 @@ tee /usr/local/share/docker-init.sh > /dev/null \
 
 set -e
 
-AZURE_DNS_AUTO_DETECTION=$AZURE_DNS_AUTO_DETECTION
+AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION}
+DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL}
 EOF
 
 tee -a /usr/local/share/docker-init.sh > /dev/null \
 << 'EOF'
-dockerd_start="$(cat << 'INNEREOF'
+dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL} $(cat << 'INNEREOF'
     # explicitly remove dockerd and containerd PID file to ensure that it can start properly if it was stopped uncleanly
     # ie: docker kill <ID>
     find /run /var/run -iname 'docker*.pid' -delete || :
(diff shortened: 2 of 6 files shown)


kklopfenstein:
@microsoft-github-policy-service agree company="github"

src/docker-in-docker/install.sh
```diff
@@ -325,7 +326,8 @@ tee /usr/local/share/docker-init.sh > /dev/null \
 
 set -e
 
-AZURE_DNS_AUTO_DETECTION=$AZURE_DNS_AUTO_DETECTION
+export AZURE_DNS_AUTO_DETECTION=$AZURE_DNS_AUTO_DETECTION
+export DOCKER_DEFAULT_ADDRESS_POOL=$DOCKER_DEFAULT_ADDRESS_POOL
```
CleanCut, line 330: These don't need to be exported as they are only being used inside the script they are being injected into.

```suggestion
AZURE_DNS_AUTO_DETECTION=$AZURE_DNS_AUTO_DETECTION
DOCKER_DEFAULT_ADDRESS_POOL=$DOCKER_DEFAULT_ADDRESS_POOL
```

src/docker-in-docker/install.sh
```diff
@@ -377,16 +379,24 @@ dockerd_start="$(cat << 'INNEREOF'
         echo "Not setting dockerd DNS manually."
         CUSTOMDNS=""
     fi
+
     set -e
 
+    if [ -z "$DOCKER_DEFAULT_ADDRESS_POOL" ] || [ "$DOCKER_DEFAULT_ADDRESS_POOL" = "" ]
```
CleanCut, line 385: These two checks are redundant.

```suggestion
    if [ -z "$DOCKER_DEFAULT_ADDRESS_POOL" ]
```

src/docker-in-docker/install.sh
```diff
@@ -377,16 +379,24 @@ dockerd_start="$(cat << 'INNEREOF'
...
     # Start docker/moby engine
-    ( dockerd $CUSTOMDNS > /tmp/dockerd.log 2>&1 ) &
+    ( dockerd $CUSTOMDNS $DEFAULT_ADDRESS_POOL > /tmp/dockerd.log 2>&1 ) &
 INNEREOF
 )"
 
 # Start using sudo if not invoked as root
 if [ "$(id -u)" -ne 0 ]; then
-    sudo /bin/sh -c "${dockerd_start}"
+    sudo -E /bin/sh -c "${dockerd_start}"
```
CleanCut, line 399: Passing the local environment to `sudo` shouldn't be necessary.

```suggestion
    sudo /bin/sh -c "${dockerd_start}"
```

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "docker-in-docker",
-    "version": "1.0.6",
+    "version": "1.0.7",
```
joshspicer, line 3: This is going to be 1.0.8 now 😅 .  It's on the list to automate bumping these with an action

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -33,6 +33,11 @@
             "type": "boolean",
             "default": true,
             "description": "Allow automatically setting the dockerd DNS server when the installation script detects it is running in Azure"
+        },
+        "dockerDefaultAddressPool": {
+            "type": "string",
```
joshspicer, line 38: Can you add an empty `proposals` array here?  The schema expects either an `enum` (cannot supply any other value besides what is provided), or `proposals` (suggestions can be placed here, and indicates to the user they can provide their own value)

src/docker-in-docker/install.sh
joshspicer:
```suggestion
    if [ -z "${DOCKER_DEFAULT_ADDRESS_POOL}" ]
```

src/docker-in-docker/install.sh
joshspicer: I'd also wrap these in `{ }` to stay consistent with the rest of the file (or at least the places touched recently).

kklopfenstein: If we don't export these, when `sudo -E /bin/sh -c "${dockerd_start}"` is run these variables are not available.

It does have the variables when run with eval, though.

src/docker-in-docker/install.sh
kklopfenstein: If `-E` isn't present then the command in `$dockerd_start` doesn't have the values of `AZURE_DNS_AUTO_DETECTION` or `DOCKER_DEFAULT_ADDRESS_POOL`.

src/docker-in-docker/install.sh
kklopfenstein: I was able to avoid exporting and using `sudo -E` by defining the variables inside `dockerd_start` directly in the generated `docker-init.sh` script. That way they're defined automatically when the script in that variable is run using `sudo`.

src/docker-in-docker/install.sh
```diff
@@ -325,12 +326,17 @@ tee /usr/local/share/docker-init.sh > /dev/null \
 
 set -e
 
-AZURE_DNS_AUTO_DETECTION=$AZURE_DNS_AUTO_DETECTION
+dockerd_start="$(cat << INNEREOF
+    AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION}
+    DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL}
+    
+INNEREOF
+)"
```
CleanCut, line 334: Would it be more straightforward to leave these as separate environment variables here:

```suggestion
AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION}
DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL}
```

...and then be explicit about injecting them down below like this?

```
dockerd_start="$(AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL} cat << 'INNEREOF'

```

kklopfenstein: That won't work. The heredoc below is quoted so it doesn't allow variable expansion.

kklopfenstein: Oh, nvm, I mis-interpreted what you were saying. I'm fine with that change.

CleanCut (dismissed):

joshspicer (approved):

joshspicer (approved):

PR #227, opened 2022-10-18
common-utils: Installs git from source

samruddhikhandale:


jkeech (changes requested):
I would prefer to keep the `common-utils` feature installing `git` via `apt`. For users who want to install from source or further customize the git installation options, the `git` feature can be used which provides those options already.

PR #228, opened 2022-10-18
Nix feature

Chuxel:
This PR adds a dev container feature for Nix. It supports using base images from Debian/Ubuntu, the RHEL tree, and Alpine.

I refactored some utilities into a utils.sh file to make the main script more readable.

The feature can just install nix, or install nix and a set of packages, a Flake, or a derivation. This is limited to things in the base image (or remote URIs in the case of Flakes) due to the lack of lifecycle hook support (devcontainers/spec#60). We can update once that is available.

Given Nix is picky about privs, the script supports doing both a single user and multi-user install. NOTES.md includes info on what each does and downsides by scenario.

Given the breadth of the capabilities, there's also a battery of tests which were unblocked by recent tweaks to the devcontainer CLI's testing setup.


Chuxel:
//cc @lostintangent

src/nix/install.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+# Move to the same directory as this script
+set -e
+FEATURE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "${FEATURE_DIR}"
+
+# Option defaults
+VERSION="${VERSION:-"latest"}"
```
samruddhikhandale, line 8: ℹ️  I don't think this is needed anymore. Can we remove it? 

`VERSION` if not provided by the user, will by default have `latest` as value (ie the default set in `devcontainer-feature.json`)

src/nix/install.sh
```diff
@@ -0,0 +1,128 @@
...
+MULTIUSER="${MULTIUSER:-"true"}"
```
samruddhikhandale, line 9: Can we remove all such variable assignments as well. Thanks!

src/nix/install.sh
```diff
@@ -0,0 +1,128 @@
...
+USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
+
+# Nix keys for securly verifying installer download signature per https://nixos.org/download.html#nix-verify-installation
+NIX_GPG_KEYS="B541D55301270E0BCF15CA5D8170B4726D7198DE"
+GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com:80
+keyserver hkps://keys.openpgp.org
+keyserver hkp://keyserver.pgp.com"
+
+if [ -e "/nix" ]; then
+    echo "(!) Nix is already installed! Aborting."
```
samruddhikhandale, line 22: I know this is a pattern we are using for our other Features, which probably needs to be changed.

As this is a brand new feature, can we check if the already installed version of nix is >= to the requested version ? If so, then exit else install the requested version.

src/nix/install.sh
```diff
@@ -0,0 +1,128 @@
...
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
+# Import common utils
+. ./utils.sh
+
+detect_user USERNAME
```
samruddhikhandale, line 34: Nice refactoring! 👏

samruddhikhandale: Once we do a cleanup of all the Features, it would be nice to adapt to this pattern.

samruddhikhandale (commented):
This is awesome. Thank you! 👏 

Left some comments.

src/nix/install.sh
Chuxel: I don't think that is a good idea in this case. You can't have multiple versions of nix installed like you could something like node. 

Installing from scratch wouldn't work without impacting was was already present. You likely would need to delete the nix folder in its entirety along with all the packages to be able to safely switch, which could have the unintended consequence of deleting something the user expects. I suspect we'd have to delete all the users created as well for the install script to succeed.

So the only other option is to try to perform an upgrade, but that is not as straight forward as it sounds and it varies based on whether the original install was multi-user or not... so we'd need to start tracking it. See https://nixos.org/manual/nix/stable/installation/upgrading.html

So, lots of complexity for questionable pay-off here.

I could skip the install but still try to install the specified packages or just fail outright if either of those are preferrable.

src/nix/install.sh
Chuxel: See prev comment

src/nix/install.sh
Chuxel: Oh - also the Nix install is like 600mb... and doing an upgrade or reinstall would land in a new layer which could ad another 600mb to the image. Generally this is something you should do once, but I could see allowing the packages/flake part to still execute.

Chuxel:
Dropped the derivation property in favor of flake given you currently can't use a .nix file from the source tree (could be added later once we have postCreateCommand support in features). Also noticed other features seem to be using comma separation for lists so adjusted to match.

src/nix/install.sh
samruddhikhandale: That's a good point and yes, it seems helpful. 👍 

My only concern with this duplication is that anyone hoping to change the `default` value of any option needs to update in these two places. ([first](https://github.com/devcontainers/features/pull/228/files/4d569087c15aba9d19f08d807d94a6ad5109c8ef#diff-def881a5fcfe8cc4e0c129257af8633f0dd6375906714c9619bb12a02c31ff6dR9) and [second](https://github.com/devcontainers/features/pull/228/files/4d569087c15aba9d19f08d807d94a6ad5109c8ef#diff-13bca258128643d15fdf43c1852e09cfaad12ceafe14818e199d4d74f36ad72aR8))

This resulted in a confusion in the past for some users, where there was a mismatch.

src/nix/install.sh
samruddhikhandale: 😮  wasn't aware of these complexities. Thanks for the detailed explanation.

> I could skip the install but still try to install the specified packages 

If we could do that, then it would be great & should sync with our existing pattern 👍  (eg. python: we install tools even if some python is already installed - details [here](https://github.com/devcontainers/features/blob/main/src/python/install.sh#L404))

Chuxel: @samruddhikhandale Ok, done!

Chuxel:
@samruddhikhandale Also I think I accidentlly deleted a comment you made when trying to delete a comment I made. 😅
image
Pretty sure it was one of the comments on the env vars, but LMK if not.  I addressed the other comments and it will now handle using a Flake or installing packages if nix is already installed.

samruddhikhandale (approved):
Looks awesome. Thank you! 👏 👏 

🚀 🚀

dsyer:
Can someone help with some more detailed instructions on how to use this? I tried adding the feature and getting a lot of errors. Different errors, no obvious pattern. This one is common:

```
[2022-11-11T11:59:37.210Z] Start: Run: tar --no-same-owner -x -f -
[2022-11-11T11:59:37.228Z] Stop (18 ms): Run: tar --no-same-owner -x -f -
[2022-11-11T11:59:37.237Z] * Processing feature: ghcr.io/devcontainers/features/nix
[2022-11-11T11:59:37.744Z] * Processing feature: ghcr.io/devcontainers/features/docker-in-docker
[2022-11-11T11:59:38.329Z] * Fetching feature: nix_1_oci
[2022-11-11T12:01:48.847Z] error: Error: connect ETIMEDOUT 185.199.111.154:443
[2022-11-11T12:01:48.847Z] (!) ERR: Failed to fetch feature: Failed to download package for ghcr.io/devcontainers/features/nix 
[2022-11-11T12:01:48.849Z] Error: Failed to download package for ghcr.io/devcontainers/features/nix
[2022-11-11T12:01:48.850Z]     at vw (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:155:6485)
[2022-11-11T12:01:48.850Z]     at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2022-11-11T12:01:48.850Z]     at async yS (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:186:6943)
[2022-11-11T12:01:48.851Z]     at async na (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:186:2444)
[2022-11-11T12:01:48.851Z]     at async oo (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:186:12686)
[2022-11-11T12:01:48.852Z]     at async sa (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:186:11651)
[2022-11-11T12:01:48.852Z]     at async bu (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:254:2847)
[2022-11-11T12:01:48.852Z]     at async Jw (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:254:1779)
[2022-11-11T12:01:48.853Z]     at async JS (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:260:2006)
[2022-11-11T12:01:48.853Z]     at async co (/home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js:260:3110)
[2022-11-11T12:01:48.857Z] Stop (131867 ms): Run: /nix/store/nrdmbm43k7xyc9nvg741zcqpi7i72rvn-vscode-1.70.2/lib/vscode/code /home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js up --user-data-folder /home/dsyer/.config/Code/User/globalStorage/ms-vscode-remote.remote-containers/data --workspace-folder /home/dsyer/dev/scratch/hello-as --workspace-mount-consistency cached --id-label devcontainer.local_folder=/home/dsyer/dev/scratch/hello-as --log-level debug --log-format json --config /home/dsyer/dev/scratch/hello-as/.devcontainer/devcontainer.json --default-user-env-probe loginInteractiveShell --mount type=volume,source=vscode,target=/vscode,external=true --skip-post-create --update-remote-user-uid-default on --mount-workspace-git-root true
[2022-11-11T12:01:48.858Z] Exit code 1
[2022-11-11T12:01:48.871Z] Command failed: /nix/store/nrdmbm43k7xyc9nvg741zcqpi7i72rvn-vscode-1.70.2/lib/vscode/code /home/dsyer/.vscode/extensions/ms-vscode-remote.remote-containers-0.247.0/dist/spec-node/devContainersSpecCLI.js up --user-data-folder /home/dsyer/.config/Code/User/globalStorage/ms-vscode-remote.remote-containers/data --workspace-folder /home/dsyer/dev/scratch/hello-as --workspace-mount-consistency cached --id-label devcontainer.local_folder=/home/dsyer/dev/scratch/hello-as --log-level debug --log-format json --config /home/dsyer/dev/scratch/hello-as/.devcontainer/devcontainer.json --default-user-env-probe loginInteractiveShell --mount type=volume,source=vscode,target=/vscode,external=true --skip-post-create --update-remote-user-uid-default on --mount-workspace-git-root true
[2022-11-11T12:01:48.871Z] Exit code 1
```

Chuxel:
@dsyer Would you mind raising an issue with your specifics?  In particular the base image you are using, OS, architecture, etc.  Possible there's a scenario where things are not working as expected. That particular error though sounds like its actually having a hard time downloading the feature itself which could be connectivity related. #275 is very specific to the universal image, for example.

dsyer:
That's it really, for the specifics. The JSON is just the default if you add ubuntu and nix. My network connectivity is generally good here locally. Maybe there was a DNS glitch out there somewhere? 185.199.111.154 looks like Github (not gcr.io).

Chuxel:
@dsyer Raised https://github.com/devcontainers/cli/issues/281 for you. Folks won't reply to a merged PR.

PR #235, opened 2022-10-24
github-cli: Remove duplicated find_version_from_git_tags ()

aureliojargas:
This function was defined twice in the install script for GitHub CLI, in old lines 76 and 161. This commit removes the first definition.

The only function calls were on old lines 201 and 222. Since they appear after the second function definition at line 161, the first one at line 76 never got called, so it's safe to remove it.

I confirmed that this was the only file in this repository where this function was duplicated:  

```console
$ git grep '^ *find_version_from_git_tags *('
src/docker-from-docker/install.sh:find_version_from_git_tags() {
src/docker-in-docker/install.sh:find_version_from_git_tags() {
src/git-lfs/install.sh:find_version_from_git_tags() {
src/github-cli/install.sh:find_version_from_git_tags() {
src/github-cli/install.sh:find_version_from_git_tags() {
src/go/install.sh:find_version_from_git_tags() {
src/kubectl-helm-minikube/install.sh:find_version_from_git_tags() {
src/php/install.sh:find_version_from_git_tags() {
src/powershell/install.sh:find_version_from_git_tags() {
src/python/install.sh:find_version_from_git_tags() {
src/ruby/install.sh:find_version_from_git_tags() {
src/rust/install.sh:find_version_from_git_tags() {
src/terraform/install.sh:find_version_from_git_tags() {
$
```

### DIFF
--- a/src/github-cli/install.sh
+++ b/src/github-cli/install.sh
@@ -72,40 +72,6 @@ receive_gpg_keys() {
     fi
 }
 
-# Figure out correct version of a three part version number is not passed
-find_version_from_git_tags() {
-    local variable_name=$1
-    local requested_version=${!variable_name}
-    if [ "${requested_version}" = "none" ]; then return; fi
-    local repository=$2
-    local prefix=${3:-"tags/v"}
-    local separator=${4:-"."}
-    local last_part_optional=${5:-"false"}    
-    if [ "$(echo "${requested_version}" | grep -o "." | wc -l)" != "2" ]; then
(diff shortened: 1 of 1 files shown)


aureliojargas:

bamurtaugh:
@samruddhikhandale @joshspicer this seems reasonable to me, what about you?

jkeech (approved):
Thanks for cleaning this up! Looks like it was accidentally duplicated in a previous refactoring.

PR #238, opened 2022-10-26
Install golangci-lint from prebuilt binaries and allow specifying its version

ItalyPaleAle:
golangci-lint (a linter) should not be installed using `go get`/`go install`, per [official guidance](https://golangci-lint.run/usage/install/#install-from-source).

This PR changes the way the linter is installed, using the official install script that fetches pre-compiled binaries.

Additionally, adds a new option to specify the version of the linter. Many projects (incl. our Dapr) really need to be able to specify the version of the linter.

### DIFF
--- a/src/go/devcontainer-feature.json
+++ b/src/go/devcontainer-feature.json
@@ -15,6 +15,14 @@
             ],
             "default": "latest",
             "description": "Select or enter a Go version to install"
+        },
+        "golangciLintVersion": {
+            "type": "string",
+            "proposals": [
+                "latest"
+            ],
+            "default": "latest",
+            "description": "Select or enter a golangci-lint version to install"
         }
     },
     "init": true,
--- a/src/go/devcontainer-feature.json
+++ b/src/go/devcontainer-feature.json
@@ -19,7 +19,7 @@
         "golangciLintVersion": {
             "type": "string",
             "default": "latest",
-            "description": "Enter a golangci-lint version to install or 'latest'"
+            "description": "Version of golangci-lint to install"
         }
     },
     "init": true,
--- a/src/go/devcontainer-feature.json
+++ b/src/go/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "go",
-    "version": "1.0.9",
+    "version": "1.1.0",
     "name": "Go",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/go",
     "description": "Installs Go and common Go utilities. Auto-detects latest version and installs needed dependencies.",
@@ -18,11 +18,8 @@
         },
         "golangciLintVersion": {
             "type": "string",
-            "proposals": [
-                "latest"
-            ],
             "default": "latest",
-            "description": "Select or enter a golangci-lint version to install"
+            "description": "Enter a golangci-lint version to install or 'latest'"
         }
     },
     "init": true,
--- a/src/go/install.sh
+++ b/src/go/install.sh
@@ -8,6 +8,7 @@
 # Maintainer: The VS Code and Codespaces Teams
 
 TARGET_GO_VERSION=${VERSION:-"latest"}
+GOLANGCILINT_VERSION=${GOLANGCILINTVERSION:-"latest"}
 
 TARGET_GOROOT=${TARGET_GOROOT:-"/usr/local/go"}
 TARGET_GOPATH=${TARGET_GOPATH:-"/go"}
@@ -197,8 +198,7 @@ GO_TOOLS="\
     github.com/mgechev/revive@latest \
     github.com/uudashr/gopkgs/v2/cmd/gopkgs@latest \
     github.com/ramya-rao-a/go-outline@latest \
-    github.com/go-delve/delve/cmd/dlv@latest \
-    github.com/golangci/golangci-lint/cmd/golangci-lint@latest"
+    github.com/go-delve/delve/cmd/dlv@latest"
 if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
     echo "Installing common Go tools..."
     export PATH=${TARGET_GOROOT}/bin:${PATH}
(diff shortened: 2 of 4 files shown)


bamurtaugh:
Thanks @ItalyPaleAle, this sounds reasonable to me!

@samruddhikhandale @joshspicer any thoughts?

src/go/devcontainer-feature.json
```diff
@@ -15,6 +15,14 @@
             ],
             "default": "latest",
             "description": "Select or enter a Go version to install"
+        },
+        "golangciLintVersion": {
+            "type": "string",
+            "proposals": [
```
samruddhikhandale, line 21: Could we add any other `proposals`? , if not, then feel free to drop this property.

src/go/devcontainer-feature.json
```diff
@@ -15,6 +15,14 @@
             ],
```
samruddhikhandale, line 15: Can we bump the minor version in here as we are adding a new `option`? (ie. `1.1.0`) 
https://github.com/devcontainers/features/blob/69743e604e5c9ea03eb7d88fef1676b4324549e9/src/go/devcontainer-feature.json#L3
The Go feature will automatically get published once this PR is merged. (Provided the version has changed)

src/go/install.sh
```diff
@@ -219,8 +219,18 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
...
     # Move Go tools into path and clean up
     mv /tmp/gotools/bin/* ${TARGET_GOPATH}/bin/
-
     rm -rf /tmp/gotools
+
+    # Install golangci-lint from precompiled binares
+    if [ "$GOLANGCILINT_VERSION" = "latest" ] || [ "$GOLANGCILINT_VERSION" = "" ]; then
+        echo "Installing golangci-lint latest..."
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+            sh -s -- -b "/usr/local/bin"
```
samruddhikhandale, line 228: I suppose this should point to the `GO` path? 

```suggestion
            sh -s -- -b "/usr/local/go/bin"
```

src/go/install.sh
```diff
@@ -219,8 +219,18 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
...
+    else
+        echo "Installing golangci-lint ${GOLANGCILINT_VERSION}..."
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+            sh -s -- -b "/usr/local/bin" "v${GOLANGCILINT_VERSION}"
```
samruddhikhandale, line 232: Same here

```suggestion
            sh -s -- -b "/usr/local/go/bin" "v${GOLANGCILINT_VERSION}"
```

samruddhikhandale (commented):
Thanks for the contribution. ✨ 

Left some comments.

src/go/devcontainer-feature.json
ItalyPaleAle: The linter gets upgraded too frequently to list proposed versions IMHO. I'll just remove the property

src/go/install.sh
ItalyPaleAle: I would say no, because `/usr/local/go/bin` is for binaries that were installed by `go install`. Technically this is not installed through that, so I think keeping it oustide of the gopath makes most sense.

src/go/devcontainer-feature.json
ItalyPaleAle: Updated

src/go/devcontainer-feature.json
```diff
@@ -15,6 +15,14 @@
...
             "default": "latest",
             "description": "Select or enter a Go version to install"
+        },
+        "golangciLintVersion": {
+            "type": "string",
+            "proposals": [
+                "latest"
+            ],
+            "default": "latest",
+            "description": "Select or enter a golangci-lint version to install"
```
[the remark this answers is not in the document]
ItalyPaleAle: Updated

src/go/install.sh
samruddhikhandale: Your point makes sense to me, however, I was trying to be consistent with the doc pointed out in the PR description. It seems to install in `GO` path which seems appropriate to me - having all `go` utils in one place. 

https://golangci-lint.run/usage/install/#linux-and-windows

src/go/devcontainer-feature.json
```diff
@@ -15,6 +15,11 @@
             ],
             "default": "latest",
             "description": "Select or enter a Go version to install"
+        },
+        "golangciLintVersion": {
+            "type": "string",
+            "default": "latest",
+            "description": "Enter a golangci-lint version to install or 'latest'"
```
samruddhikhandale, line 22:
```suggestion
            "description": "Version of golangci-lint to install"
```

src/go/install.sh
ItalyPaleAle: Ok. In that case, it should probably be `${TARGET_GOPATH}/bin`. Updated

src/go/install.sh
```diff
@@ -219,8 +219,18 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
...
     # Move Go tools into path and clean up
     mv /tmp/gotools/bin/* ${TARGET_GOPATH}/bin/
-
     rm -rf /tmp/gotools
+
+    # Install golangci-lint from precompiled binares
+    if [ "$GOLANGCILINT_VERSION" = "latest" ] || [ "$GOLANGCILINT_VERSION" = "" ]; then
+        echo "Installing golangci-lint latest..."
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+            sh -s -- -b "${TARGET_GOPATH}/bin"
```
samruddhikhandale, line 228: `TARGET_GOROOT: /usr/local/go` whereas `TARGET_GOPATH: /go`

```suggestion
            sh -s -- -b "${TARGET_GOROOT}/bin"
```

src/go/install.sh
```diff
@@ -219,8 +219,18 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
...
+    else
+        echo "Installing golangci-lint ${GOLANGCILINT_VERSION}..."
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+            sh -s -- -b "${TARGET_GOPATH}/bin" "v${GOLANGCILINT_VERSION}"
```
samruddhikhandale, line 232:
```suggestion
            sh -s -- -b "${TARGET_GOROOT}/bin" "v${GOLANGCILINT_VERSION}"
```

src/go/install.sh
ItalyPaleAle: Why GOROOT? The Go tools are copied into GOPATH: https://github.com/devcontainers/features/pull/238/files#diff-8d791f705e135e8d78a36c8218fd50d0d199af1f72c8df7721ad34f8865b627fR221

samruddhikhandale: I thought the copying over happens later in the code. But doesn't look like it.

Feel free to ignore the comment.

samruddhikhandale (approved):
Looks great 👏

PR #239, opened 2022-10-26
Fix oryx bug - Add PIP_CACHE_DIR

samruddhikhandale:
Creates a  "/usr/local/share/pip-cache/lib" directory which is used by the oryx tool to cache installed python packages automatically from `requirements.txt`

Fixes bug 

image


jungaretti (approved):
Looks good, thanks for fixing this.

PR #240, opened 2022-10-26
(GH-198) Add `extended` option to hugo feature

### LINKED ISSUE #198: Hugo: Add option for extended release
In addition to version, operating system, and architecture the hugo release packages are distinguished by whether or not they're "extended," which enables post-processing behavior and using SASS/SCSS. Numerous hugo themes rely on this functionality and are currently unusable with the feature, which is only capable of installing non-extended hugo.

The current definition for options is:

https://github.com/devcontainers/features/blob/82960fd16b04fb4003a8765b2381bd23553737b5/src/hugo/devcontainer-feature.json#L6-L15

I propose it be updated to this:

```json
"options": {
  "version": {
    "type": "string",
    "proposals": [
      "latest"
    ],
    "default": "latest",
    "description": "Select or enter a version."
  },
  "extended": {
    "type": "boolean",
    "default": false,
    "description": "Choose whether to install hugo extended."
  }
}
```

Then, instead of:

https://github.com/devcontainers/features/blob/82960fd16b04fb4003a8765b2381bd23553737b5/src/hugo/install.sh#L97-L104

It would be:

```sh
# Install ARM or x86 version of hugo based on current machine architecture
if [ "$(uname -m)" == "aarch64" ]; then
    arch="ARM64"
else
    arch="64bit"
fi

# Install extended version of hugo if desired
if [ "${EXTENDED}" = "true" ]; then
    extended="extended_"
else
    extended=""
fi

hugo_filename="hugo_${extended}${VERSION}_Linux-${arch}.tar.gz"
```

michaeltlombardi:
Prior to this change, the hugo feature only supported specifying the version of Hugo to install. Hugo is available in two builds: standard, which the feature already installs, and extended, which includes functionality for post-processing CSS/SCSS/SASS and JavaScript.

This change adds a new `extended` option to the hugo feature, allowing users to specify that they require the extended build of Hugo. It defaults to `false` and installs the standard build of Hugo.

- Resolves #198

### DIFF
--- a/src/hugo/devcontainer-feature.json
+++ b/src/hugo/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "hugo",
-  "version": "1.0.6",
+  "version": "1.1.0",
   "name": "Hugo",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/hugo",
   "options": {
@@ -11,6 +11,11 @@
       ],
       "default": "latest",
       "description": "Select or enter a version."
+    },
+    "extended": {
+      "type": "boolean",
+      "default": false,
+      "description": "Install Hugo extended for SASS/SCSS changes"
     }
   },
   "containerEnv": {
--- a/src/hugo/install.sh
+++ b/src/hugo/install.sh
@@ -106,7 +106,14 @@ if ! hugo version &> /dev/null ; then
         arch="64bit"
     fi
 
-    hugo_filename="hugo_${VERSION}_Linux-${arch}.tar.gz"
+    # Install extended version of hugo if desired
+    if [ "${EXTENDED}" = "true" ]; then
+        extended="extended_"
+    else
+        extended=""
+    fi
+
+    hugo_filename="hugo_${extended}${VERSION}_Linux-${arch}.tar.gz"
 
     curl -fsSLO --compressed "https://github.com/gohugoio/hugo/releases/download/v${VERSION}/${hugo_filename}"
     tar -xzf "$hugo_filename" -C "$installation_dir"
--- a/test/hugo/install_hugo_extended.sh
+++ b/test/hugo/install_hugo_extended.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Ensure extended version is installed
+check "extended_installed"  bash -c "hugo version | grep extended"
+
+# Report result
+reportResults
--- a/test/hugo/scenarios.json
+++ b/test/hugo/scenarios.json
@@ -1,0 +1,11 @@
+{
+    "install_hugo_extended": {
+        "image": "mcr.microsoft.com/devcontainers/base",
+        "features": {
+            "hugo": {
+                "version": "latest",
+                "extended": true
+            }
+        }
+    }
+}


src/hugo/install.sh
```diff
@@ -106,7 +106,14 @@ if ! hugo version &> /dev/null ; then
...
 
-    hugo_filename="hugo_${VERSION}_Linux-${arch}.tar.gz"
+    # Install extended version of hugo if desired
+    if [ "${EXTENDED}" = "true" ]; then
+        extended="extended_"
+    else
+        extended=""
+    fi
+
+    hugo_filename="hugo_${extended}${VERSION}_Linux-${arch}.tar.gz"
```
samruddhikhandale, line 116: Could we add a test (`scenarios.json`) for this? You could use https://github.com/devcontainers/features/tree/main/test/node for reference.

Let me know if you have any questions, happy to walk you through it.

samruddhikhandale (commented):
Thanks for the contribution. 👏 
Left some comments.

src/hugo/install.sh
samruddhikhandale: Ideally, in https://github.com/devcontainers/features/tree/main/test/hugo

You'd need to create `scenarios.json` with 

```
{
    "install_extended_hugo": {
        "image": "mcr.microsoft.com/devcontainers/base",
        "features": {
            "hugo": {
                "version": "latest",
                "extended": true
            }
        }
    }
}
```

Then create a `install_extended_hugo.sh` file with a test which ensure that hugo's extended version is installed.

src/hugo/README.md
```diff
@@ -18,6 +18,7 @@
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Select or enter a version. | string | latest |
+| extended | Choose whether to install hugo extended. | boolean | false |
```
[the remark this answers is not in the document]
michaeltlombardi: Definitely, good shout! 😅

src/hugo/devcontainer-feature.json
```diff
@@ -11,6 +11,11 @@
       ],
```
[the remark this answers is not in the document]
michaeltlombardi: Great point, this should definitely get bumped, will do! Sorry for missing this. 😔

joshspicer:
Hey @michaeltlombardi! If there's anything I can help with to get this PR over the line let me know :) Namely, @samruddhikhandale 's suggestion of using our testing framework to write a "scenario" test.

I recently wrote up some [docs on the process here](https://github.com/devcontainers/cli/blob/main/docs/features/test.md), and additionally there are many examples in this repo.

michaeltlombardi:
@joshspicer, @samruddhikhandale - sorry for the late reply! I got sucked into other things, but I've got the PR fixed up with a pass at tests added. Looks like it's waiting on approval to run the workflows. I haven't validated the tests locally, but I'm setting up my environment to do so and will report back if the automated tests don't beat me to the results.

test/hugo/install_hugo_extended.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Ensure extended version is installed
+check "extended_installed"  hugo version | grep extended
```
samruddhikhandale, line 9:
```suggestion
check "extended_installed"  bash -c "hugo version | grep extended"
```
I understand that you've added a similar test as the sample. However, looks like it is not working as expected. I'll look at fixing others.

image

☝️  The tests are passing even though the output doesn't contain `extended`.

michaeltlombardi: I wasn't able to get a devcontainer test environment up locally (WSL seemed to hang on the upgrade and I don't have a license for docker desktop), but I'll add this suggestion in the meantime.

michaeltlombardi: It seems that [fixed the test](https://github.com/devcontainers/features/actions/runs/3422593175/jobs/5700189310#step:4:212).

samruddhikhandale (approved):
Looks great 🚀 

Thank you for the contribution! ✨

michaeltlombardi:
@samruddhikhandale thanks to everyone for your help getting this over the line! Hugely appreciated, this is going to make a big difference for contributors to my projects. 😊

PR #241, opened 2022-10-28
Fix privs in Node, regressions from #199

Chuxel:
Privs in the node script are completely broken which prevents its use with Actions or any scenario where the user's UID/GID does not match what is in the container.  There were serveral issues here:

1. There was a bug in nvm that incorrectly set the user on installed node versions. Bumped to 0.39.2 to fix that and added an option to pick a node version while I was at it.

2. #199 set the sticky bit in the wrong place which prevented group privs from going to the correct locations.

3. The updates to add addtional versions sufffered from similar issues before #199 which was an attempt to fix it.

3. The existing test was busted since Node 19 came out and it was looking for Node 18 with "latest".  Switched to "lts" which is node 18 (and eventually 20... so an update will be needed.)

This required a bit of refactoring and I updated comments to explain the changes.

### DIFF
--- a/test/node/scenarios.json
+++ b/test/node/scenarios.json
@@ -3,8 +3,17 @@
         "image": "debian:11",
         "features": {
             "node": {
-                "version": "latest",
+                "version": "lts",
                 "additionalVersions": "v17.9.1,v14.19.3"
+            }
+        }
+    },
+    "non_root_user": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base",
+        "remoteUser": "vscode",
+        "features": {
+            "node": {
+                "version": "latest"
             }
         }
     }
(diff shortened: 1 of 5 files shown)


test/node/scenarios.json
```diff
@@ -3,9 +3,18 @@
...
         "features": {
             "node": {
-                "version": "latest",
+                "version": "lts",
                 "additionalVersions": "v17.9.1,v14.19.3"
             }
         }
+    },
+    "non_root_user": {
+        "image": "mcr.microsoft.com/vscode/devcontainers/base",
```
samruddhikhandale, line 12:
```suggestion
        "image": "mcr.microsoft.com/devcontainers/base",
```

samruddhikhandale (dismissed):
Looks good to me. Approving with one suggestion.

samruddhikhandale (approved):
Awesome. Thanks 🚀

Issue #226, opened 2022-10-18
Adding jupyterlab to python:3.10-bullseye image


natescherer:
Hello! Not really sure if this is an issue for here for `devcontainers/images`, but I'm having some trouble figuring out how to add jupyterlab to the `mcr.microsoft.com/devcontainers/python:3.10-bullseye` image using the new features. This worked fine before when jupyterlab was its own feature, but now that it's been moved into the overall python feature I haven't been able to get it to work. I tried this, but it errors out:

```
	"features": {
		"ghcr.io/devcontainers/features/python:1": {
			"version": "none",
			"installTools": false,
			"installJupyterlab": true
		}
	}
```
Error
```
  #32 [linux/amd64 dev_containers_target_stage 7/8] RUN cd /tmp/build-features/python_2 && chmod +x ./devcontainer-features-install.sh && ./devcontainer-features-install.sh
  #0 0.062 ===========================================================================
  #0 0.062 Feature       : Python 
  [333840 ms] #32 0.065 Description   : Installs the provided version of Python, as well as PIPX, and other common Python utilities.  JupyterLab is conditionally installed with the python feature. Note: May require source code compilation.
  #32 0.065 Id            : ghcr.io/devcontainers/features/python
  #32 0.065 Version       : 1.0.9
  #32 0.065 Documentation : https://github.com/devcontainers/features/tree/main/src/python
  #32 0.065 Options       :
  #32 0.066     VERSION="none"
  #32 0.066     INSTALLTOOLS="false"
  #32 0.066     OPTIMIZE="false"
  #32 0.066     INSTALLPATH="/usr/local/python"
  #32 0.066     INSTALLJUPYTERLAB="true"
  #32 0.066     CONFIGUREJUPYTERLABALLOWORIGIN=""
  #32 0.066 ===========================================================================
  #32 0.118 -bash: line 1: /bin/python3: No such file or directory
  #32 0.120 ERROR: Feature "Python" (ghcr.io/devcontainers/features/python) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/python for help troubleshooting this error.
  #32 ERROR: process "/bin/bash -c cd /tmp/build-features/python_2 && chmod +x ./devcontainer-features-install.sh && ./devcontainer-features-install.sh" did not complete successfully: exit code: 127
```

Any recommendations on how to do this?

bamurtaugh:
@samruddhikhandale any thoughts / recs here?

[...]

PR #244, opened 2022-10-28
Python bug fix - Add jupyterlab even if python is not installed by Feature but exists

samruddhikhandale:
Fixes - https://github.com/devcontainers/features/issues/226
The python feature was failing to add `jupyterlab` if python was not installed by the Feature. Fixes that.


src/python/install.sh
```diff
@@ -449,12 +453,17 @@ fi
...
-    if [ -n "${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}" ]; then
-        add_user_jupyter_config "c.ServerApp.allow_origin = '${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}'"
-        add_user_jupyter_config "c.NotebookApp.allow_origin = '${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}'"
+        # Configure JupyterLab if needed
+        if [ -n "${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}" ]; then
+            add_user_jupyter_config "c.ServerApp.allow_origin = '${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}'"
+            add_user_jupyter_config "c.NotebookApp.allow_origin = '${CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN}'"
+        fi
+    else
+        echo "(!) Could not install jupyterlab. Python not found."
```
jungaretti, line 465:
```suggestion
        echo "(!) Could not install JupyterLab. Python not found."
```

src/python/install.sh
```diff
@@ -449,12 +453,17 @@ fi
 
 # Install JupyterLab if needed
 if [ "${INSTALL_JUPYTERLAB}" = "true" ]; then
-    install_user_package jupyterlab
+    if [[ $(python --version) != "" ]]; then
```
jungaretti, line 456: Can you use the binary at `$PYTHON_SRC` for the check here instead of `which python`? `install_user_package` will use the binary at `$PYTHON_SRC`

jungaretti (changes requested):

src/python/install.sh
samruddhikhandale: Good point.

src/python/install.sh
```diff
@@ -449,12 +453,17 @@ fi
 
 # Install JupyterLab if needed
 if [ "${INSTALL_JUPYTERLAB}" = "true" ]; then
-    install_user_package jupyterlab
+    if [ ! -z "${PYTHON_SRC}" ]; then
```
jungaretti, line 456: Another approach here is a guard condition:

```bash
if [ -z "${PYTHON_SRC}" ]; then
    echo "(!) Could not install JupyterLab. Python not found."
    exit 1
fi

install_user_package jupyterlab

# Configure JupyterLab if needed
```

This keeps the error close to the condition and saves us some indenting

jungaretti (dismissed):
Thanks!

joshspicer (dismissed):
If @jungaretti likes it, I like it too!

src/python/install.sh
```diff
@@ -408,11 +408,15 @@ if [ "${PYTHON_VERSION}" != "none" ]; then
         updaterc "if [[ \"\${PATH}\" != *\"${CURRENT_PATH}/bin\"* ]]; then export PATH=${CURRENT_PATH}/bin:\${PATH}; fi"
         PATH="${INSTALL_PATH}/bin:${PATH}"
     fi
-
+
```
jungaretti, line 411: Nit: Added whitespace

jungaretti (approved):

PR #249, opened 2022-10-30
Use asc format instead of gpg for not to install gnupg2

eitsupi:
This change may reduce the installation of unnecessary packages.

### DIFF
--- a/src/conda/install.sh
+++ b/src/conda/install.sh
@@ -69,12 +69,12 @@ if ! conda --version &> /dev/null ; then
     usermod -a -G conda "${USERNAME}"
 
     # Install dependencies
-    check_packages curl ca-certificates gnupg2
+    check_packages curl ca-certificates
 
     echo "Installing Conda..."
 
-    curl -sS https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc | gpg --dearmor > /usr/share/keyrings/conda-archive-keyring.gpg
-    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/conda-archive-keyring.gpg] https://repo.anaconda.com/pkgs/misc/debrepo/conda stable main" > /etc/apt/sources.list.d/conda.list
+    curl -sS https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc | tee -a /usr/share/keyrings/conda-archive-keyring.asc
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/conda-archive-keyring.asc] https://repo.anaconda.com/pkgs/misc/debrepo/conda stable main" > /etc/apt/sources.list.d/conda.list
     apt-get update -y
 
     CONDA_PKG="conda=${VERSION}-0"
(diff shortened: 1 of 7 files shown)


samruddhikhandale:
Hi @eitsupi could you sync the PR with `main`? It should help fix oryx changes.

eitsupi:
> Hi @eitsupi could you sync the PR with `main`? It should help fix oryx tests.

I did, but the test does not run automatically, so I do not know if it was resolved or not.

I know I'm not a first-time contributor, but is it intentional that non-maintainers always require the approval of the maintainer to run tests?
(I recall a similar change was accidentally made at Apache some time ago, but was quickly restored.)

eitsupi:
Thank you for running the tests.
Does it appear to be only test failures unrelated to this update?

src/conda/install.sh
```diff
@@ -69,12 +69,12 @@ if ! conda --version &> /dev/null ; then
     usermod -a -G conda "${USERNAME}"
 
     # Install dependencies
-    check_packages curl ca-certificates gnupg2
+    check_packages curl ca-certificates
 
     echo "Installing Conda..."
 
-    curl -sS https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc | gpg --dearmor > /usr/share/keyrings/conda-archive-keyring.gpg
```
samruddhikhandale, line 76: I think storing keyring as `.gpg` is intentional [details](https://wiki.debian.org/DebianRepository/UseThirdParty#OpenPGP_key__handling).  

image

`.gpg` uses a binary format and `.asc` an ASCII-armored signature with or without wrapped document, plain text format

@Chuxel could you help confirm if that's the case?

samruddhikhandale:
> I know I'm not a first-time contributor, but is it intentional that non-maintainers always require the approval of the maintainer to run tests?

Yes, looks intentional - it is made restrictive to avoid GitHub Actions workflow's modifications for abusive purposes. 😅

Chuxel:
@samruddhikhandale The failures here look like they are related to other issues.  Namely:

1. PHP seems to try to install a package not available on Ubuntu bionic.  It's likely there's another similar package required... or PHP has dropped support for it. You'd need to check.
2. Debian 9 is out of support, so we've since dropped those tests.

That said, yes, we're following guidance from Debian here. At one point `asc` failed completely - so it's interesting it now appears to be working. Something may have changed that the wiki does not reflect yet - but it is certainly not what people are being guided to do.

Looks like gnupg2 is a ~19mb addition to `debian` with nothing else added. Given the usefulness of gpg and the unclear nature of whether `asc` is really officially supported, we may want to hold off on switching. I'd love to know if the guidance here as changed - but haven't found anything so far. Anyone else seen anything?

eitsupi:
Ok, it seems there is little value in merging this, so I will close it.

eitsupi:
> Looks like gnupg2 is a ~19mb addition to `debian` with nothing else added. Given the usefulness of gpg and the unclear nature of whether `asc` is really officially supported, we may want to hold off on switching. I'd love to know if the guidance here as changed - but haven't found anything so far. Anyone else seen anything?

@Chuxel FYI

> bullseye is the final Debian release to ship apt-key. Keys should be managed by dropping files into /etc/apt/trusted.gpg.d instead, in binary format as created by gpg --export with a .gpg extension, or ASCII armored with a .asc extension. 

https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#deprecated-components

Issue #266, opened 2022-11-10
CUDA feature example shows wrong version option


jungaretti:
[The CUDA feature's description reads `"version": "latest"` instead of `"cudaVersion": "11.7"`](https://github.com/devcontainers/features/tree/main/src/nvidia-cuda) (the latest supported version). This description is misleading and confusing. Here are some ideas for cleaning up the feature:

- [x] https://github.com/devcontainers/features/pull/267
- [x] Fix the usage example OR handle a `latest` version

PR #267, opened 2022-11-10
Add support for CUDA 11.8 in nvidia-cuda

jungaretti:
Related: https://github.com/devcontainers/features/issues/266

Changes

- Adds support for the latest versions of CUDA (11.8) and cuDNN (8.5.0.96).
- Moves the cuDNN version check to the cuDNN install block. If the feature isn't installing cuDNN, then there's no need to check for a valid cuDNN version. We only need to check for a valid cuDNN version before we install it.

### DIFF
--- a/src/nvidia-cuda/devcontainer-feature.json
+++ b/src/nvidia-cuda/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "nvidia-cuda",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "name": "NVIDIA CUDA",
   "description": "Installs shared libraries for NVIDIA CUDA.",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/nvidia-cuda",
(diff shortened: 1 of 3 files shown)


src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -18,19 +18,21 @@
     "cudaVersion": {
       "type": "string",
       "enum": [
+        "11.8",
```
samruddhikhandale, line 21: Could we add a scenario (aka test) to confirm that the new values work as expected?

KassoThaArt: Yes

jungaretti: Added! How about a test in `test.sh`? This feature doesn't support `latest` for `cudaVersion`.

jungaretti: Thanks for your feedback @HiddenLeafVill

joshspicer (approved):

PR #268, opened 2022-11-10
Rename cudaVersion option to version

jungaretti:
Related: https://github.com/devcontainers/features/issues/266

Changes

- Renames the `cudaVersion` option to `version` without breaking the `cudaVersion` route.

History

I named this option `cudaVersion` instead of `version` because it doesn't support `latest`. Unfortunately, our auto-generated Example Usage doesn't understand this and recommends `"version": "latest"`: https://github.com/devcontainers/features/tree/main/src/nvidia-cuda

### DIFF
--- a/src/nvidia-cuda/devcontainer-feature.json
+++ b/src/nvidia-cuda/devcontainer-feature.json
@@ -15,7 +15,7 @@
       "default": false,
       "description": "Additionally install NVIDIA Tools Extension (NVTX)"
     },
-    "cudaVersion": {
+    "version": {
       "type": "string",
       "enum": [
         "11.7",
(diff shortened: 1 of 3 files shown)


src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -15,7 +15,7 @@
       "default": false,
       "description": "Additionally install NVIDIA Tools Extension (NVTX)"
     },
-    "cudaVersion": {
```
samruddhikhandale, line 18: I know this is still a draft PR. However, do we really want to rename the option? Renaming options seems breaking/unexpected change and we don't have back compat setup. Isn't description sufficient to explain the same?

However, if we still want to rename the option, then we'd like to bump the major version.

PR #269, opened 2022-11-10
Add .NET 7, drop .NET 5

Chuxel:
.NET 7 is now out and .NET 5 went out of support back in May '22.


src/dotnet/devcontainer-feature.json
```diff
@@ -9,8 +9,8 @@
             "type": "string",
             "proposals": [
                 "latest",
+                "7",
```
samruddhikhandale, line 12: For `latest` looks like we have hardcoded to `6`. Should we update that as well? 

https://github.com/devcontainers/features/blob/285097064cf58bd228e1a231421ccdf9feb894f4/src/dotnet/install.sh#L213

If we decide to do so, then can we bump the minor version instead? 😄

Chuxel:
Related to https://github.com/devcontainers/images/issues/200

src/dotnet/devcontainer-feature.json
Chuxel: Ugh - yeah that needs to be updated.  We also shouldn't really hard code this - but that's better for a different PR.

Chuxel: Looks like the tests are also going to return false positives since the `| grep ...` would apply to the output of the test command. Will fix that.

Chuxel:
Also looks like libssl1.1 is not available on Ubuntu Jammy, so we should be installing libssl3. There was a bug in the .NET script that had the wrong package name, and then common script is missing adding libssl3 as the replacement for 1.1. Will fix that too.

Chuxel:
Okay - this required quite a bit of refactoring to get to work since the regression tests showed a number of problems.  So, here's what has been done:

1.  The apt-based install process now handles both native and non-native packages along with sub-versions. This was broken somewhere along the way. Furthermore, the version in the OS is more fixed than the version in the MS repo, so we need to adapt to that.  This actually simplified the total code since it was already set up to do this... package name detection was present before.
2. The symlink to "current" was also getting messed up which could result in updates still pointing to the old location.
3. I removed hard coding DOTNET_ROOT from `devcontainer-features.json` since there can be more than one location on disk. These days dotnet itself should set this well enough.

TBH - Looking at the complexity now that there's two potential apt sources PLUS the multi-install location that could result in a third possible spot.  I actually suspect it may be better to always use curl to pull down the packages rather than apt. 

I also noticed that the way the multiple versions of .NET are getting installed is probably not the way that it should be since the pattern is to put multiple SDKs under the same dotnet directory.  But again, did not change that.

I didn't quite go that far though.

src/common-utils/install.sh
```diff
@@ -140,7 +140,12 @@ if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
     if [[ ! -z $(apt-cache --names-only search ^libssl1.1$) ]]; then
         package_list="${package_list}       libssl1.1"
     fi
-    
+
+    # Install libssl3 if available
+    if [[ ! -z $(apt-cache --names-only search ^libssl3$) ]]; then
```
samruddhikhandale, line 145: Can we bump the minor version for `common-utils` Feature for the changes to take effect?

src/dotnet/devcontainer-feature.json
```diff
@@ -28,7 +28,6 @@
         }
     },
     "containerEnv": {
-        "DOTNET_ROOT": "/usr/local/dotnet/current",
```
samruddhikhandale, line 31: Curious, won't we still need this? 

> As `DOTNET_ROOT`specifies the location of the . NET runtimes, if they are not installed in the default location. (ie /usr/local/share/dotnet for linux)

src/dotnet/install.sh
```diff
@@ -194,79 +197,83 @@ apt_cache_package_and_version_soft_match() {
...
+install_using_apt() {
     local sdk_or_runtime="$1"
-    local dotnet_major_minor_version
-    export DOTNET_PACKAGE="dotnet-${sdk_or_runtime}"
-
-    # Install dependencies
-    check_packages apt-transport-https curl ca-certificates gnupg2 dirmngr
-
-    # Import key safely and import Microsoft apt repo
-    get_common_setting MICROSOFT_GPG_KEYS_URI
```
samruddhikhandale, line 206: Should we remove the `get_common_setting()` function as it's no longer needed?

test/dotnet/scenarios.json
```diff
@@ -16,11 +16,60 @@
...
+    "install_dotnet_6_focal": {
         "image": "ubuntu:focal",
         "features": {
             "dotnet": {
-                "version": "5"
+                "version": "6"
+            }
+        }
+    },
+    "install_dotnet_7_bullseye": {
```
samruddhikhandale, line 35: :info I'm sure you must be familiar, however fyi, the [test.sh](https://github.com/devcontainers/features/blob/01bbe2445f73c741aa6f0b05d93f5ecd259fab0f/test/dotnet/test.sh) file is executed for the Dotnet Feature (with default values (`7` in this case)) against these base images https://github.com/devcontainers/features/blob/main/.github/workflows/test-all.yaml#L45-L51

These dotnet scenarios for `7` are also covered under those, but I agree it's great to have them as the default could change anytime.

samruddhikhandale (approved):
This looks great, left some comments. Thank you for the changes and adding test coverage.

src/dotnet/devcontainer-feature.json
Chuxel: My understanding is that we don't since the CLI itself will figure it out.

test/dotnet/scenarios.json
Chuxel: Yep that was the idea!

src/common-utils/install.sh
Chuxel: Fixed!

src/dotnet/install.sh
Chuxel: Fixed!

src/dotnet/devcontainer-feature.json
Chuxel: Restored it just to avoid yet another change.  We can remove it later.

PR #271, opened 2022-11-10
Git: change git source location

samruddhikhandale:
When `git` is installed from source, change the location from `/usr/local/bin/git` to `/usr/bin` to match with the os provided git location.

### DIFF
--- a/src/git/devcontainer-feature.json
+++ b/src/git/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "git",
-    "version": "1.0.5",
+    "version": "1.0.6",
     "name": "Git (from source)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/git",
     "description": "Install an up-to-date version of Git, built from source as needed. Useful for when you want the latest and greatest features. Auto-detects latest stable version and installs needed dependencies.",
--- a/src/git/install.sh
+++ b/src/git/install.sh
@@ -151,7 +151,7 @@ echo "Downloading source for ${GIT_VERSION}..."
 curl -sL https://github.com/git/git/archive/v${GIT_VERSION}.tar.gz | tar -xzC /tmp 2>&1
 echo "Building..."
 cd /tmp/git-${GIT_VERSION}
-make -s prefix=/usr/local all && make -s prefix=/usr/local install 2>&1
+make -s prefix=/usr all && make -s prefix=/usr install 2>&1
 rm -rf /tmp/git-${GIT_VERSION}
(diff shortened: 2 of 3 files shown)


Chuxel:
IMPORTANT: We should not install git from source to `/usr` - this can calls all sorts of problems since the OS expects to be able to manage packages there.

samruddhikhandale:
> IMPORTANT: We should not install git from source to `/usr` - this can calls all sorts of problems since the OS expects to be able to manage packages there. Apt is the only thing that should manage the contents in this location.
> 
> You could symlink /etc/gitconfig since that's a config file, but not all of git.

Makes sense 🤔 
@jkeech any concerns about symlinking `/etc/gitconfig => /usr/local/etc/gitconfig` to fix the user issue? (instead of changing the installation location to `/usr/bin`)

jkeech:
> ℹ️ For codespaces kitchensink, we have always built `git` from source - https://github.com/microsoft/vscode-dev-containers/blob/main/containers/codespaces-linux/.devcontainer/base.Dockerfile#L64

Thanks, I confirmed on an older universal image v1.7.2 built in Jan 2022 that this is true and that the system config path in that image is `/usr/local/etc/gitconfig`.

> @jkeech any concerns about symlinking `/etc/gitconfig => /usr/local/etc/gitconfig` to fix the user issue? (instead of changing the installation location to `/usr/bin`)

I don't think we should symlink that path while continuing to install in the the local directory. If git is installed in `/usr/local/bin/git`, the system-level config file should be at `/usr/local/etc/gitconfig`. The main thing is we should try to move away from building git from source as much as possible as preferring prebuilt binaries. Of course, if we need to move to newer builds due to security patches, etc, then building from source is a good temporary solution.

samruddhikhandale:
> I don't think we should symlink that path while continuing to install in the the local directory. If git is installed in /usr/local/bin/git, the system-level config file should be at /usr/local/etc/gitconfig. The main thing is we should try to move away from building git from source as much as possible as preferring prebuilt binaries. Of course, if we need to move to newer builds due to security patches, etc, then building from source is a good temporary solution.

@jkeech Until we have prebuilt binaries for `git`, should `universal` start using git from OSS feeds (upstream git has patch for security vulnerability)? or should I start working on building prebuilt binaries?

jkeech:
>@jkeech Until we have prebuilt binaries for git, should universal start using git from OSS feeds (upstream git has patch for security vulnerability)? or should I start working on building prebuilt binaries?

Yes, we should switch back to the git package from apt feeds once they include patched versions to avoid having to build from source.

PR #276, opened 2022-11-11
Add azure-functions-core-tools

jlaundry:
This PR is to add the `azure-functions-core-tools` deb package, which is needed to develop and test Azure Functions apps in codespaces.

I've created this based on the `azure-cli` feature (most of `install.sh` is borrowed), and tested it through a feature template: https://github.com/jlaundry/devcontainers-feature (`ghcr.io/jlaundry/devcontainers-feature/azure-functions-core-tools:1`)


jlaundry:
Hey @joshspicer, @jkeech, have you had a chance to look at this please?

jkeech (changes requested):
Thanks for building and sharing this feature! ✨

The goal of this repo is not to provide a monolithic/comprehensive set of all features, but rather to provide a small set of well maintained example features that the community can use as a guide to help build other features.

Instead of adding your feature here, we'd prefer if you add your feature repo to the [index](https://github.com/devcontainers/devcontainers.github.io/blob/gh-pages/_data/collection-index.yml) so it will show up on `https://containers.dev` and in the various UIs for adding devcontainer config (commands in VS Code as well as the github.com devcontainer.json file editor). That will also give you more flexibility as the feature author to easily make changes to the feature directly within your repository without additional PR review overhead.

Let us know if you have any questions!

PR #277, opened 2022-11-12
Nix feature: allow extra configuration of /etc/nix/nix.conf

NickHu:
This would allow one to specify e.g.

```
  "features": {
    "ghcr.io/devcontainers/features/nix:1": {
      "extraNixConfig": "extra-substituters = https://foo.cachix.org,extra-trusted-public-keys = foo.cachix.org-1:bar="
    }
  }
```

so that the nix daemon can run against additional binary caches.


NickHu:

Chuxel (approved):
Great addition!  Thanks!!

PR #294, opened 2022-11-15
Node: Fix bug - version set by Feature is not default

samruddhikhandale:
Version specified by the `node` Feature should be set as default. It was failing when `node` Feature was added on top of the `universal` image.


joshspicer (approved):
Awesome!

joshspicer (approved):

Issue #293, opened 2022-11-15
dotnet CLI feature is installed with wrong permissions


salvador-barboza:
Hello devcontainers team,

I was using the dotnet CLI feature, and when running `dotnet restore` on my Codespace, it throws the following error:
```
System.UnauthorizedAccessException: Access to the path '/home/codespace/.dotnet/7.0.100.toolpath.sentinel' is denied.
 ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode, Func`4 createOpenException)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, UnixFileMode openPermissions, Int64& fileLength, UnixFileMode& filePermissions, Func`4 createOpenException)
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)
   at System.IO.File.Create(String path)
   at Microsoft.Extensions.EnvironmentAbstractions.FileWrapper.CreateEmptyFile(String path)
   at Microsoft.DotNet.Cli.Utils.FileAccessRetrier.RetryOnIOException(Action action)
   at Microsoft.DotNet.Configurer.DotnetFirstTimeUseConfigurer.Configure()
   at Microsoft.DotNet.Cli.Program.ConfigureDotNetForFirstTimeUse(IFirstTimeUseNoticeSentinel firstTimeUseNoticeSentinel, IAspNetCertificateSentinel aspNetCertificateSentinel, IFileSentinel toolPathSentinel, Boolean isDotnetBeingInvokedFromNativeInstaller, DotnetFirstRunConfiguration dotnetFirstRunConfiguration, IEnvironmentProvider environmentProvider, Dictionary`2 performanceMeasurements)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, TimeSpan startupTime, ITelemetry telemetryClient)
```

Repro steps:
[...]

PR #297, opened 2022-11-16
[dotnet] Follow symlinks during `chown` and add practical dotnet test

joshspicer:
closes https://github.com/devcontainers/features/issues/293

Adds the `-H` flag when running `chown` to follow symlinks.  

Additionally, adds a regression test validating the change fixes https://github.com/devcontainers/features/issues/293.  The regression test will attempt to run a dotnet program, which includes a nuget dependency that needs to be resolved.  

Takes advantage of:
- https://github.com/devcontainers/cli/pull/286 ~~**(⚠️ not yet released, test will fail until then)**~~
- https://github.com/devcontainers/cli/pull/265


src/dotnet/install.sh
```diff
@@ -15,7 +15,7 @@ INSTALL_USING_APT=${INSTALLUSINGAPT:-"true"}
 DOTNET_LATEST="7"
 DOTNET_LTS="6"
 
-USERNAME=${USERNAME:-"automatic"}
+USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
```
samruddhikhandale, line 18: Can we bump the version for this Feature?

src/dotnet/install.sh
```diff
@@ -452,16 +457,21 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
...
 
-    chown -R "${USERNAME}:dotnet" "${TARGET_DOTNET_ROOT}"
+    chown -H -R "${USERNAME}:dotnet" "${TARGET_DOTNET_ROOT}" # Recursively traverse (-R), also following symbolic links (-H)
     chmod -R g+r+w "${TARGET_DOTNET_ROOT}"
     find "${TARGET_DOTNET_ROOT}" -type d -print0 | xargs -n 1 -0 chmod g+s
+else
+    echo "Not changing ownership ("${USERNAME}"  <-  "${TARGET_DOTNET_ROOT}")"
 fi
 
+
```
samruddhikhandale, line 473: nit: Remove?

src/dotnet/install.sh
```diff
@@ -452,16 +457,21 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
...
+
```
samruddhikhandale, line 474: nit: Remove?

samruddhikhandale (dismissed):
Nice, approving with nit comments> 🚀

samruddhikhandale:
@joshspicer Are these changes still needed? If so, then can you help resolve the conflicts and I could help review. Thanks!

PR #302, opened 2022-11-19
Fix bad path to systemctl

Chuxel:
This prevents systemctl from being picked up if systemd is running. You need to rm -f /usr/local/bin/systemctl to fix it currently.

### DIFF
--- a/src/common-utils/install.sh
+++ b/src/common-utils/install.sh
@@ -292,7 +292,7 @@ cat << 'EOF' > /usr/local/bin/systemctl
 #!/bin/sh
 set -e
 if [ -d "/run/systemd/system" ]; then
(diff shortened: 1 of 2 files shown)


edgonmsft (approved):

src/common-utils/install.sh
```diff
@@ -292,7 +292,7 @@ cat << 'EOF' > /usr/local/bin/systemctl
 #!/bin/sh
 set -e
 if [ -d "/run/systemd/system" ]; then
-    exec /bin/systemctl/systemctl "$@"
+    exec /bin/systemctl "$@"
```
samruddhikhandale, line 295: We'd need to bump the version for this Feature 😅

Chuxel: Updated!

Issue #212, opened 2022-10-10
Go doesn't update version if go is already installed


craiglpeters:
The go feature ignores the current version, and the version option specified, when go is already installed.

## Steps to reproduce
Create a .devcontainer as below
```json
{
    "features": {
        "ghcr.io/devcontainers/features/go:1": {
            "version": "1.18"
        }
    }
}
```

Start a codespace using the universal container image which already has go installed

At the terminal note that the version of go does NOT match the option specified in the go feature
```bash
$ go version
go version go1.19.2 linux/amd64
```

Check the creation logs and note it says `Go already installed. Skipping`
```
2022-10-07 23:22:30.315Z: #13 [dev_containers_target_stage 3/3] RUN cd /tmp/build-features/go_1 && chmod +x ./devcontainer-features-install.sh && ./devcontainer-features-install.sh
2022-10-07 23:22:30.931Z: #13 0.617 ===========================================================================
2022-10-07 23:22:30.931Z: #13 0.617 Feature       : Go
2022-10-07 23:22:30.931Z: #13 0.617 Description   : Installs Go and common Go utilities. Auto-detects latest version and installs needed dependencies.
2022-10-07 23:22:30.931Z: #13 0.617 Id            : ghcr.io/devcontainers/features/go
2022-10-07 23:22:30.931Z: 
2022-10-07 23:22:30.931Z: #13 0.617 Version       : 1.0.7
2022-10-07 23:22:30.931Z: #13 0.617 Documentation : https://github.com/devcontainers/features/tree/main/src/go
2022-10-07 23:22:30.931Z: #13 0.617 Options       :
2022-10-07 23:22:30.931Z: #13 0.617     VERSION="1.18"
2022-10-07 23:22:30.931Z: #13 0.617 ===========================================================================
2022-10-07 23:22:32.722Z: #13 2.416 TARGET_GO_VERSION=1.18.7
2022-10-07 23:22:32.861Z: #13 2.573 Go already installed. Skipping.
```

## Expected result
At terminal
```bash
$ go version
go version go1.18.x linux/amd64
```

samruddhikhandale:
Closing as fixed with https://github.com/devcontainers/features/pull/303

Feel free to re-open, if any issue resurfaces.

PR #303, opened 2022-11-22
go: fix bug - Go doesn't update version if go is already installed

samruddhikhandale:
Closes - https://github.com/devcontainers/features/issues/212


joshspicer (approved):
Love it 🔥

PR #305, opened 2022-11-23
Optionally install Azure Bicep as part of azure-cli feature

pauldotyu:
When users run the `az bicep` command for the first time, Azure CLI looks for the Bicep binary in `$HOME/.azure/bin` and if is not found, it will auto-install Bicep. This is a convenience feature of Azure CLI; however, it incorrectly installs Bicep on `arm64` OS architectures. 

Here is an example of an error a user will encounter when running `az bicep` commands in a devcontainer hosted on M1 Macs:

```
$ az bicep version

qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

The error is due to OS architecture mismatch between the container `arm64` and the Bicep binary `amd64` that was auto-installed by Azure CLI.

The long term fix for the Azure CLI to properly download the correct Bicep binary based on OS architecture. However, this may have a longer turnaround time.

The short term fix is in this PR which gives the user an option to install the Azure Bicep as part of the `azure-cli` feature. The install script will manually install Bicep based on OS architecture and create a symlink so that `bicep` is accessible as both a stand-alone executable and as part of `az`.

This will immediately unblock users who wish to use Bicep on arm64 based devcontainers and will submit an issue in the azure-cli repo to address the underlying root cause.

### DIFF
--- a/src/azure-cli/install.sh
+++ b/src/azure-cli/install.sh
@@ -14,6 +14,7 @@ rm -rf /var/lib/apt/lists/*
 
 AZ_VERSION=${VERSION:-"latest"}
 AZ_EXTENSIONS=${EXTENSIONS}
+AZ_INSTALLBICEP=${INSTALLBICEP:-false}
 
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 AZCLI_ARCHIVE_ARCHITECTURES="amd64"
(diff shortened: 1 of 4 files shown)


src/azure-cli/install.sh
```diff
@@ -197,6 +198,28 @@ if [ ${#AZ_EXTENSIONS[@]} -gt 0 ]; then
...
+    if [ "${architecture}" = "arm64" ]; then
+        curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-arm64
+    else 
+        curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+    fi
+    
+    chmod +x ./bicep
+    mv ./bicep /usr/local/bin/bicep
+
+    # Add a synlink so bicep can be accessed as a standalone executable or as part of az
```
joshspicer, line 217:
```suggestion
    # Add a symlink so bicep can be accessed as a standalone executable or as part of az
```

src/azure-cli/install.sh
```diff
@@ -197,6 +198,28 @@ if [ ${#AZ_EXTENSIONS[@]} -gt 0 ]; then
     done
 fi
```
joshspicer, line 200: To tell the `apt` that you've added runs in non-interactive mode

```suggestion
export DEBIAN_FRONTEND=noninteractive
```

src/azure-cli/install.sh
```diff
@@ -197,6 +198,28 @@ if [ ${#AZ_EXTENSIONS[@]} -gt 0 ]; then
...
 
+if [ "${AZ_INSTALLBICEP}" = "true" ]; then
+    # Properly install Azure Bicep based on current architecture
+    # The `az bicep install` command installs the linux-x64 binary even on arm64 devcontainers
+    # The `az bicep install --target-platform` could be a solution; however, linux-arm64 is not an allowed value for this argument yet
+    # Manually installing Bicep and moving to the appropriate directory where az expects it to be
+    apt-get install curl -y
```
joshspicer, line 206: Perhaps add `--no-install-recommends` for speed?

```suggestion
    apt-get install curl -y --no-install-recommends
```

joshspicer, line 206: We might need an `apt update` in certain cases if the lists haven't been populated yet (it's not safe to assume it has, and each of our features `rm`'d the list at the end to reduce the size of each Feature layer)

Feel free to rip off this function used in a lot of other scripts in this repo (hopefully will move this to a centralized spot in the future, ref: https://github.com/devcontainers/spec/issues/129#issuecomment-1316040043)

https://github.com/devcontainers/features/blob/c4648ab98cdeeb520e7e629c5db4a471ec802954/src/git/install.sh#L89-L94

joshspicer (commented):
This is awesome! Thank you for the detailed PR

src/azure-cli/install.sh
pauldotyu: Good catch! This has been fixed

src/azure-cli/install.sh
pauldotyu: looks like this line of code was already included on ln#60

src/azure-cli/install.sh
pauldotyu: Great call! I just noticed this script already includes the `check_packages` function so I will call that to ensure apt update is run and install curl.

pauldotyu:
> This is awesome! Thank you for the detailed PR. Left a couple suggestions if you have the time.

Thanks! Some edits have been made based on your feedback. Let me know if I am missing anything else.

joshspicer (approved):
🎉

PR #308, opened 2022-11-23
Adds test coverage

samruddhikhandale:
Adds test coverage to some Features.

### DIFF
--- a/test/docker-from-docker/docker_init.sh
+++ b/test/docker-from-docker/docker_init.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+check "docker-init-exists" bash -c "ls /usr/local/share | grep docker-init.sh"
+check "log-exists" bash -c "ls /tmp | grep vscr-docker-from-docker.log"
+check "log-contents-for-success" bash -c "cat /tmp/vscr-docker-from-docker.log | grep 'Success'"
+check "log-contents" bash -c "cat /tmp/vscr-docker-from-docker.log | grep 'Proxying /var/run/docker-host.sock to /var/run/docker.sock for vscode'"
+
+# Report result
+reportResults
--- a/test/docker-from-docker/docker_init.sh
+++ b/test/docker-from-docker/docker_init.sh
@@ -5,10 +5,11 @@ set -e
 # Optional: Import test library
 source dev-container-features-test-lib
 
-check "docker-init-exists" bash -c "ls /usr/local/share | grep docker-init.sh"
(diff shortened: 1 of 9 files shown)


test/docker-from-docker/docker_init.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+check "docker-init-exists" bash -c "ls /usr/local/share | grep docker-init.sh"
+check "log-exists" bash -c "ls /tmp | grep vscr-docker-from-docker.log"
+check "log-contents-for-success" bash -c "cat /tmp/vscr-docker-from-docker.log | grep 'Success'"
+check "log-contents" bash -c "cat /tmp/vscr-docker-from-docker.log | grep 'Proxying /var/run/docker-host.sock to /var/run/docker.sock for vscode'"
+
```
jkeech, line 12: Can we add another check that uses the docker daemon to ensure it's working properly beyond just checking for a successful startup message in the logs? For instance, running `docker ps`. Same thing for the docker-in-docker feature.

test/sshd/test.sh
```diff
@@ -5,7 +5,9 @@ set -e
 # Optional: Import test library
 source dev-container-features-test-lib
 
-echo 'NO TESTS FOR ME YET!'
+check "shh-init-exists" bash -c "ls /usr/local/share | grep ssh-init.sh"
```
jkeech, line 8: Typos on these check names here and below:
```suggestion
check "ssh-init-exists" bash -c "ls /usr/local/share | grep ssh-init.sh"
```

test/sshd/test.sh
```diff
@@ -5,7 +5,9 @@ set -e
...
+check "shhd-log-exists" bash -c "ls /tmp | grep sshd.log"
+check "shhd-log-contents" bash -c "cat /tmp/sshd.log | grep 'Starting OpenBSD Secure Shell server sshd'"
```
jkeech, line 11: Can we make sure the `sshd` process is running? Something like `ps -aux | grep -v grep | grep sshd`. Note that we need the `| grep -v grep` to remove the grep processes from the `ps` output, otherwise we get false positive matches. There's actually a bug in some of the [other test assertions ](https://github.com/search?q=repo%3Adevcontainers%2Ffeatures+%22ps+-ax%22+path%3Atest%2F**%2F*.sh&type=code) that don't do this.

test/docker-from-docker/docker_init.sh
samruddhikhandale: Yep, added

test/sshd/test.sh
samruddhikhandale: Added check & fixed the 🐛  ones

jkeech (approved):

Issue #307, opened 2022-11-23
Node feature permissions incorrect with custom non-root user


benjaminwood:
When the container is running with a non-root user, does not have a UID of `1000`, and is not named `vscode`, `node`, or `codespace`, the install script fails to set correct permissions on the node installation.

For example, say you're using Docker for mac and the container is running a user called `app` with a UID of `501`. The node installation will proceed as root, which results in a dev container with a broken node installation (for the `app` user).

The code in question is seen here:

https://github.com/devcontainers/features/blob/188f1ef26bc8fa942d552dfaa8596091a8a39126/src/node/install.sh#L38-L52

A workaround is to set `username` in the feature options, which is undocumented:

```json
"features": {
  "ghcr.io/devcontainers/features/node:1": {
    "version": "18",
    "username": "app"
  }
}
```

There is a `_DEV_CONTAINERS_IMAGE_USER` build arg passed to Dockerfile-with-features. If this build arg could be set as an environment variable when the feature install.sh is run, it could be used to determine the correct user without the "automatic" behavior that is currently used. This approach would require some changes outside of the install script itself (devcontainer cli?).

jkeech:
Thanks for reporting! I think we just need to update the install script to leverage the new `_REMOTE_USER` and `_CONTAINER_USER` variables which get passed in automatically. Reference: https://github.com/devcontainers/spec/blob/main/proposals/features-user-env-variables.md

cc @samruddhikhandale

samruddhikhandale:
Closing as fixed with https://github.com/devcontainers/features/pull/264

@benjaminwood feel free to re-open if you find any other issues! 😄

PR #264, opened 2022-11-08
Use _REMOTE_USER when available

Chuxel:
This PR updates features to take advantage of the new _REMOTE_USER environment variable. It leaves the existing logic in place to handle older versions of the dev container CLI (and to try to be smart in the case when no remote user is set).

Priority is:
1. `USERNAME` (if this was added as an option)
2. `_REMOTE_USER`
3. `automatic` - this is the final fallback to try to auto-detect and for back compat.

The PR also makes sure we are quoting inbound feature options to avoid spaces, newlines, etc causing problems.

PR #309, opened 2022-11-23
Features: Set USERNAME to _REMOTE_USER if available

samruddhikhandale:
Helps fix https://github.com/devcontainers/features/issues/307


src/anaconda/install.sh
```diff
@@ -9,7 +9,7 @@
 
 
 VERSION=${VERSION:-"latest"}
-USERNAME=${USERNAME:-"automatic"}
+USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
```
samruddhikhandale, line 12:
```suggestion
USERNAME="${USERNAME:-"${_REMOTE_USER:-"_CONTAINER_USER:-"automatic""}"}"
```

Not sure if we should also consider `_CONTAINER_USER` 💭

jkeech: According to the spec, we shouldn't need to. https://github.com/devcontainers/spec/blob/main/proposals/features-user-env-variables.md

>If no remoteUser is configured, _REMOTE_USER is set to the same value as _CONTAINER_USER.

jkeech (commented):
Can you please add some test cases that verify the features work properly for non-standard `remoteUser`s?

samruddhikhandale:
> Can you please add some test cases that verify the features work properly for non-standard `remoteUser`s?

For node, I tired adding a test-scenario as follows 

```
    "test_non_root_user": {
        "image": "ubuntu:focal",
        "features": {
            "common-utils": {
                "username": "app"
            },
            "node": {
                "version": "19.1.0"
            }
        }
    }
```

However, even before the changes in the PR, the tests pass. The `nvm` folder is set with correct user permissions & node is installed fine.

image

I thought of repro-ing the scenario using `dockerfile` which creates this user, however the features test command doesn't support that. Need to fix that. // cc - @joshspicer 

image

@jkeech Not sure how else I could write tests 😕

samruddhikhandale:
Closing this in favor of https://github.com/devcontainers/features/pull/264

Issue #259, opened 2022-11-03
[github-cli] Race condition when checking for latest version


mislav:
By default, `ghcr.io/devcontainers/features/github-cli` is configured to fetch and install the latest GitHub CLI version available.

However, it determines the latest release not by checking GitHub Releases in the `cli/cli` repository, but by scanning git tags in the same repository:
https://github.com/devcontainers/features/blob/73a9c7d38293067a4081b8d79dd12e92a585ef83/src/github-cli/install.sh#L168

This is wrong because a presence of a git tag does not guarantee presence of release assets (the script tries to download the `https://github.com/cli/cli/releases/download/v${CLI_VERSION}/gh_${CLI_VERSION}_linux_${arch}.deb` asset). Assets can only be downloaded from published Releases, not from git tags. Therefore, the script should check for published Releases in the `cli/cli` repository and ignore git tags.

Due to our release automation in the `cli/cli` repository, it can take 10–20 minutes between a git tag is published and the associated Release being published that contains all the compiled assets. During that time, this devcontainer feature is vulnerable:

```
2022-11-03 16:21:51.228Z: #14 0.678 Downloading github CLI...
2022-11-03 16:21:51.512Z: #14 1.068 CLI_VERSION=2.19.0
...
2022-11-03 16:21:51.624Z: #14 1.158 ... https://github.com/cli/cli/releases/download/v2.19.0/gh_2.19.0_linux_amd64.deb
...
2022-11-03 16:21:51.763Z: #14 1.265 HTTP request sent, awaiting response... 
2022-11-03 16:21:51.876Z: 404 Not Found
```

Thanks @tspascoal who ran into and reported this.

jkeech:
Thanks for reporting. I think one issue is that switching to the Releases API could very quickly run into auth issues and rate limiting when un-authed. Using the git tags helps avoid rate limiting on the APIs since this is a public feature that can be used in Dev Containers without being signed in to GitHub locally.

Even if we can't use the Releases API directly due to auth/rate limiting issues, we could definitely make the tag lookup smarter by trying the latest and then walking backwards if the latest one fails, resulting in matching the most recent one that is fully uploaded.

Chuxel:
Ideally the CLI repo would have a file in main with the list of current releases that we could parse for this.

But, yeah the release API is not an option due to rate limits but we can do a fallback. This is how the nix feature handles this.

[...]

PR #310, opened 2022-11-24
Github cli - Adds retry logic for find git tags

samruddhikhandale:
Closes - https://github.com/devcontainers/features/issues/259

and Fixes `find_prev_version_from_git_tags()` for Nix feature - scenario: finding latest previous major release.


src/github-cli/install.sh
```diff
@@ -158,6 +158,45 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
 }
 
+# Use semver logic to decrement a version number then look for the closest match
+find_prev_version_from_git_tags() {
```
samruddhikhandale, line 162: Copied from nix feature

src/nix/utils.sh
```diff
@@ -282,19 +282,26 @@ find_prev_version_from_git_tags() {
...
     breakfix="$(echo "${current_version}" | grep -oP '^[0-9]+\.[0-9]+\.\K[0-9]+' 2>/dev/null || echo '')"
-    set -e
+
+    if [ "${minor}" = "0" ] && [ "${breakfix}" = "0" ]; then
+        ((major=major-1))
+        declare -g ${variable_name}="${major}"
+        # Look for latest version from previous major release
+        find_version_from_git_tags "${variable_name}" "${repository}" "${prefix}" "${separator}" "${last_part_optional}"
     # Handle situations like Go's odd version pattern where "0" releases omit the last part
-    if [ "${breakfix}" = "" ] || [ "${breakfix}" = "0" ]; then
```
joshspicer, line 287: Was the change in the `nix` Feature intentional?  Curious if @Chuxel intended for this condition to be separate?

src/nix/utils.sh
```diff
@@ -282,19 +282,26 @@ find_prev_version_from_git_tags() {
...
+    else
         ((breakfix=breakfix-1))
         if [ "${breakfix}" = "0" ] && [ "${last_part_optional}" = "true" ]; then
             declare -g ${variable_name}="${major}.${minor}"
         else 
             declare -g ${variable_name}="${major}.${minor}.${breakfix}"
         fi
     fi
+
+    set -e
```
joshspicer, line 306: What error in the above block are you trying to swallow?

src/github-cli/install.sh
```diff
@@ -158,6 +158,47 @@ find_version_from_git_tags() {
...
+    local separator=${4:-"."}
+    # Some tools release versions that omit the last digit (e.g. go)
+    local last_part_optional=${5:-"false"}
+    # Some repositories may have tags that include a suffix (e.g. actions/node-versions)
+    local version_suffix_regex=$6
+    # Try one break fix version number less if we get a failure. Use "set +e" since "set -e" can cause failures in valid scenarios.
+    set +e
+    major="$(echo "${current_version}" | grep -oE '^[0-9]+' || echo '')"
+    minor="$(echo "${current_version}" | grep -oP '^[0-9]+\.\K[0-9]+' || echo '')"
+    breakfix="$(echo "${current_version}" | grep -oP '^[0-9]+\.[0-9]+\.\K[0-9]+' 2>/dev/null || echo '')"
```
joshspicer, line 178: Personal preference, but I like to indent the block of code that is affected by a change in `set +e` (like how we would indent an `if` statement's body for readability.)

```suggestion
    set +e
        major="$(echo "${current_version}" | grep -oE '^[0-9]+' || echo '')"
        minor="$(echo "${current_version}" | grep -oP '^[0-9]+\.\K[0-9]+' || echo '')"
        breakfix="$(echo "${current_version}" | grep -oP '^[0-9]+\.[0-9]+\.\K[0-9]+' 2>/dev/null || echo '')"
```
(and any other lines within the "scope" of this `set +e`)

This would also make it clearer to me that the `set -e` at the end is meant to restore from the `set +e` on this line

src/github-cli/install.sh
```diff
@@ -171,6 +212,15 @@ install_deb_using_github() {
     mkdir -p /tmp/ghcli
     pushd /tmp/ghcli
     wget https://github.com/cli/cli/releases/download/v${CLI_VERSION}/${cli_filename}
+    exit_code=$?
+    set -e
+    if [ "$exit_code" != "0" ]; then
```
joshspicer, line 217: This looks good to me, but wanted to share that I confirmed that both `"0"` and `0` would work here:


```bash
#!/bin/bash


blahblahblah
exit_code=$?

if [ "$exit_code" != "0" ]; then
   echo 'exit code does not equal the string "0"'
fi


if [ "$exit_code" != 0 ]; then
   echo 'exit code does not equal the number 0'
fi


date
exit_code=$?

if [ "$exit_code" == "0" ]; then
   echo 'exit code does equal the string "0"'
fi


if [ "$exit_code" == 0 ]; then
   echo 'exit code does equal the number 0'
fi

```

#### Result:
```
joshspicer@gh-m1 /tmp % ./test.sh

./test.sh: line 4: blahblahblah: command not found
exit code does not equal the string "0"
exit code does not equal the number 0
Mon Nov 28 16:57:19 PST 2022
exit code does equal the string "0"
exit code does equal the number 0

```

joshspicer (commented):
Have you considered approaching from the other angle, and contributing back a change to the cli/cli repo with something like @Chuxel recommended here: https://github.com/devcontainers/features/issues/259#issuecomment-1304405293 ?

Assuming they accept the contribution, that might make our Feature significantly simpler and easier to maintain/debug in the future?

src/nix/utils.sh
Chuxel: Go omits the last digit for major versions.  It's 1.19 not 1.19.0, so yes, that was intentional.

src/nix/utils.sh
Chuxel: The set `+e` was to handle non-zero exit codes from grep.  There may be a flag that could be passed in instead.

src/nix/utils.sh
samruddhikhandale: > Was the change in the nix Feature intentional? 

Yep, it was made to handle major version changes (eg. previous version from `2.0.0` for git is `1.14.1`)

src/github-cli/install.sh
samruddhikhandale: Nice, good to know!

src/github-cli/install.sh
samruddhikhandale: Makes sense. Updated

joshspicer (approved):

PR #318, opened 2022-12-01
github-cli: simplify logic for detecting available gh releases

mislav:
This change relies on `git ls-remote --tags --sort=version:refname` to sort git versions in a semver-compatible mode and alleviate the need to scan for version numbers manually. The new approach selects at most two latest releases that match VERSION and installs the first `.deb` that was successfully downloaded from a GitHub CLI release.

This also removes support for installing a specific version number with `apt-get install gh=VERSION`, since the GitHub CLI package repository does not retain older gh versions and only allows installing the latest one.

Finally, this removes version-selecting tests that only tested an implementation that was entirely contained within tests.

Bonuses:
- Only install curl, apt-transport-https, dirmngr, gnupg2 when needed (the default installDirectlyFromGitHubRelease mode does not need them).
- Ensure `apt-get update` runs at most once.
- Remove the side-effect of `/var/lib/apt/lists` being wiped out when the github-cli feature is activated.
- Add ability to export bash functions for testing: `TEST_EXPORT=find_version_from_git_tags ./src/github-cli/install.sh`
- Errors are printed to stderr.

Followup to https://github.com/devcontainers/features/pull/310

/cc @samruddhikhandale: I thought I'd take a stab at reducing the size of the implementation so that the github-cli feature is easier to maintain over time. I hope this can be useful! 🙇


src/github-cli/install.sh
```diff
@@ -124,138 +124,85 @@ check_packages() {
...
+type -P git >/dev/null || check_packages git
+
+is_latest="false"
+
+case "$CLI_VERSION" in
+none )
+    # opt out of installing gh
+    exit 0
+    ;;
+latest | lts | stable | current )
```
mislav, line 167: Is my interpretation of `"version": "none"` and `"version": "current"` correct? I wasn't sure how those values are supposed to be handled since they aren't documented

src/github-cli/install.sh
```diff
@@ -9,6 +9,7 @@
 
 CLI_VERSION=${VERSION:-"latest"}
 INSTALL_DIRECTLY_FROM_GITHUB_RELEASE=${INSTALLDIRECTLYFROMGITHUBRELEASE:-"true"}
+TEST_EXPORT=${TEST_EXPORT:-}
```
samruddhikhandale, line 12: > Add ability to export bash functions for testing: TEST_EXPORT=find_version_from_git_tags ./src/github-cli/install.sh

Could you help me explain the scenarios in which this is helpful?  I read this part ☝️ , but usually tests for Features are run with `devcontainers features test` cmd. Curious to understand your thought behind it's usage 😄

src/github-cli/install.sh
```diff
@@ -17,11 +18,8 @@ keyserver hkp://keyserver.pgp.com"
 
 set -e
 
-# Clean up
```
samruddhikhandale, line 20: This PR https://github.com/devcontainers/features/pull/210 intentionally added the cleaning up of old packages at the start and end of the script (for all Features) which makes sense to me. Hence, each Feature cleans up the cache and starts afresh. Do you have any other thoughts where this would not be helpful ?

src/github-cli/install.sh
```diff
@@ -108,12 +106,14 @@ receive_gpg_keys() {
     fi
 }
 
-apt_get_update()
-{
-    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
-        echo "Running apt-get update..."
-        apt-get update -y
-    fi
+ran_apt_get_update="false"
```
samruddhikhandale, line 109: If we decide to re-add the Cleanup, we could remove this change as well 😄

src/github-cli/install.sh
```diff
@@ -124,138 +124,85 @@ check_packages() {
...
-            find_version_from_git_tags "${variable_name}" "${repository}" "${prefix}" "${separator}" "${last_part_optional}"
-        else
-            ((breakfix=breakfix-1))
-            if [ "${breakfix}" = "0" ] && [ "${last_part_optional}" = "true" ]; then
-                declare -g ${variable_name}="${major}.${minor}"
-            else 
-                declare -g ${variable_name}="${major}.${minor}.${breakfix}"
-            fi
+    local prefix="${1:-}"
+    git ls-remote --tags --sort=-version:refname https://github.com/cli/cli.git | while read -r _ ref; do
```
samruddhikhandale, line 130: Woah! This looks so much cleaner and nicer 👏 

Wondering, if you could add few test scenarios to validate that these changes work as expected? 

We have some guidance on how write tests here - https://github.com/devcontainers/cli/blob/main/docs/features/test.md#scenarios

Also, this repo has quite a few samples (eg - [dotnet](https://github.com/devcontainers/features/tree/main/test/dotnet))

src/github-cli/install.sh
```diff
@@ -124,138 +124,85 @@ check_packages() {
...
+latest | lts | stable | current )
+    # find two latest versions
+    versions=( $(find_version_from_git_tags | head -n2) )
+    is_latest="true"
+    ;;
+*.*.* )
+    # the exact version is specified explicitly
+    versions=( "$CLI_VERSION" )
+    ;;
+* )
```
samruddhikhandale, line 176: Can we also add another case for `*.*` ? that's something users might use as well.

src/github-cli/install.sh
samruddhikhandale: The current `proposals` property mentions `latest` and `none`. But if the gihutb-cli supports other keywords, then it's nice to add. Thanks! 

Feel free to update the proposals here as well - [here](https://github.com/devcontainers/features/blob/e7ebc0e9de1e3cf61424096352561d94b7e3e9a0/src/github-cli/devcontainer-feature.json#L10)

src/github-cli/install.sh
```diff
@@ -124,138 +124,85 @@ check_packages() {
...
 else
-   # Import key safely (new method rather than deprecated apt-key approach) and install
+    check_packages curl apt-transport-https dirmngr gnupg2
+    # Import key safely (new method rather than deprecated apt-key approach) and install
     . /etc/os-release
     receive_gpg_keys GITHUB_CLI_ARCHIVE_GPG_KEY /usr/share/keyrings/githubcli-archive-keyring.gpg
     echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list
     apt-get update
-    apt-get -y install "gh${version_suffix}"
+    apt-get -y install gh
```
samruddhikhandale, line 205: Good to know. Thanks! 

Can we also add a description to [installDirectlyFromGitHubRelease property](https://github.com/devcontainers/features/blob/e7ebc0e9de1e3cf61424096352561d94b7e3e9a0/src/github-cli/devcontainer-feature.json#L17) to prompt that?

Something like 👇 

> If false, it installs the latest version of GitHub CLI

samruddhikhandale (commented):
This is awesome! Thank you for providing a cleaner and sleek approach for detecting releases. Much appreciated 👏 

I have left some comments/questions. Would appreciate if you could help add tests to validate this change. 😄

src/github-cli/install.sh
mislav: From existing code, it looked like `github-cli` feature supported many different keywords, but I only want to support the old ones for backwards compatibility and not actively suggest their use. I do not even understand what `none` is about. The usages I'd like to document are:

1. `"version": "latest"` - default
2. `"version": "2.20.1"` - lock to a specific version
3. `"version": "2.20"` - install the latest 2.20.x
4. `"version": "2"` - install the latest 2.x

Any suggestions how to best document this?

src/github-cli/install.sh
mislav: I used this "export" mechanism in my development to manually test a single bash function from the `install.sh` script without running the whole `install.sh` script (which wouldn't even work on macOS). I noticed that you copied over bash functions for testing into the `test.sh` script, and with this mechanism you could possibly export functions for testing in CI rather that copying over the functions. However, I haven't made use of that in the current PR, and I'm not really attached to the TEST_EXPORT mechanism either. We could also just drop it from the final implementation.

src/github-cli/install.sh
mislav: Thanks for linking to the more general PR. I did not realize this pattern spans multiple different features.

I found it odd that simply activating the `github-cli` feature would have side-effects on the apt-get mechanism, even if `apt-get update` wasn't invoked at all. Also, `/var/lib/apt/lists` is an implementation detail of apt-get and I don't think it's healthy for scripts like ours to rely on it in the long run. Finally, I find that cleaning the `/var/lib/apt/lists` cache after the script is done is unnecessary as caches exist for a reason. So my suggestion would be to just do `apt-get update` once, then `apt-get install` a few times to get what we need, and to leave everything else to the system.

src/github-cli/install.sh
mislav: Thanks for the pointers! I'll try to do that

src/github-cli/install.sh
mislav: Good point, but the case for `*.*` is already handled by the general fallback case `*`.

mislav:
Looking at the CI failure: it looks like the git version on Bionic is too old to support the `--sort=version:refname` option. I'll update the PR to achieve semantic version sorting differently than relying on this git feature.

src/github-cli/install.sh
samruddhikhandale: I see, for `version` option we are accepting semver values so I think it should be self explanatory of how versions are computed. However, it's great to document and make the usage clearer. Feel free to add a `NOTES.md` file in https://github.com/devcontainers/features/tree/main/src/github-cli. The action would automatically add the contents to the [README](https://github.com/devcontainers/features/blob/main/src/github-cli/README.md) file.

(example - [docker-in-docker](https://github.com/devcontainers/features/tree/main/src/docker-in-docker))

src/github-cli/install.sh
samruddhikhandale: Makes sense. I agree that if we aren't planning to make use of it, it's great to drop the changes. thanks!

samruddhikhandale:
Hi @mislav 👋 

This PR would be super useful and will improve the current Feature. Let me know if I could somehow help to push it through. thanks!

mislav:
@samruddhikhandale Thanks; I intend to finish this PR soon by addressing review comments.

- [ ] Address case where git is old (Bionic)
- [ ] Restore apt-get cleanup step (for consistency with other features in this repo)
- [ ] Add more scenarios to tests
- [ ] Remove TEST_EXPORT hack (since we will be using scenarios for integration testing)

mislav:
Thanks for all your guidance @samruddhikhandale, but I'm going to close this because it's no longer needed. In GitHub CLI we've just changed our deployment process so that release assets are published at the same time as the git tag itself. Therefore, there isn't a period of time anymore where a git tag exists but the assets do not exist, so there is no need for a fallback mechanism to select the previous git tag.

samruddhikhandale:
Nice, that's awesome. Thank you for the update!

PR #322, opened 2022-12-05
Add NbDime to Jupyter Lab

gwincr11:
Git base notebook diffing is not a great experience, lets make it better! NbDime is a jupyterlab plugin that will help with this and is also what we base notebook diffing on in dotcom so it will bring some consistency to the experience.

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.0.17",
+  "version": "1.0.18",
   "name": "Python",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/python",
   "description": "Installs the provided version of Python, as well as PIPX, and other common Python utilities.  JupyterLab is conditionally installed with the python feature. Note: May require source code compilation.",
@@ -55,10 +55,7 @@
   },
   "customizations": {
     "vscode": {
-      "extensions": [
-        "ms-python.python",
-        "ms-python.vscode-pylance"
-      ],
+      "extensions": ["ms-python.python", "ms-python.vscode-pylance"],
       "settings": {
         "python.defaultInterpreterPath": "/usr/local/python/current/bin/python",
         "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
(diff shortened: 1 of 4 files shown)


samruddhikhandale:
Thanks @gwincr11 👋  This definitely looks like a useful package and +1 for having a consistent experience. Looking at the consumed disk size of this package (& it's dependencies), it takes **~47 MB** of disk space. 

@Chuxel / @jkeech / @jungaretti / @joshspicer Any thoughts/concerns about adding `nbdime` package to the Feature?

Chuxel:
For the core feature, I think in principle we shouldn't add things that are not mandatory by default. Having an option to specify a list extensions to install would be more flexible. Otherwise there will be an eternal debate as to which ones to add or remove.

To say it again, generally I think Jupyter shouldn't be a part of the python feature either for this same reason (it should be its own), but that's a separate discussion.

For the universal/kitchen sink image, that's a bit more opinionated, so I think the main concern is size there.

samruddhikhandale:
@gwincr11 Hi 👋 

Wondering if you have any thoughts on Chuck's comment https://github.com/devcontainers/features/pull/322#issuecomment-1341691077 Thanks!

gwincr11:
> For the core feature, I think in principle we shouldn't add things that are not mandatory by default. Is this one mandatory/must have?

So, there is an argument to be made towards yes. Jupyter Notebooks are stored as a json structure that is representative of the information needed to re-construct the notebook by a computer. When using Git the diff output by default is a diff of this json structure often containing full binary blocks. NbDime makes Git diffing usable to a human. So if the goal is to give the user a usable Git Diff, then it or a package like it is mandatory. Being that we are GitHub and every thing we want them to accomplish involves Git I think there is a strong case yes.

samruddhikhandale:
Thanks for the detailed explanation. It makes sense.

We recently added a `jupyterlab-git` to help with with the basic Git version control.
I could see  `jupyterlab-git`  doing some similar diffing for Jupyter Notebooks as that of `nbdime`

`jupyterlab-git` output from [Binder](https://mybinder.org/v2/gh/jupyterlab/jupyterlab-git/master?urlpath=lab/tree/examples/demo.ipynb) demo provided in [jupyterlab/jupyterlab-git](https://github.com/jupyterlab/jupyterlab-git)

image

@gwincr11 Do you know how these two tools compare with one another, also should we install both or favor one?

gwincr11:
Oh looks like they just pull in nbdime https://github.com/jupyterlab/jupyterlab-git/blob/0ddaf8dc7068d19a9c94e8645a5f8c467e70e777/setup.cfg#L26

TIL I think we are good here.

PR #325, opened 2022-12-07
Change default shell if installing ZSH

### LINKED ISSUE #324: Common Utils - If installing ZSH, set as default shell for user
Hi,

When using the `common-utils` feature, and setting `installZsh` to `true` (default behaviour), I think it would be useful to change the default shell for the determined `${USERNAME}`. I've found that I either need to do it in `postCreateCommand` (e.g. `sudo chsh --shell /bin/zsh vscode`) or configure it via [Terminal Profiles](https://code.visualstudio.com/docs/terminal/profiles), although I haven't managed to get this working yet using [`customizations`](https://containers.dev/supporting#visual-studio-code)

jacobwoffenden:
Changes default shell to `/bin/zsh` 

Resolves #324

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "1.1.6",
+    "version": "1.2.6",
     "name": "Common Debian Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",
@@ -9,6 +9,11 @@
             "type": "boolean",
             "default": true,
             "description": "Install ZSH?"
+        },
+        "configureZshAsDefaultShell": {
+            "type": "boolean",
+            "default": false,
+            "description": "Change default shell to ZSH?"
         },
         "installOhMyZsh": {
             "type": "boolean",
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "1.2.6",
+    "version": "1.2.0",
     "name": "Common Debian Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",
--- a/src/common-utils/install.sh
+++ b/src/common-utils/install.sh
@@ -13,6 +13,7 @@ set -e
 rm -rf /var/lib/apt/lists/*
 
 INSTALL_ZSH="${INSTALLZSH:-"true"}"
+CONFIGURE_ZSH_AS_DEFAULT_SHELL="${CONFIGUREZSHASDEFAULTSHELL:-"false"}"
 INSTALL_OH_MY_ZSH="${INSTALLOHMYZSH:-"true"}"
 UPGRADE_PACKAGES="${UPGRADEPACKAGES:-"true"}"
 USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
@@ -377,6 +378,12 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
         echo "${rc_snippet}" >> /etc/zsh/zshrc
         ZSH_ALREADY_INSTALLED="true"
     fi
+
+    # Change shell of determined user
+    if [[ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]]; then
+      chsh --shell /bin/zsh ${USERNAME}
+    fi
+
 
     # Adapted, simplified inline Oh My Zsh! install steps that adds, defaults to a codespaces theme.
     # See https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/install.sh for official script.
--- a/test/common-utils/configure_zsh_as_default_shell.sh
+++ b/test/common-utils/configure_zsh_as_default_shell.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+check "default-shell-is-zsh" getent passwd vscode | awk -F: '{ print $7 }' | grep "/bin/zsh"
+
+# Report result
+reportResults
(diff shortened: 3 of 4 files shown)


samruddhikhandale:
Added my thoughts to the original issue https://github.com/devcontainers/features/issues/324#issuecomment-1343303523

> Hi @jacobwoffenden 👋 ,
> 
> As a heads up, we're discussing splitting up the `common-utils` Feature: https://github.com/devcontainers/features/issues/67 where we have discussed of creating a new feature for `Zsh`.
> 
> >  When using the common-utils feature, and setting installZsh to true (default behaviour), I think it would be useful to change the default shell for the determined ${USERNAME}
> 
> This is a good point and makes sense to me. However, making this change now could be a little disruptive for the users who expects `bash` as their default shell when they use this `common-utils` Feature. Also, most of the [dev container images](https://github.com/devcontainers/images) rely on this `common-utils` Feature. The community using these images have also been expecting `bash` as the default shell. We could add a command to change the shell to `bash` for the image but again it would be little contradictory.
> 
> I wonder if we could hold on making this change in https://github.com/devcontainers/features/pull/325 for now as it will be soon addressed by https://github.com/devcontainers/features/issues/67#issuecomment-1191525229? I can add this suggestion to ensure that `zsh` is configured for the default user when the `zsh` Feature is used. We could also add a new `isConfigured: true` property for the Feature so that anyone could disable and simply install `zsh`. (That'll be useful for the dev container images)
> 
> @jacobwoffenden what do you think?

jacobwoffenden:

src/common-utils/devcontainer-feature.json
```diff
@@ -10,6 +10,11 @@
             "default": true,
             "description": "Install ZSH?"
         },
+        "changeDefaultShellZsh": {
```
samruddhikhandale, line 13:
```suggestion
        "configureZshAsDefaultShell": {
```

src/common-utils/devcontainer-feature.json
```diff
@@ -10,6 +10,11 @@
             "default": true,
```
samruddhikhandale, line 10: Can we bump the minor version in here? 

https://github.com/devcontainers/features/blob/1c60fd71daa84c5fb81b4daf499974dbffb9e029/src/common-utils/devcontainer-feature.json#L3

src/common-utils/install.sh
```diff
@@ -13,6 +13,7 @@ set -e
 rm -rf /var/lib/apt/lists/*
 
 INSTALL_ZSH="${INSTALLZSH:-"true"}"
+CHANGE_DEFAULT_SHELL_ZSH="${CHANGEDEFAULTSHELLZSH:-"false"}"
```
samruddhikhandale, line 16: ℹ️  If you decide to rename the option, you'd need to change this as well

src/common-utils/install.sh
```diff
@@ -378,6 +379,12 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
         ZSH_ALREADY_INSTALLED="true"
     fi
 
+    # Change shell of determined user
+    if [[ "${CHANGE_DEFAULT_SHELL_ZSH}" == "true" ]]; then
+      chsh --shell /bin/zsh ${USERNAME}
```
samruddhikhandale, line 384: Could you write a test scenario and help validate this change? Thanks! 

Ref doc - https://github.com/devcontainers/cli/blob/main/docs/features/test.md
Sample - [dotnet tests](https://github.com/devcontainers/features/tree/1c60fd71daa84c5fb81b4daf499974dbffb9e029/test/dotnet)

samruddhikhandale (commented):
Thank you, appreciate it! Left some comments

src/common-utils/devcontainer-feature.json
jacobwoffenden: Looks good, will update!

src/common-utils/install.sh
jacobwoffenden: Committed, please let me know if it needs amending

test/common-utils/configure_zsh_as_default_shell.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+check "default-shell-is-zsh" getent passwd vscode | awk -F: '{ print $7 }' | grep "/bin/zsh"
```
jacobwoffenden, line 9: I've hardcoded `vscode` here as its the default user created by common-utils. Ideally I'd not want to hardcode this, but the logic for getting the user isn't the nicest

test/common-utils/scenarios.json
```diff
@@ -0,0 +1,10 @@
+{
+  "configure_zsh_as_default_shell": {
+    "image": "ubuntu:focal",
```
samruddhikhandale, line 3:
```suggestion
    "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
```

💡  If you'd change the `image`, then in the test instead of hardcoding `vscode`, you could use `$(whoami)`

test/common-utils/configure_zsh_as_default_shell.sh
samruddhikhandale: Left a comment on how you could possibly avoid hardcoding `vscode`

src/common-utils/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "1.1.6",
+    "version": "1.2.6",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.2.0",
```

jacobwoffenden: Sorry, it was late when I made this edit!

test/common-utils/configure_zsh_as_default_shell.sh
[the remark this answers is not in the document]
jacobwoffenden: Yes it's set in here to run as vscode but the built container still runs as root https://github.com/devcontainers/images/blob/main/src/base-ubuntu/.devcontainer/devcontainer.json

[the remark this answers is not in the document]
samruddhikhandale: Looks like you are on an older version of the CLI. We recently had an image metadata change which should run the test as `remoteUser`

image

Hence, the suggested change should be fine!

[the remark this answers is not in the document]
samruddhikhandale: It doesn't even need the `scenario` to mention `remoteUser`. It just works as the metadata flows through it!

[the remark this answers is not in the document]
jacobwoffenden: Fantastic, anything else to address before its ready?

samruddhikhandale (approved):
Looks amazing! Thank you 👏

PR #326, opened 2022-12-08
Multi-distro support for common script, refactor, fixes

Chuxel:
This PR refactors the common script to support Debian/Ubuntu, RedHat/Fedora, and Alpine based images.  

Rather than having one massive script, it also refactors it into multiple parts for better maintainability.

Since in the Alpine case, we cannot assume bash is present, it bootstraps using `sh` in `install.sh` and then, after installing bash, execs `main.sh` where the core logic exists.

In addition, due to the performance impact of the "dirty" flag in the prompt, this is now moved behind an option and off by default.

Finally, the theme was renamed "devcontainers" rather than codespaces, though a symlink is in place for back compat.


src/common-utils/devcontainer-feature.json
```diff
@@ -23,36 +23,37 @@
...
             "proposals": [
                 "1000",
                 "automatic"
             ],
             "default": "automatic",
-            "description": "Enter uid for non-root user"
+            "description": "Enter UID for non-root user"
         },
-        "gid": {
+        "userGid": {
```
eitsupi, line 44: Is a major version upgrade required because of the option name change?

Chuxel: Yeah, possibly. What I noticed was that `UID` was set by the OS, so this overlapped and could lead to odd behaviors. When testing, I kept hitting a problem with the UID always being `0` and then I realized what happened.

Chuxel: Bumped major to 2.0.0.

samruddhikhandale: > What I noticed was that UID was set by the OS, so this overlapped and could lead to odd behaviors. When testing, I kept hitting a problem with the UID always being 0 and then I realized what happened.

Curious, @Chuxel how widespread is this issue? 
I wonder if we'd want to be more caution with bumping major versions? People might get stuck on old versions until they update it. 🤔 

@joshspicer  would soon have make this communication/signaling of needing to bump major versions a bit smoother/clearer. However, until then, I wonder if we could avoid renaming the options and bumping major versions till then? 🤔 

I completely understand if we have to fix the bug if it's widespread, but wondering if we could delay if it's not? 🤔

Chuxel: @samruddhikhandale It happened 100% of the time when I was doing testing via `devcontainer feature test`. The UID was always `0` and I got test failures because of it.  I can't speak to its use in the wild... I wouldn't expect most people would set these two properties TBH, so people just may not have noticed.

Chuxel:
Manually merged in the "default shell to zsh" option.

Chuxel:
Failures for [test (common-utils, mcr.microsoft.com/devcontainers/base:ubuntu)](https://github.com/devcontainers/features/actions/runs/3686963675/jobs/6239916670#logs) and [test (common-utils, mcr.microsoft.com/devcontainers/base:debian)](https://github.com/devcontainers/features/actions/runs/3686963675/jobs/6239916832s#logs) are expected since the theme was renamed to devcontainers.zsh-theme, and the current images have the old name. That will go away with the next published image.

src/common-utils/main.sh
```diff
@@ -0,0 +1,492 @@
...
+    ADJUSTED_ID="rhel"
+elif [ "${ID}" = "alpine" ]; then
+    ADJUSTED_ID="alpine"
+else
+    echo "Linux distro ${ID} not supported."
+    exit 1
+fi
+
+# Install packages for appropriate OS
+if [ "${PACKAGES_ALREADY_INSTALLED}" != "true" ]; then
```
samruddhikhandale, line 302: 🎉 🎉

src/common-utils/main.sh
```diff
@@ -0,0 +1,492 @@
...
+            -c fetch.fsck.zeroPaddedFilemode=ignore \
+            -c receive.fsck.zeroPaddedFilemode=ignore \
+            "https://github.com/ohmyzsh/ohmyzsh" "${oh_my_install_dir}" 2>&1
+        echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
+        sed -i -e 's/ZSH_THEME=.*/ZSH_THEME="devcontainers"/g' ${user_rc_file}
+
+        # Add Dev Containers theme
+        mkdir -p ${oh_my_install_dir}/custom/themes
+        cp -f "${FEATURE_DIR}/scripts/devcontainers.zsh-theme" "${oh_my_install_dir}/custom/themes/devcontainers.zsh-theme"
+        ln -s "${oh_my_install_dir}/custom/themes/devcontainers.zsh-theme" "${oh_my_install_dir}/custom/themes/codespaces.zsh-theme"
```
samruddhikhandale, line 448: > Finally, the theme was renamed "devcontainers" rather than codespaces, though a symlink is in place for back compat.

I understand that the renamed theme has a back compat for codespaces. I don't believe this would impact codepsaces. However, would just like to double check if Codespaces might be impacted in any way due to this?

samruddhikhandale (approved):
That's a massive refactoring! Thanks for making that change.
Overall it LGTM! ✨

src/common-utils/main.sh
Chuxel: Nope.  Both the name "devcontainer" and "codespace" for the theme will work. So existing dotfiles wouldn't be affected either if they referene the existing id.

samruddhikhandale: Sounds good 👍

Chuxel:
@samruddhikhandale Found another bug in the original script that caused it to not use `_REMOTE_USER`.  Fixed that too and added tests.  Ready to merge on your approval.

samruddhikhandale (approved):
Looks good! Thank you!

PR #344, opened 2022-12-17
Add support for mariner

markphip:
Verified this works by just adding "mariner" to the list of distro ID's you treat as "rhel"

Manually published the feature and built a Codespace that uses a Mariner image and this feature to configure it.

Also ran the tests in my fork before submitting the PR: https://github.com/markphip/features/actions

If you do not want to "fully support" mariner you could just drop the tests aspect of this PR


joshspicer (approved):
Thanks for this!

PR #346, opened 2022-12-18
Add amazon coretto on java feature

mgoldschmidt-ds:


mgoldschmidt-ds:
Sorry guys, just playng around and made this PR by accident.

Issue #323, opened 2022-12-06
Terraform - Downloading GPG Key step gets stuck


DevSecNinja:
Hello,

Since a few days I notice that the GPG Key download step in the Terraform feature hangs:

````
=> [internal] load metadata for mcr.microsoft.com/devcontainers/base:j  608.5s
 => => extracting sha256:9a586f3d84de83b25cb9ca6d0e733d37d5283da35a837917  0.8s
 => => extracting sha256:5ac951de24f647413f348bfa4183b92c9962b167e0a3106d  0.3s
 => => extracting sha256:1aa90a4f5c993fd6d1e5f0885a1794ee59253ee8d35463ea  0.0s
 => => extracting sha256:074fbd7f2200028d96665f87580813cabf3361b85a4217ce  0.0s
 => => extracting sha256:ee36e8b50d1a0a2f6169666f26f3938fe1ea3efabfb6df01  0.0s
 => => extracting sha256:9e8c9f9ef24f45765c7592d55a952cbe2d0d4fefcd797237  5.5s
 => [dev_containers_feature_content_normalize 2/3] COPY --from=dev_contai  4.8s
 => [dev_containers_feature_content_normalize 3/3] RUN chmod -R 0700 /tmp  1.7s
 => [dev_containers_target_stage  2/12] COPY --from=dev_containers_featur  0.3s
 => [dev_containers_target_stage  3/12] RUN echo "_CONTAINER_USER_HOME=$(  0.7s
 => [dev_containers_target_stage  4/12] RUN cd /tmp/build-features/azure  61.0s
 => [dev_containers_target_stage  5/12] RUN cd /tmp/build-features/github  6.1s
 => [dev_containers_target_stage  6/12] RUN cd /tmp/build-features/kubec  51.5s
 => [dev_containers_target_stage  7/12] RUN cd /tmp/build-features/power  34.3s
 => [dev_containers_target_stage  8/12] RUN cd /tmp/build-features/terr  431.1s
 => => # Downloading terraform...
 => => # TERRAFORM_GPG_KEY=72D7468F
 => => # GPG_KEY_SERVERS=keyserver hkp://keyserver.ubuntu.com:80
 => => # keyserver hkps://keys.openpgp.org
 => => # keyserver hkp://keyserver.pgp.com
 => => # (*) Downloading GPG key...
````

Screenshot:

[...]

PR #353, opened 2022-12-20
Terraform: Update GPG_KEY_SERVERS

samruddhikhandale:
Addresses - https://github.com/devcontainers/features/issues/323

### DIFF
--- a/src/terraform/install.sh
+++ b/src/terraform/install.sh
@@ -26,7 +26,7 @@ TERRAFORM_DOCS_SHA256="${TERRAFORM_DOCS_SHA256:-"automatic"}"
 
 TERRAFORM_GPG_KEY="72D7468F"
 TFLINT_GPG_KEY_URI="https://raw.githubusercontent.com/terraform-linters/tflint/master/8CE69160EB3F2FE9.key"
-GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com:80
+GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com
 keyserver hkps://keys.openpgp.org
 keyserver hkp://keyserver.pgp.com"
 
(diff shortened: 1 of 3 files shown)


src/terraform/install.sh
```diff
@@ -44,26 +44,9 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
-# Get central common setting
-get_common_setting() {
```
samruddhikhandale, line 48: @Chuxel Any concerns with removing this? 
We already have the correct variables set in this `install.sh` file.

src/terraform/install.sh
```diff
@@ -26,7 +26,7 @@ TERRAFORM_DOCS_SHA256="${TERRAFORM_DOCS_SHA256:-"automatic"}"
 
 TERRAFORM_GPG_KEY="72D7468F"
 TFLINT_GPG_KEY_URI="https://raw.githubusercontent.com/terraform-linters/tflint/master/8CE69160EB3F2FE9.key"
-GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com:80
```
samruddhikhandale, line 29: Many folks have reported that the fetching GPG key from `hkp://keyserver.ubuntu.com:80` stuck for them and works with `hkp://keyserver.ubuntu.com`. 

https://github.com/devcontainers/features/issues/323

src/terraform/install.sh
samruddhikhandale: If we happen to need this `settings.env` file, we should think of migrating it from the old repo.

joshspicer (approved):

seBasKov:
Hi samruddhikhandale, I noticed after this change my terraform feature install hangs at downloading the GPG key. I was referring to major version 1 as a result it picks up your change 1.1.1. 

My issue is solved using version 1.1.0 again. 

Might be worth to have a look as it seems that this bugfix introduces a new one.

PR #359, opened 2022-12-24
feat(common-utils): Support the new path of `meta.env` file and the new field for saving source revisions

### LINKED ISSUE #345: [common-utils] Update meta.env file's path and format
I am looking to implement the `devcontainer-info` command in our build of Dev Container images.

https://github.com/devcontainers/features/blob/2af02c198adabacff30c400b0bfcad972ce5abcc/src/common-utils/bin/devcontainer-info#L2

- Could it also support the new path like `/usr/local/etc/dev-containers/meta.env`?
- Could you add a new field to store the Git commit hash?

eitsupi:
Close #345

Add the following two features to the `devcontainer-info` command:

- Allows `meta.env` file to be placed in `/usr/local/etc/dev-containers/meta.env`.
- Reads and displays variable `GIT_REPOSITORY_REVISION`, which is supposed to record the revision of the source code of the image.

### DIFF
--- a/src/common-utils/bin/devcontainer-info
+++ b/src/common-utils/bin/devcontainer-info
@@ -1,5 +1,12 @@
 #!/bin/sh
-. /usr/local/etc/vscode-dev-containers/meta.env
+
+# Load meta.env
+if [ -f "/usr/local/etc/vscode-dev-containers/meta.env" ]; then
+    . /usr/local/etc/vscode-dev-containers/meta.env
+fi
+if [ -f "/usr/local/etc/dev-containers/meta.env" ]; then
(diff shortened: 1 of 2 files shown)


joshspicer (approved):

eitsupi:
Thank you for merging!
By the way, since I did not bump the version number, should I create a separate PR just to bump the version number if there doesn't seem to be any other updates in the near future?

samruddhikhandale:
> Thank you for merging! By the way, since I did not bump the version number, should I create a separate PR just to bump the version number if there doesn't seem to be any other updates in the near future?

Thanks, @eitsupi 
Created a PR for version bump! https://github.com/devcontainers/features/pull/367

PR #370, opened 2022-12-30
desktop-lite: fixed qemu container

mrkbac:
`pidof` does not work as intended when using, a qemu or rosetta container, `pgrep -x` has the desired effect.

### DIFF
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -299,9 +299,9 @@ startInBackgroundIfNotRunning()
 {
     log "Starting \$1."
     echo -e "\n** \$(date) **" | sudoIf tee -a /tmp/\$1.log > /dev/null
-    if ! pidof \$1 > /dev/null; then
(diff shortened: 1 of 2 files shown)


src/desktop-lite/install.sh
```diff
@@ -299,9 +299,9 @@ startInBackgroundIfNotRunning()
 {
     log "Starting \$1."
```
samruddhikhandale, line 300: Can we bump the patch version in here? thanks! 

https://github.com/devcontainers/features/blob/883cf50eb84f2ad3dada748401a5ad6c06bf43b6/src/desktop-lite/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Left a comment. Besides that LGTM, thanks!

mrkbac:

samruddhikhandale (approved):
Nice!

PR #371, opened 2022-12-31
[PHP] Enable Mbstring

julien-boudry:
Mbstring is such a popular PHP extension that it is used everywhere and most libraries don't work without it. It can be compiled directly with PHP. The official Docker images of PHP include it.

### DIFF
--- a/src/php/install.sh
+++ b/src/php/install.sh
@@ -169,7 +169,7 @@ install_php() {
         VERSION_CONFIG="--with-pear"
     fi
 
-    ./configure --prefix="${PHP_INSTALL_DIR}" --with-config-file-path="$PHP_INI_DIR" --with-config-file-scan-dir="$CONF_DIR" --enable-option-checking=fatal --with-curl --with-libedit --with-openssl --with-zlib --with-password-argon2 --with-sodium=shared "$VERSION_CONFIG" EXTENSION_DIR="$PHP_EXT_DIR";
+    ./configure --prefix="${PHP_INSTALL_DIR}" --with-config-file-path="$PHP_INI_DIR" --with-config-file-scan-dir="$CONF_DIR" --enable-option-checking=fatal --with-curl --with-libedit --enable-mbstring --with-openssl --with-zlib --with-password-argon2 --with-sodium=shared "$VERSION_CONFIG" EXTENSION_DIR="$PHP_EXT_DIR";
 
     make -j "$(nproc)"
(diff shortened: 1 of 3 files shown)


src/php/install.sh
```diff
@@ -169,7 +169,7 @@ install_php() {
         VERSION_CONFIG="--with-pear"
     fi
```
samruddhikhandale, line 171: Can you help bump the minor version in here? thanks! 

https://github.com/devcontainers/features/blob/2b778155760bf3e248bce5ca998fcc487ec8cd80/src/php/devcontainer-feature.json#L3

src/php/install.sh
```diff
@@ -169,7 +169,7 @@ install_php() {
...
 
-    ./configure --prefix="${PHP_INSTALL_DIR}" --with-config-file-path="$PHP_INI_DIR" --with-config-file-scan-dir="$CONF_DIR" --enable-option-checking=fatal --with-curl --with-libedit --with-openssl --with-zlib --with-password-argon2 --with-sodium=shared "$VERSION_CONFIG" EXTENSION_DIR="$PHP_EXT_DIR";
+    ./configure --prefix="${PHP_INSTALL_DIR}" --with-config-file-path="$PHP_INI_DIR" --with-config-file-scan-dir="$CONF_DIR" --enable-option-checking=fatal --with-curl --with-libedit --enable-mbstring --with-openssl --with-zlib --with-password-argon2 --with-sodium=shared "$VERSION_CONFIG" EXTENSION_DIR="$PHP_EXT_DIR";
```
samruddhikhandale, line 172: Can you help add a test to validate this change? 

Adding a check to [test/php/test.sh](https://github.com/devcontainers/features/blob/main/test/php/test.sh) should be fine.

samruddhikhandale (commented):
Thanks for adding this. Left few comments.

ℹ️  The [mcr.microsoft.com/devcontainers/php](https://github.com/devcontainers/images/tree/main/src/php) image also includes this extension. Nice to have the same consistency here.

julien-boudry:

@samruddhikhandale I just push the test and the minor version. Hope I have done it well.

samruddhikhandale (approved):
great, thanks!

ℹ️  There's a test failure which also happens on main and not an issue with this PR changes.
https://github.com/devcontainers/features/actions/runs/3836439825/jobs/6541495566

PR #380, opened 2023-01-05
Install gawk RPM

markphip:
This script requires awk and it may not be available on a minimal distro like Mariner. I tested with Mariner, Fedora and Rocky Linux. Theoretically could also add this for Debian and Alpine based distros but I leave that for you to decide.

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -146,6 +146,7 @@ install_debian_packages() {
 # RedHat / RockyLinux / CentOS / Fedora packages
 install_redhat_packages() {
     local package_list="\
+        gawk \
         openssh-clients \
         gnupg2 \
         iproute \
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.0.3",
+    "version": "2.0.4",
     "name": "Common Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",


src/common-utils/main.sh
```diff
@@ -146,6 +146,7 @@ install_debian_packages() {
 # RedHat / RockyLinux / CentOS / Fedora packages
 install_redhat_packages() {
     local package_list="\
+        gawk \
```
samruddhikhandale, line 149: > This script requires awk and it may not be available on a minimal distro like Mariner.

Would you mind adding a test to validate this change? You'd add a check to this test file -https://github.com/devcontainers/features/blob/main/test/common-utils/mariner.sh

src/common-utils/main.sh
```diff
@@ -146,6 +146,7 @@ install_debian_packages() {
 # RedHat / RockyLinux / CentOS / Fedora packages
 install_redhat_packages() {
```
samruddhikhandale, line 147: Can you bump the minor version in here?

https://github.com/devcontainers/features/blob/25942c4359491a0ead6cb755d6ce41e41ee0be05/src/common-utils/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks, left some comments!

src/common-utils/main.sh
markphip: Honestly, I am bit confused by how I missed this before. Maybe the Mariner upstream container has changed but it seems unlikely. The existing test ought to cover this because what I am seeing is something like this when trying to use common-utils with Mariner base image:

```
[2023-01-05T18:13:47.541Z] #15 42.06 ./main.sh: line 328: awk: command not found
[2023-01-05T18:13:47.542Z] 

[2023-01-05T18:13:47.543Z] #15 42.07 ERROR: Feature "Common Utilities" (ghcr.io/devcontainers/features/common-utils) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/common-utils for help troubleshooting this error.
```

samruddhikhandale: https://github.com/devcontainers/features/blob/main/test/common-utils/scenarios.json#L67 

The test scenarios currently passes because the `remoteUser` value is not `automatic`, hence, it doesn't need to execute `awk`. You may have `automatic` ?

markphip: Ah, thanks. That explains it. I think the one place I had used this I specified "vscode" as user as well. Which explains why it did not fail.

I assume this change will resolve that problem?

samruddhikhandale (approved):

src/common-utils/main.sh
samruddhikhandale: Yep!

PR #388, opened 2023-01-09
[common-utils] Fix permission of scripts in containers

### LINKED ISSUE #387: The devcontainer-info script in common-utils 2 cannot be run by non-root users
It looks like the script's permissions are being updated by the Dev Container CLI?
devcontainers/cli#263

eitsupi:
Fix #387

The Dev Container CLI changes permission when copying scripts into the container and read permissions are removed for the non-root users.
Read permission must be added after copying the scripts into the container.

### DIFF
--- a/test/common-utils/devcontainer-info.sh
+++ b/test/common-utils/devcontainer-info.sh
@@ -5,15 +5,20 @@ set -e
 # Optional: Import test library
 source dev-container-features-test-lib
 
+check_info() {
+    local info=$1
+    check "devcontainer-info ${info}" sh -c "devcontainer-info | grep test-${info}"
+}
+
 # Definition specific tests
-check "devcontainer-info version" sh -c "devcontainer-info | grep test-version"
-check "devcontainer-info id" sh -c "devcontainer-info | grep test-build"
-check "devcontainer-info variant" sh -c "devcontainer-info | grep test-variant"
-check "devcontainer-info repository" sh -c "devcontainer-info | grep test-repository"
-check "devcontainer-info release" sh -c "devcontainer-info | grep test-release"
-check "devcontainer-info revigion" sh -c "devcontainer-info | grep test-revision"
-check "devcontainer-info timestamp" sh -c "devcontainer-info | grep test-time"
-check "devcontainer-info url" sh -c "devcontainer-info | grep test-url"
+check_info "version"
+check_info "id"
+check_info "variant"
+check_info "repository"
+check_info "release"
+check_info "revision"
+check_info "time"
+check_info "url"
 
 # Report result
 reportResults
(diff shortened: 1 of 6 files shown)


test/common-utils/devcontainer-info/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM mcr.microsoft.com/devcontainers/base:debian
```
samruddhikhandale, line 1:
```suggestion
FROM ubuntu:focal
```

The latest MCR image is built with `common-utils:1`, hence the `devcontainer-info` command is correctly configured. The newly added tests in this PR still pass against it without the `chmod +rx` changes.

Can we change the image here to avoid the false positives? thanks!

test/common-utils/devcontainer-info.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+check_info() {
+    local info=$1
+    check "devcontainer-info ${info}" sh -c "devcontainer-info | grep test-${info}"
+}
+
+# Definition specific tests
```
samruddhikhandale, line 13: Can we add a check to validate the non user?

> check "user" bash -c "whoami | grep vscode"

test/common-utils/scenarios.json
```diff
@@ -114,5 +114,13 @@
                 "configureZshAsDefaultShell": true
             }
         }
+    },
+    "devcontainer-info": {
+        "build": {
+            "dockerfile": "Dockerfile"
+        },
+        "features": {
```
samruddhikhandale, line 122:
```suggestion
        "features": {
            "common-utils": {
                "username": "vscode",
                "userUid": "1000",
                "userGid": "1000",
                "upgradePackages": true,
                "installZsh": true
            }
        },
        "remoteUser": "vscode"
```

test/common-utils/devcontainer-info/Dockerfile
samruddhikhandale: Have a PR out to make that version bump switch for the images. But I'll be waiting to merge that until this PR gets merged.

https://github.com/devcontainers/images/pull/307/files#r1065017365

samruddhikhandale (changes requested):
Thanks for the fix, appreciate it! ✨ 

Left some minor comments.

eitsupi:
Thanks for your review.
Fixed these and CI on my fork passed!
https://github.com/eitsupi/devcontainers-devcontainer-features/actions/runs/3878411868/jobs/6614551708

samruddhikhandale (approved):
🚀

eitsupi:
Could you merge this?

PR #395, opened 2023-01-10
Git: Use sysconfdir=/etc for git built from source

samruddhikhandale:
Generally, libraries/tools expects the `gitconfig` to be located at `/etc/gitconfig`. However, when 'git' is built from source, it uses the `/usr/local` prefix & creates a `gitconfig` file at `${prefix}/etc/config` ie `/usr/local/etc/config` .

Building with `sysconfdir=/etc` to fix this issue.


jkeech (changes requested):
Instead of adding a symlink for the git config, I think it would be much better to build `git` from source with the correct prefix set to target a global installation. if we set `prefix` to be `/usr`, the system config path will automatically be [derived correctly](https://github.com/git/git/blob/master/Makefile#L1458-L1464), which will also set up system-level [gitattributes](https://github.com/git/git/blob/master/Makefile#L619-L620) and other files properly.

samruddhikhandale:
> Instead of adding a symlink for the git config, I think it would be much better to build `git` from source with the correct prefix set to target a global installation. if we set `prefix` to be `/usr`, the system config path will automatically be [derived correctly](https://github.com/git/git/blob/master/Makefile#L1458-L1464), which will also set up system-level [gitattributes](https://github.com/git/git/blob/master/Makefile#L619-L620) and other files properly.

@jkeech As `git` is part of the distro. Shouldn't we always install tools which are not managed by OS in `/usr/local/*` ? Also, isn't that the reason `/usr/local/bin` always exist in `PATH` ? 
Hence, the correct location of `gitconfig` will be `/usr/local/etc` 🤔 

That's the reason I avoided changing the `prefix` to `/usr`.

jkeech:
> @jkeech As `git` is part of the distro. Shouldn't we always install tools which are not managed by OS in `/usr/local/*` ? Also, isn't that the reason `/usr/local/bin` always exist in `PATH` ?
> Hence, the correct location of `gitconfig` will be `/usr/local/etc`

I thought the point of this PR is to make building `git` from source behave the same as installing from `apt` (and in fact to replace a potentially previously-installed version of `git` from `apt`). If so, we'll want to build git with the `/usr` prefix and make sure it's installed into `/usr`.

From https://www.pathname.com/fhs/pub/fhs-2.3.html#USRLOCALLOCALHIERARCHY:
>Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr. [[27]](https://www.pathname.com/fhs/pub/fhs-2.3.html#FTN.AEN1450)

If we really don't want to modify `prefix`, then `git` does allow you to directly set `sysconfdir` to `/etc` which would be a more robust solution than symlinking the system config file after `git` is built/installed.

Chuxel:
> > @jkeech As `git` is part of the distro. Shouldn't we always install tools which are not managed by OS in `/usr/local/*` ? Also, isn't that the reason `/usr/local/bin` always exist in `PATH` ?
> > Hence, the correct location of `gitconfig` will be `/usr/local/etc`
> 
> I thought the point of this PR is to make building `git` from source behave the same as installing from `apt` (and in fact to replace a potentially previously-installed version of `git` from `apt`). If so, we'll want to build git with the `/usr` prefix and make sure it's installed into `/usr`.
> 
> From https://www.pathname.com/fhs/pub/fhs-2.3.html#USRLOCALLOCALHIERARCHY:
> 
> > Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr. [[27]](https://www.pathname.com/fhs/pub/fhs-2.3.html#FTN.AEN1450)
> 
> If we really don't want to modify `prefix`, then `git` does allow you to directly set `sysconfdir` to `/etc` which would be a more robust solution than symlinking the system config file after `git` is built/installed.

@jkeech Installing under `/` will have very unpredictable behaviors and should not be done.  This is why `/usr/local` exists to begin with. If it doesn't fully break apt, it can result in strange behaviors like `apt-get update && apt-get upgrade` actually downgrading git. `/` is only for things managed by the distro.

Chuxel:
One possible way to handle this would be to generate a `deb` and let apt manage the install, but that's quite a bit more work.

jkeech:
@Chuxel, that's where the `sysconfdir` build variable comes into play when building `git` from source. It can be modified independently of `prefix` if we want to keep the prefix as `/usr/local`. I'm just not sure if there are other things we are missing if we only modify the `sysconfdir`. We got reports about the system level gitconfig file path changing and breaking tools when the feature builds from source, so we know that's one specific problem, but there might be other subtle differences beyond that. The closer we can get to the same behavior as the `git` build that's installed by `apt`, the better IMO.

>If we really don't want to modify prefix, then git does allow you to directly set sysconfdir to /etc which would be a more robust solution than symlinking the system config file after git is built/installed.

* https://github.com/git/git/blob/master/Makefile#L619-L620
* https://github.com/git/git/blob/master/Makefile#L1458-L1464

jkeech (approved):

PR #398, opened 2023-01-12
CommonUtils - ensure home directory exists.

Clockwork-Muse:
The common utils attempts to copy certain config files into the user's home directory... without first ensuring the directory exists!
Create the user's home directory if it doesn't exist.

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -383,11 +383,10 @@ if [ "${USERNAME}" = "root" ]; then
     user_rc_path="/root"
 else
     user_rc_path="/home/${USERNAME}"
-fi
-
-if [ ! -d "${user_rc_path}" ]; then
-    mkdir -p "${user_rc_path}"
-    chown ${USERNAME}:${group_name} "${user_rc_path}"
+    if [ ! -d "${user_rc_path}" ]; then
+        mkdir -p "${user_rc_path}"
+        chown ${USERNAME}:${group_name} "${user_rc_path}"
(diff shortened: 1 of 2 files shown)


src/common-utils/main.sh
```diff
@@ -385,6 +385,11 @@ else
     user_rc_path="/home/${USERNAME}"
 fi
 
+if [ ! -d "${user_rc_path}" ]; then
```
samruddhikhandale, line 388: Nit: Can we move this block after L:385? 

https://github.com/devcontainers/features/blob/1a3145d6c52a6b8d319a5639acdcff408f721c67/src/common-utils/main.sh#L385

`/root` will always exists so it's cleaner to add it in there.

src/common-utils/main.sh
```diff
@@ -385,6 +385,11 @@ else
...
+    mkdir -p "${user_rc_path}"
+    chown ${USERNAME}:${group_name} "${user_rc_path}"
+fi
+
```
samruddhikhandale, line 392: Can you bump the patch version? Thanks

https://github.com/devcontainers/features/blob/1a3145d6c52a6b8d319a5639acdcff408f721c67/src/common-utils/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks! Left minor comments

Clockwork-Muse:

samruddhikhandale (approved):
Thank you! 🚀

PR #402, opened 2023-01-15
[docker-outside-of-docker] Change owner of source socket if non-root access is enabled

amartincic:
This change will fix the permission problem of running the Docker command in the non-root mode. When the dood feature was executed as part of the devcontainer specifying the non-root mode and the non-root user, such as 'www' in the example below, only the target socket would get the 'www' owner, which is not sufficient since the docker.sock (target socket) is merely a symbolic link to docker-host.sock (source socket).

```
www@a73c2f1d2a7f:/var/www$ whoami
www

www@a73c2f1d2a7f:/var/www$ docker ps
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json": dial unix /var/run/docker.sock: connect: permission denied

root@a73c2f1d2a7f:/var/run# ls -la
total 12
drwxr-xr-x 1 root root 4096 Jan 15 13:09 .
drwxr-xr-x 1 root root 4096 Jan 15 13:09 ..
srwxrwxr-x 1 root root    0 Jan 12 14:22 docker-host.sock
lrwxrwxrwx 1 www  root   25 Jan 15 13:09 docker.sock -> /var/run/docker-host.sock
drwxrwxrwt 2 root root 4096 Jan  9 01:00 lock
-rw-rw-r-- 1 root utmp    0 Jan  9 01:00 utmp
 ```

### DIFF
--- a/src/docker-outside-of-docker/install.sh
+++ b/src/docker-outside-of-docker/install.sh
@@ -86,7 +86,7 @@ find_version_from_git_tags() {
     local repository=$2
     local prefix=${3:-"tags/v"}
     local separator=${4:-"."}
-    local last_part_optional=${5:-"false"}    
+    local last_part_optional=${5:-"false"}
     if [ "$(echo "${requested_version}" | grep -o "." | wc -l)" != "2" ]; then
         local escaped_separator=${separator//./\\.}
         local last_part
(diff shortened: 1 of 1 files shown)


amartincic:

samruddhikhandale (changes requested):
Thanks for the fix, left a minor comment.

AlvaroRausell:
Closing as abandoned. Feel free to reopen if need be

PR #404, opened 2023-01-16
checkout action update

steled:
- updated the action checkout from v2 to v3 as required by action warning
- this PR should fix the warning that appears at some actions:

```bash
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```


steled:

samruddhikhandale (approved):
Nice, thanks!

Issue #413, opened 2023-01-21
Nix: v2.13.0+ are Missing Signatures


dylhack:
For some reason the recently published 2.13.0 and 2.13.1 are missing signature files (`install.asc`) causing curl to 404 and the devcontainer to fail.

 https://github.com/devcontainers/features/blob/7a3a9c5fcaa59cf4d7dbbcece47094a6d642a9b0/src/nix/install.sh#L64

dylhack:
https://github.com/NixOS/nix/issues/7652

PR #414, opened 2023-01-21
features/nix: remove PGP check

dylhack:
Nix abruptly removed the signature file (https://github.com/NixOS/nix/pull/7411), so the current install script will no longer work.

fixes: #413

### DIFF
--- a/src/nix/install.sh
+++ b/src/nix/install.sh
@@ -12,12 +12,6 @@ FLAKEURI="${FLAKEURI:-""}"
 EXTRANIXCONFIG="${EXTRANIXCONFIG:-""}"
 USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
 
-# Nix keys for securely verifying installer download signature per https://nixos.org/download.html#nix-verify-installation
-NIX_GPG_KEYS="B541D55301270E0BCF15CA5D8170B4726D7198DE"
-GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com
-keyserver hkps://keys.openpgp.org
-keyserver hkp://keyserver.pgp.com"
-
 if [ "$(id -u)" -ne 0 ]; then
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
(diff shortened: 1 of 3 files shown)


src/nix/install.sh
```diff
@@ -61,10 +55,6 @@ else
         find_prev_version_from_git_tags VERSION https://github.com/NixOS/nix "tags/"
         curl -sSLf -o "${tmpdir}/install-nix" https://releases.nixos.org/nix/nix-${VERSION}/install
     fi
-    curl -sSLf -o "${tmpdir}/install-nix.asc" https://releases.nixos.org/nix/nix-${VERSION}/install.asc
-    cd "${tmpdir}"
-    receive_gpg_keys NIX_GPG_KEYS
```
samruddhikhandale, line 66: Can we remove this function? 

https://github.com/devcontainers/features/blob/3921f5917148583af38d1d96a1b7e0f31d22181c/src/nix/utils.sh#L86

src/nix/install.sh
```diff
@@ -12,12 +12,6 @@ FLAKEURI="${FLAKEURI:-""}"
 EXTRANIXCONFIG="${EXTRANIXCONFIG:-""}"
 USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
 
-# Nix keys for securely verifying installer download signature per https://nixos.org/download.html#nix-verify-installation
```
samruddhikhandale, line 15: Can we bump the patch version? 

https://github.com/devcontainers/features/blob/3921f5917148583af38d1d96a1b7e0f31d22181c/src/nix/devcontainer-feature.json#L3

samruddhikhandale (commented):
Thanks for the fix, left minor comments.

Chuxel:
@samruddhikhandale This is a big enough break, I might suggest merging and making the requested tweaks ourselves.

src/nix/install.sh
dylhack: resolved in aa00f714313489f3d3eb2c3c738870023cf6302d

src/nix/install.sh
dylhack: resolved in aa00f714313489f3d3eb2c3c738870023cf6302d

dylhack:

samruddhikhandale (approved):
Thanks! 🚀

PR #421, opened 2023-01-25
[common-utils] Fix nonempty .bashrc being restored

levinwinter:
The user's dotfiles shall only be restored to their defaults if they do not exist or are empty. A missing negation caused the files to be overwritten even when they were nonempty.

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -393,7 +393,7 @@ fi
 possible_rc_files=( ".bashrc" ".profile" ".zshrc" )
 for rc_file in "${possible_rc_files[@]}"; do
     if [ -f "/etc/skel/${rc_file}" ]; then
-        if [ ! -e "${user_rc_path}/${rc_file}" ] || [ -s "${user_rc_path}/${rc_file}" ]; then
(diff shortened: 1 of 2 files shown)


src/common-utils/main.sh
```diff
@@ -393,7 +393,7 @@ fi
 possible_rc_files=( ".bashrc" ".profile" ".zshrc" )
 for rc_file in "${possible_rc_files[@]}"; do
     if [ -f "/etc/skel/${rc_file}" ]; then
-        if [ ! -e "${user_rc_path}/${rc_file}" ] || [ -s "${user_rc_path}/${rc_file}" ]; then
```
samruddhikhandale, line 396: Can we bump the patch version here? Thanks!

samruddhikhandale (changes requested):
Thanks for fixing this, left a minor comment.

levinwinter:

src/common-utils/main.sh
levinwinter: Sure, I forgot about that! Feel free to squash the additional commit when merging the PR.

samruddhikhandale (approved):
Thanks!

PR #427, opened 2023-01-28
Install PowerShell Modules

### LINKED ISSUE #424: [Powershell] Allow a list of modules to be installed
It would be nice if you could define a list of modules that should be installed after PowerShell was installed.

As an example, we could define

```json
"features": {
    "ghcr.io/devcontainers/features/powershell:1": {
            "module": ["Az.Resources"]
    }
}
```

This would prevent having to add a postCreateCommand
```
Install-Module -Name Az.Resources -AllowClobber -Scope CurrentUser
```

### LINKED ISSUE #425: [PowerShell] Install the bicep cli when Powershell is installed
PowerShell does not install bicep by default, which is needed if you are using it to do simple Azure deploys - (this is much smaller and quicker to install than the whole az cli which includes bicep)

[https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/install#azure-powershell](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/install#azure-powershell)

Suggested default feature
```json
  "features": {
    "ghcr.io/devcontainers/features/powershell:1": {
        "installBicep": false
    }
  }
```

ross-p-smith:
This PR adds to the current PowerShell feature and allows the user to define a comma-separated list of Modules that installed after PowerShell is installed. 

Fixes #424


src/powershell/README.md
```diff
@@ -16,6 +16,8 @@ Installs PowerShell along with needed dependencies. Useful for base Dockerfiles
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Select or enter a version of PowerShell. | string | latest |
+| modules | Optional comma separated list of PowerShell modules to install. | string | - |
```
samruddhikhandale, line 19: Thanks for being considerate. Once, this PR is merged, an action would autogenerate this doc with similar contents.

Can we revert this to avoid merge conflicts?

src/powershell/install.sh
```diff
@@ -13,6 +13,8 @@ set -e
 rm -rf /var/lib/apt/lists/*
 
 POWERSHELL_VERSION=${VERSION:-"latest"}
+POWERSHELL_MODULES=${MODULES}
```
samruddhikhandale, line 16:
```suggestion
POWERSHELL_MODULES="${MODULES}"
```

src/powershell/install.sh
```diff
@@ -149,6 +151,32 @@ if [ "${use_github}" = "true" ]; then
...
 fi
 
+# If PowerShell modules are requested, loop through and install 
+if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
+    echo "Installing PowerShell Modules: ${POWERSHELL_MODULES}"
+    modules=(`echo ${POWERSHELL_MODULES} | tr ',' ' '`)
+    for i in "${modules[@]}"
+    do
+        echo "Installing ${i}"
+        pwsh -Command "Install-Module -Name ${i} -AllowClobber -Force -Scope AllUsers"|| continue
```
samruddhikhandale, line 161:
```suggestion
        pwsh -Command "Install-Module -Name ${i} -AllowClobber -Force -Scope AllUsers" || continue
```

src/powershell/install.sh
```diff
@@ -149,6 +151,32 @@ if [ "${use_github}" = "true" ]; then
...
+    done
+fi
+
+if [ "${POWERSHELL_INSTALLBICEP}" = "true" ]; then
```
samruddhikhandale, line 165: Don't we need to install az powershell to make use of az bicep?

samruddhikhandale (changes requested):
Thanks for the PR, left some comments.

src/powershell/install.sh
ross-p-smith: We are installing standalone bicep, not az. My use case is actually because I want a mechanism to install a bicep file without az cli as it takes 90seconds and is a 1Gb in size. Using Powershell with bicep standalone and "Az.Resource" module is actually much smaller and much quicker than az cli 🤷‍♂️

ross-p-smith: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/install#azure-powershell

src/powershell/README.md
ross-p-smith: Sure - all done!

src/powershell/devcontainer-feature.json
```diff
@@ -14,8 +14,18 @@
...
             "description": "Select or enter a version of PowerShell."
-        }
-    },
+        },
+        "modules": {
+            "type": "string",
+            "default": "",
+            "description": "Optional comma separated list of PowerShell modules to install."
+          },
+        "installBicep": {
```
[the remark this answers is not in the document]
ross-p-smith: Removed Bicep

src/powershell/devcontainer-feature.json
```diff
@@ -14,8 +14,13 @@
             ],
             "default": "latest",
             "description": "Select or enter a version of PowerShell."
-        }
-    },
+        },
```
samruddhikhandale, line 17: Can we bump the patch version in here? https://github.com/devcontainers/features/blob/262a7424215b67d89f1ab26483ac4cad90d87dcf/src/powershell/devcontainer-feature.json#L3

samruddhikhandale (commented):
Left one minor comment, but looks good otherwise!

src/powershell/install.sh
ross-p-smith: No longer installing bicep

src/powershell/devcontainer-feature.json
ross-p-smith: Done!

samruddhikhandale (approved):
⚡

PR #433, opened 2023-01-31
added bash-completion

rsuchecki:
adding bash-completion package enables git command completion / autocomplete


samruddhikhandale:
Hi 👋 

I agree that "bash-completion" is helpful, however, we only add tools which are core to the Feature. In this case, this util is not essential to the `git` Feature and we'd like to avoid adding it.

let me know if you have other question!

rsuchecki:
Fair enough!

Issue #420, opened 2023-01-25
Rust feature uses unknown settings


chrmarti:
From https://github.com/microsoft/vscode-remote-release/issues/7901, the image's devcontainer.json has the same settings:

Followed the steps in the test plan item. When inspecting the machine settings I noticed that some of the settings are now unknown. may be we need to update the devcontainer.json

<img width="412" alt="image" src="https://user-images.githubusercontent.com/1931590/214284193-497f3b6e-0448-42d3-9b26-c86162548de6.png">

PR #434, opened 2023-01-31
Rust - Clean up unwanted settings

samruddhikhandale:
Closes - https://github.com/devcontainers/features/issues/420

### DIFF
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.0.10",
+    "version": "1.0.11",
     "name": "Rust",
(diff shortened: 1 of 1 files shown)


src/rust/devcontainer-feature.json
```diff
@@ -36,14 +36,7 @@
                 "rust-lang.rust-analyzer",
                 "tamasfe.even-better-toml",
                 "serayuzgur.crates"
-            ],
-            "settings": {
-                "lldb.executable": "/usr/bin/lldb",
-                "files.watcherExclude": {
```
chrmarti, line 42: Would it make sense to keep the watcherExclude? The target folder is the Rust compiler's output folder and I think also caches dependencies.

chrmarti (approved):
👍

PR #443, opened 2023-02-07
Fix several issues with docker-outside-of-docker

felipecrs:
Fixes the following issues:

- `docker compose` does not work when `moby: false`
- `docker-compose` does not work when `moby: false` and `dockerDashCompose: v2`
- ~Fix `docker build: buildkit is enabled but buildx is not installed` when `moby: false`~

Improves:

- No longer needed to install docker-compose v1 if user hasn't asked for it `dockerDashComposeVersion: v2`. This speeds up the installation of this feature.
- No longer installs compose-switch if user wants docker-compose v1. It's pointless to install, and again, this speeds up the feature.


src/docker-outside-of-docker/install.sh
```diff
@@ -196,14 +196,14 @@ else
         apt-get -y install --no-install-recommends moby-cli${cli_version_suffix} moby-buildx
         apt-get -y install --no-install-recommends moby-compose || echo "(*) Package moby-compose (Docker Compose v2) not available for OS ${ID} ${VERSION_CODENAME} (${architecture}). Skipping."
     else
-        apt-get -y install --no-install-recommends docker-ce-cli${cli_version_suffix}
+        apt-get -y install --no-install-recommends docker-ce-cli${cli_version_suffix} docker-buildx-plugin docker-compose-plugin
```
samruddhikhandale, line 199: > docker-compose does not work when moby: false and dockerDashCompose: v2

`docker-compose` is a v1 command, so wondering why do you think it should exist when `dockerDashCompose: v2` ?

Hence, can we remove `docker-compose-plugin`?

src/docker-outside-of-docker/install.sh
```diff
@@ -196,14 +196,14 @@ else
...
     fi
 fi
 
 # Install Docker Compose if not already installed  and is on a supported architecture
 if type docker-compose > /dev/null 2>&1; then
     echo "Docker Compose already installed."
-else
+elif [ "${DOCKER_DASH_COMPOSE_VERSION}" = "v1" ]; then
```
samruddhikhandale, line 206: > Improves:
> 
> No longer needed to install docker-compose v1 if user hasn't asked for it dockerDashComposeVersion: v2. This speeds up the installation of this feature.
> No longer installs compose-switch if user wants docker-compose v1. It's pointless to install, and again, this speeds up the feature.

Looks like we are currently providing forward compatibility to people who want to install `v1`. Your improvements make sense to me, and would improve performance, however I wonder if people are now expecting to use `docker compose` and `docker-compose` with this Feature.

@Chuxel wondering if you have any other thoughts/concerns with this approach.
We'd need to bump the major version of the Feature if we decide to go ahead with the changes.

src/docker-outside-of-docker/install.sh
felipecrs: First, `docker-compose-plugin` is for `docker compose` to work. It should work no matter what users set for `dockerDashCompose`.

Now, if users set `dockerDashCompose: v2`, then yes, as the option name implies, `docker-compose` commands should be transmitted to docker compose v2.

src/docker-outside-of-docker/install.sh
felipecrs: The thing is, the currently already provides `docker compose` with always v2, and `docker-compose` for v1 or v2 depending on how users set `dockerDashCompose`. However, it only works with `moby: true`. I think you guys probably forgot to test `docker compose` or `dockerDashCompose: v2` when using `moby: false`.

felipecrs: To summarize: this PR changes nothing from users perspective, it only fixes bugs and optimizes stuff.

Chuxel: Yeah, `docker compose` can only be v2. There's a "switch" package you can use to decide whether `docker-compose` is mapped to v1 or v2.  So, I suspect @felipecrs is correct.

src/docker-outside-of-docker/install.sh
Chuxel: @samruddhikhandale `docker-compose-plugin` populates `docker compose` which is always v2. (Yes this is odd). They changed their packaging I think since docker-ce-cli used to include it and most people are using moby.

samruddhikhandale, line 199: Can you bump the patch version for this Feature?
https://github.com/devcontainers/features/blob/d20c30770746c2d88553d3aca54d433a02ab025a/src/docker-outside-of-docker/devcontainer-feature.json#L3

samruddhikhandale (commented):
> Fixes the following issues:
> 
> docker compose does not work when moby: false
> docker-compose does not work when moby: false and dockerDashCompose: v2
> Fix docker build: buildkit is enabled but buildx is not installed when moby: false

Can you help add tests to validate these fixes? Thank you! 
Doc for reference - https://github.com/devcontainers/cli/blob/main/docs/features/test.md

felipecrs:
@samruddhikhandale can you please approve the CI? I want to see if the tests are passing.

felipecrs:
I'm also basically reimplementing https://github.com/devcontainers/features/pull/478 as well. I don't know why that approach was taken since the packages can be installed from the distribution repos as well.

felipecrs:
And honestly I would not give the user the choice of installing buildx or not. But now it's too late anyway.

samruddhikhandale:
> @samruddhikhandale can you please approve the CI? I want to see if the tests are passing.

Looks like there are few scenario failures --> https://github.com/devcontainers/features/actions/runs/4569913747/jobs/8081761698?pr=443#step:4:5101

In case you are unaware, you can test the Features locally as well. See https://github.com/devcontainers/cli/blob/main/docs/features/test.md

> And honestly I would not give the user the choice of installing buildx or not (it should always be installed IMO). But now it's too late anyway.

I agree that `buildx` should be always installed, that's the reason why the default value for `installDockerBuildx` is `true`. Added it as an option in case someone wants to disable it, provides flexibility.

felipecrs:
I totally revamped the tests...

- Added some variants like Ubuntu 22.04 and Debian
- Added moby variants for all tests
- Added more tests around docker buildx
- Used symbolic links to avoid duplicating all the code of all these test variants
- Configure all tests which do not test docker-compose to use v2 since it's a lot faster to install

Also made the following changes:

- Removed the DOCKER_BUILDKIT environment variable (it now defaults to enabled on docker latest)
- Ensure buildx is not installed when installDockerBuildx=false
- As part of the tests, I figured out an issue when using moby. It seems that the group which moby creates hasn't the gid of 999 which is the default docker gid. I added a fix for such cases as well.

felipecrs:
@chrmarti I also think so. This speeds up the installation as well.

However, this is likely a breaking change. Should I do it, and bump the major version number?

chrmarti:
I remember breakage from Docker Desktop switching to v2 as minimal. We can do without a major version increase, I think. What do others think? @samruddhikhandale?

samruddhikhandale:
> I remember breakage from Docker Desktop switching to v2 as minimal. We can do without a major version increase, I think. What do others think? @samruddhikhandale?

Same, we could bump the minor version instead as the impact is minimal.

In my opinion, we should try to avoid bumping major versions as people usually have Features pinned to major version ; they would stay behind with security fixes/latest updates if they fail to update the version.

Besides that the PR looks great, @felipecrs I can help approve if you could help bump the minor version.

felipecrs:
@samruddhikhandale, done! :)

samruddhikhandale (approved):
Appreciate your contribution, thank you! ✨

PR #449, opened 2023-02-10
Turn off linting by default in VS Code for Python

brettcannon:
Linting is on by default if you specify the paths to any linters, so the previous settings would turn on **all** linters.

Also drop linters that don't see much use.

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -63,12 +63,13 @@
         "python.defaultInterpreterPath": "/usr/local/python/current/bin/python",
         "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
         "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
+        "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
+        "python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
         "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
-        "python.linting.flake8Enabled": false,
         "python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
-        "python.linting.mypyEnabled": false,
-        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint",
-        "python.linting.pylintEnabled": false
+        "python.linting.pycodestylePath": "/usr/local/py-utils/bin/pycodestyle",
+        "python.linting.pydocstylePath": "/usr/local/py-utils/bin/pydocstyle",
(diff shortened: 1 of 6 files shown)


src/python/install.sh
```diff
@@ -26,7 +26,7 @@ CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN="${CONFIGUREJUPYTERLABALLOWORIGIN:-""}"
 # alongside PYTHON_VERSION, but not set as default.
 ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
-DEFAULT_UTILS=("pylint" "flake8" "autopep8" "black" "yapf" "mypy" "pydocstyle" "pycodestyle" "bandit" "pipenv" "virtualenv")
+DEFAULT_UTILS=("pylint" "flake8" "autopep8" "black" "mypy" "pipenv" "virtualenv")
```
samruddhikhandale, line 29: From https://github.com/devcontainers/images/pull/418#issuecomment-1426490478, I agree with @bamurtaugh thoughts of keeping these utils as is for now. Let us know what you think!

src/python/devcontainer-feature.json
```diff
@@ -63,13 +63,12 @@
         "python.defaultInterpreterPath": "/usr/local/python/current/bin/python",
         "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
         "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
-        "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
-        "python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
         "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
+        "python.linting.flake8Enabled": false,
```
samruddhikhandale, line 67: Can we bump the patch version of this Feature? Thanks
https://github.com/devcontainers/features/blob/ad0b0cf5e9344131a0d72b614d138335f4bbd5b5/src/python/devcontainer-feature.json#L3

samruddhikhandale (commented):
Thanks, left some thoughts!

src/python/devcontainer-feature.json
brettcannon: How would you like the version number bumped: major, minor, or micro?

samruddhikhandale: We can bump as patch (ie micro)

brettcannon:
@samruddhikhandale I updated the PR to only remove settings related to the under-utilized tool support in the Python extension for VS Code and to turn linting off by default so users can choose which linter they want.

samruddhikhandale (approved):
Looks great, thank you! 🚢

PR #450, opened 2023-02-11
Add pnpm in Node

akihironagai:
I have added pnpm which is a package manager to Node features.

### DIFF
--- a/src/node/README.md
+++ b/src/node/README.md
@@ -1,7 +1,7 @@
 
-# Node.js (via nvm), yarn and pnpm (node)
+# Node.js (via nvm) and yarn (node)
 
-Installs Node.js, nvm, yarn, pnpm, and needed dependencies.
+Installs Node.js, nvm, yarn, and needed dependencies.
 
 ## Example Usage
 
--- a/src/node/README.md
+++ b/src/node/README.md
@@ -1,7 +1,7 @@
 
-# Node.js (via nvm) and yarn (node)
+# Node.js (via nvm), yarn and pnpm (node)
 
-Installs Node.js, nvm, yarn, and needed dependencies.
+Installs Node.js, nvm, yarn, pnpm, and needed dependencies.
 
 ## Example Usage
 
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -103,6 +103,13 @@ else
     apt-get -y install --no-install-recommends yarn
 fi
 
+# Install pnpm
+if type pnpm > /dev/null 2>&1; then
+    echo "pnpm already installed."
+else
+    curl -fsSL https://get.pnpm.io/install.sh | sh -
+fi
+
 # Adjust node version if required
 if [ "${NODE_VERSION}" = "none" ]; then
     export NODE_VERSION=
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -103,13 +103,6 @@ else
     apt-get -y install --no-install-recommends yarn
 fi
 
-# Install pnpm
-if type pnpm > /dev/null 2>&1; then
-    echo "pnpm already installed."
-else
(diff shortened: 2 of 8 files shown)


src/node/README.md
```diff
@@ -1,7 +1,7 @@
 
-# Node.js (via nvm) and yarn (node)
+# Node.js (via nvm), yarn and pnpm (node)
```
samruddhikhandale, line 2: This README file is autogenerated from `devcontainer-feature.json` and will be overwritten once this PR is merged.

Can you update 👇 instead? https://github.com/devcontainers/features/blob/c8590bef3dce55bfc110fd3abdf6cd28e7d7e93f/src/node/devcontainer-feature.json#L4

src/node/README.md
```diff
@@ -1,7 +1,7 @@
...
 
-Installs Node.js, nvm, yarn, and needed dependencies.
+Installs Node.js, nvm, yarn, pnpm, and needed dependencies.
```
samruddhikhandale, line 4: Same here. Can you update 👇 instead? 

https://github.com/devcontainers/features/blob/c8590bef3dce55bfc110fd3abdf6cd28e7d7e93f/src/node/devcontainer-feature.json#L6

src/node/install.sh
```diff
@@ -103,6 +103,13 @@ else
     apt-get -y install --no-install-recommends yarn
 fi
 
+# Install pnpm
```
samruddhikhandale, line 106: Can you bump the minor version? 
https://github.com/devcontainers/features/blob/c8590bef3dce55bfc110fd3abdf6cd28e7d7e93f/src/node/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks, left some comments

src/node/install.sh
```diff
@@ -103,6 +103,13 @@ else
...
+if type pnpm > /dev/null 2>&1; then
+    echo "pnpm already installed."
+else
+    curl -fsSL https://get.pnpm.io/install.sh | sh -
```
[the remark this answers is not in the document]
akihironagai: I avoided curling by installing pnpm from npm.

However, when using nvm and multiple versions are installed, pnpm can only be executed in first installed environment.

src/node/install.sh
```diff
@@ -164,6 +164,13 @@ else
     fi
 fi
 
+# Install pnpm
+if type pnpm > /dev/null 2>&1; then
+    echo "pnpm already installed."
+else
+    npm install -g pnpm
```
samruddhikhandale, line 171: Can you help write a test to validate these changes? Thanks! 
Adding checks to existing scenarios should be fine too - https://github.com/devcontainers/features/tree/main/test/node

src/node/README.md
samruddhikhandale, line 2: Can we revert these changes to avoid merge conflicts?

src/node/install.sh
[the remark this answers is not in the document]
samruddhikhandale: thanks! 

> However, when using nvm and multiple versions are installed, pnpm can only be executed in first installed environment.

I think that's the way `pnpm` is currently setup. To install multiple versions of Node, one should use `pnpm` over `nvm` to make sure `pnpm` works with other versions or reinstall `pnpm` - https://pnpm.io/cli/env#use

This Feature, currently supports installing a single Node version, so I think should be fine. We can reiterate on this Feature if needed based on feedbacks.

src/node/README.md
akihironagai: Reverted.

src/node/install.sh
akihironagai: I've added tests.

samruddhikhandale (approved):
Nice ⚡

Issue #412, opened 2023-01-20
Avoid piping curl'ed scripts into bash


hajapy:
In 3 installers (java, node, ruby), the install.sh script includes a step where a script is curl'ed and piped into bash: 
https://github.com/devcontainers/features/blob/7a3a9c5fcaa59cf4d7dbbcece47094a6d642a9b0/src/java/install.sh#L155
https://github.com/devcontainers/features/blob/7a3a9c5fcaa59cf4d7dbbcece47094a6d642a9b0/src/node/install.sh#L121
https://github.com/devcontainers/features/blob/7a3a9c5fcaa59cf4d7dbbcece47094a6d642a9b0/src/ruby/install.sh#L211

Those could be entrypoints for supply chain attacks. I think it would be preferable if these tools could be installed using a step that does checksum verification on the bash scripts before running them.

samruddhikhandale:
Hi 👋 

Thanks for pointing it out. We have added checksum verification for the Features which installs from artifacts, but looks like they were missed for script verifications.

samruddhikhandale:
https://github.com/devcontainers/features/blob/7a3a9c5fcaa59cf4d7dbbcece47094a6d642a9b0/src/java/install.sh#L155

Note: SDKMAN started providing checksum verification for the script.
Ref: https://github.com/sdkman/sdkman-cli/issues/1040#issuecomment-1023455180

akihironagai:
I have found alternatives to installing Node safely.

The first method is to use [NodeSource's manual installation](https://github.com/nodesource/distributions#debmanual).
The second method is to download a source code from [nodejs.org](https://nodejs.org/en/download/) and then [verify](https://github.com/nodejs/node#verifying-binaries) it.
The third method is to install pnpm from the [pnpm standalone script (it can be verified)](https://github.com/pnpm/get.pnpm.io) and then use [pnpm env](https://pnpm.io/cli/env).

Using the above two methods, it is impossible to manage multiple Node versions.
Using the third method makes corepack unavailable. Currently, corepack is the recommended way to install yarn berry.

Currently, the feature uses nvm, but there is no reason to manage multiple Nodes.
(I think that managing multiple sources in a single container is against the purpose of Dev Container.)

For this reason, I would suggest using the above two means.
Choosing between them is a matter of preference.

PR #470, opened 2023-03-02
Add a feature for Haskell (GHC)

brendandburns:


bamurtaugh:
Hi @brendandburns, thanks so much for the PR!

We try to keep a small subset of languages in this repo. We then absolutely welcome further community-published contributions to our [Feature index](https://github.com/devcontainers/devcontainers.github.io/blob/gh-pages/_data/collection-index.yml). 

Would you be interested in publishing this Feature in your own repo (i.e. at https://github.com/brendandburns/ghc-feature) and then opening a PR to update our index? We have a [sample repo](https://github.com/devcontainers/feature-starter) that walks through the process.

Let us know if you have any questions, thank you!

brendandburns:
@bamurtaugh I'm happy to put it in a separate repo and indeed, I've already published some features here, so I'm familiar with the process.

https://github.com/dev-wasm/dev-wasm-feature

However in this case it seems odd to have an independent repo for a (relatively) mainstream language, since it will lead to a variety of different people implementing the same things. How does the team decide which languages to support?

I'm happy to sign up to be a maintainer of this. However, if it truly doesn't fit in this repo, I will do it as a separate feature.

bamurtaugh:
> However in this case it seems odd to have an independent repo for a (relatively) mainstream language, since it will lead to a variety of different people implementing the same things. How does the team decide which languages to support?

Thanks for opening this dialogue, this is a great question.

For broader context (especially if other folks may discover this PR):

We previously hosted all dev container templates and features in one large mono-repo: https://github.com/microsoft/vscode-dev-containers/tree/main/containers. This became difficult to manage as our maintainer group doesn't have expertise in each language, and it became especially challenging in the case of issues that needed quick turnaround time.

When we moved to the open spec, we saw that as a great opportunity to move to a more distributed, community-involved model, hoping to avoid the mono-repo issues we had faced for a few years. When determining which languages to bring over, we looked at a variety of data points and usage trends to determine overall language popularity and usage of our existing artifacts (i.e. templates and features). 

> I'm happy to sign up to be a maintainer of this. However, if it truly doesn't fit in this repo, I will do it as a separate feature.

As we've been discussing, we'd love for you to be a maintainer! We're actively working on this now. Thank you so much for your contributions so far and your interest in maintaining longer term, we're really excited to collaborate with you.

@samruddhikhandale @joshspicer if you have any specific feedback on the contents of this Feature, it'd be great to get your review, thank you!

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
+{
+    "id": "haskell",
+    "version": "0.0.1",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.0.0",
```

We have been publishing all the Features in this repo from major version `v1`. Can we keep this in sync? Thanks!

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
...
+    "name": "Glasgow Haskell Compiler",
+    "documentationURL": "https://github.com/brendandburns/ghc-feature",
```
samruddhikhandale, line 5: README files for this repo are auto-generated from `devcontainers-features.json` contents. Can we update this link? Feel free to add any other details to the `NOTES.md` file

```suggestion
    "documentationURL": "https://github.com/devcontainers/features/tree/main/src/haskell",
```

Ref: https://github.com/devcontainers/features/blob/main/src/anaconda/devcontainer-feature.json

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
...
+            "proposals": [
+                "latest",
+                "recommended",
+                "9.4.4",
+                "9.2.7"
+            ],
+            "default": "recommended",
+            "description": "Select or enter a GHC Version to install"
+        },
+        "cabal_version": {
```
samruddhikhandale, line 19:
```suggestion
        "cabalVersion": {
```

From https://containers.dev/implementors/features/#options-property, we use camel casing! Can we update this?

Note: The ID becomes the name of the environment variable in all caps

We'd need to change the `install.sh` to adapt it.
eg. CABAL_VERSION="${CABALVERSION:-"recommended"}"

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
...
+            "type": "string",
+            "proposals": [
+                "latest",
+                "recommended",
+                "3.6.2.0"
+            ],
+            "default": "recommended",
+            "description": "Select or enter a Cabal Version to install"
+        },
+        "hls_version": {
```
samruddhikhandale, line 29: Same here! 

```suggestion
        "hlsVersion": {
```

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
...
+            "type": "string",
+            "proposals": [
+                "latest",
+                "recommended",
+                "1.9.0.0"
+            ],
+            "default": "recommended",
+            "description": "Select or enter a Haskell Language Server Version to install"
+        },
+        "stack_version": {
```
samruddhikhandale, line 39:
```suggestion
        "stackVersion": {
```

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,54 @@
...
+            "proposals": [
+                "latest",
+                "recommended",
+                "2.9.3"
+            ],
+            "default": "recommended",
+            "description": "Select or enter a Stack Version to install"
+        }
+    },
+    "mounts": [],
```
samruddhikhandale, line 50: Feel free to remove this property as it's unused?

src/haskell/install.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e
+
+BOOTSTRAP_HASKELL_GHC_VERSION="${VERSION:-"recommended "}"
+
+# Maybe install curl, gcc, make
+for x in curl gcc make; do
+    which $x > /dev/null || (apt update && apt install $x -y -qq)
+done
+
+curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_MINIMAL=1 sh
```
samruddhikhandale, line 11: Downloading files and executing them makes it prone to supply chain attacks. We have https://github.com/devcontainers/features/issues/412 opened which fixes current occurrences. 

Can we verify integrity of this file?

test/haskell/scenarios.json
```diff
@@ -0,0 +1,17 @@
...
+            "haskell": {
+                "version": "9.2.6"
+            }
+        }
+    },
+    "default_version": {
+        "image": "ubuntu:jammy",
+        "features": {
+            "haskell": {
+            }
```
samruddhikhandale, line 14:
```suggestion
            "haskell": {}
```

samruddhikhandale (changes requested):
Thank you for the contribution, we appreciate it 🎉 

Left some comments/thoughts!

src/haskell/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -e
+
+BOOTSTRAP_HASKELL_GHC_VERSION="${VERSION:-"recommended "}"
+
+# Maybe install curl, gcc, make
+for x in curl gcc make; do
+    which $x > /dev/null || (apt update && apt install $x -y -qq)
+done
```
eitsupi, line 9: It would be better to use a generic snippet to check other Features to do this.

src/haskell/devcontainer-feature.json
brendandburns: done.

src/haskell/devcontainer-feature.json
brendandburns: done.

test/haskell/scenarios.json
brendandburns: done.

src/haskell/install.sh
brendandburns: switched to direct download and sha verify.

src/haskell/devcontainer-feature.json
brendandburns: done

src/haskell/devcontainer-feature.json
brendandburns: done

src/haskell/devcontainer-feature.json
brendandburns: done.

src/haskell/devcontainer-feature.json
brendandburns: done.

brendandburns:
Comments addressed except one I didn't quite understand. Please take another look.

src/haskell/install.sh
```diff
@@ -1,23 +1,29 @@
 #!/usr/bin/env bash
 set -e
 
+GHCUP_VERSION="0.1.19.2"
```
samruddhikhandale, line 4: Wondering, is there a way to automatically find the latest version for `GHCUP` instead of pinning this? We would like to remove such dependencies that needs manual updates (if possible) 

eg. For git releases we have helper functions which looks at git tags and finds the latest version, hence we are always up-to-date without modifying scripts.

Doesn't look like https://github.com/haskell/ghcup/tags has tags updated. 😕

src/haskell/install.sh
```diff
@@ -1,23 +1,29 @@
...
+GHCUP_BIN="x86_64-linux-ghcup-${GHCUP_VERSION}"
```
samruddhikhandale, line 5:
```suggestion
# Source /etc/os-release to get OS info
. /etc/os-release
# Fetch host/container arch.
architecture="$(dpkg --print-architecture)"
GHCUP_BIN="${architecture}-linux-ghcup-${GHCUP_VERSION}"
```

src/haskell/install.sh
```diff
@@ -1,23 +1,29 @@
...
 # Maybe install curl, gcc, make
 for x in curl gcc make; do
     which $x > /dev/null || (apt update && apt install $x -y -qq)
 done
 
-curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_MINIMAL=1 sh
- 
 GHCUP_DIR=~/.ghcup/bin
+mkdir -p $GHCUP_DIR
+curl https://downloads.haskell.org/~ghcup/${GHCUP_VERSION}/x86_64-linux-ghcup-${GHCUP_VERSION} --output ${GHCUP_BIN}
```
samruddhikhandale, line 15:
```suggestion
curl https://downloads.haskell.org/~ghcup/${GHCUP_VERSION}/${architecture}-linux-ghcup-${GHCUP_VERSION} --output ${GHCUP_BIN}
```

src/haskell/install.sh
samruddhikhandale: I believe @eitsupi is referring to using `check_packages` helper function instead of this logic.
https://github.com/devcontainers/features/blob/main/src/docker-in-docker/install.sh#LL86-L91C2

I think it would be nice to use `check_packages` in this Feature as well for consistency.
(Would also need to copy `apt_get_update` along with `check_packages`)

test/haskell/default_version.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+EXPECTED_VERSION=$(~/.ghcup/bin/ghcup list | grep "ghc " | grep recommended | awk '{print $3}')
```
samruddhikhandale, line 9:
```suggestion
EXPECTED_VERSION=$(${_REMOTE_USER_HOME}/.ghcup/bin/ghcup list | grep "ghc " | grep recommended | awk '{print $3}')
```

For the failing MCR tests https://github.com/devcontainers/features/actions/runs/4369291649/jobs/7642925890#step:4:358

I think, ghcup is getting installed at `/root/*` but is the tests are checking at `/home/vscode/*`
The suggestion should help.

samruddhikhandale: Ooops wrong place!

src/haskell/install.sh
```diff
@@ -0,0 +1,29 @@
...
+GHCUP_VERSION="0.1.19.2"
+GHCUP_BIN="x86_64-linux-ghcup-${GHCUP_VERSION}"
+BOOTSTRAP_HASKELL_GHC_VERSION="${VERSION:-"recommended "}"
+
+# Maybe install curl, gcc, make
+for x in curl gcc make; do
+    which $x > /dev/null || (apt update && apt install $x -y -qq)
+done
+
+GHCUP_DIR=~/.ghcup/bin
```
samruddhikhandale, line 13:
```suggestion
GHCUP_DIR=${_REMOTE_USER_HOME}/.ghcup/bin
```

For the failing MCR tests https://github.com/devcontainers/features/actions/runs/4369291649/jobs/7642925890#step:4:358

I think, ghcup is getting installed at `/root/*` but is the tests are checking at `/home/vscode/*`
The suggestion should help.

src/haskell/install.sh
```diff
@@ -0,0 +1,29 @@
...
+mkdir -p $GHCUP_DIR
```
samruddhikhandale, line 14: I wonder if we should add a new user group `haskell` and chown & set sticky bit for this folder `.ghcup`

Ref: https://github.com/devcontainers/features/blob/main/src/ruby/install.sh#LL206-L209C7 &
https://github.com/devcontainers/features/blob/main/src/ruby/install.sh#L283-L285

samruddhikhandale: where 

> USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"

We have https://github.com/devcontainers/features/blob/main/src/ruby/install.sh#L39-L59 to automatically figure out the USERNAME

src/haskell/install.sh
```diff
@@ -0,0 +1,29 @@
...
+
+${GHCUP_DIR}/ghcup install ghc $BOOTSTRAP_HASKELL_GHC_VERSION
+${GHCUP_DIR}/ghcup install cabal $CABALVERSION
+${GHCUP_DIR}/ghcup install hls $HLSVERSION
+${GHCUP_DIR}/ghcup install stack $STACKVERSION
+
+${GHCUP_DIR}/ghcup set ghc $BOOTSTRAP_HASKELL_GHC_VERSION
+${GHCUP_DIR}/ghcup set cabal $CABALVERSION
+${GHCUP_DIR}/ghcup set hls $HLSVERSION
+${GHCUP_DIR}/ghcup set stack $STACKVERSION
```
samruddhikhandale, line 29:
```suggestion
${GHCUP_DIR}/ghcup set stack $STACKVERSION

# Clean up
rm -rf /var/lib/apt/lists/*

echo "Done!"
```

Trying to sync with the current Features :)

src/haskell/install.sh
brendandburns: We have to pin on a version if we want to do SHA256SUM validation. Otherwise the attacker can just change both the download and the SHA256SUM.

src/haskell/install.sh
brendandburns: Hrm, I'm not sure that those scripts do exactly what I want to do here. For example, they won't work correctly on a non-debian image where `gcc` is already installed.

Additionally, copying functions from one install.sh to another doesn't seem to be a great practice, is there a way to share common script snippets between features?

src/haskell/install.sh
samruddhikhandale: Makes sense 👍

src/haskell/install.sh
brendandburns: done.

src/haskell/install.sh
brendandburns: I'm not sure it's necessary, this will land in the home directory, but it doesn't really need a different username imho.

src/haskell/install.sh
brendandburns: Done.

src/haskell/install.sh
brendandburns: Done, though I just used the `arch` command instead because /etc/os-release returns `amd64` instead of `x86_64`

brendandburns:
@samruddhikhandale comments mostly addressed except the creating a new `haskell` user which doesn't seem necessary, and also the standardized package install.

I'm open to a more standard package install solution if it is 100% required, but it seems not super useful to just cut and paste the same thing into multiple locations. Additionally, I think we would need to edit it so that it doesn't require a `deb` based package manager.

But I'm motivated to get this merged :) so if those changes are required, let me know and I will make them.

src/haskell/install.sh
eitsupi: > I'm not sure that those scripts do exactly what I want to do here. For example, they won't work correctly on a non-debian image where `gcc` is already installed.

Note that this repository is focused on debian-based containers at this time. Most Features will only work with debian-based images.

https://github.com/devcontainers/features/blob/1449f416a0d5c65fbd7e7bfac041eb43f1ecd98d/README.md?plain=1#L32

> Additionally, copying functions from one install.sh to another doesn't seem to be a great practice, is there a way to share common script snippets between features?

See devcontainers/spec#209

eitsupi:
> But I'm motivated to get this merged :)

I don't think there is any chance at all that this repository will merge this PR.
Please check the README.

https://github.com/devcontainers/features/blob/1449f416a0d5c65fbd7e7bfac041eb43f1ecd98d/README.md?plain=1#L98-L111

So the standard practice is for you to create your own repository and manage it there.
For example, I manage the Features to install R in https://github.com/rocker-org/devcontainer-features.

But if you would prefer to manage this in a multi-person organization rather than managing this on your own, @jcbhmr and I are discussing in https://github.com/orgs/devcontainers-community/discussions/1 about creating such a repository. (I am not sure if such a repository or organization will really materialize.)

src/haskell/install.sh
eitsupi: > Doesn't look like [haskell/ghcup/tags](https://github.com/haskell/ghcup/tags) has tags updated. 😕

It seems to refer to an old repository. The new repository have tags.
https://github.com/haskell/ghcup-hs

eitsupi:
Cross ref microsoft/vscode-dev-containers#1705

jcbhmr:
@brendandburns  Would it be worth reaching out to https://github.com/haskell somehow? Maybe via https://github.com/haskell/meta to ask if they can add your code to like a haskell/devcontainer-features or similar repo? I know there has been [some discussion](https://github.com/orgs/devcontainers-community/discussions/1#discussioncomment-5260701) about what to do about non-`@devcontainers` features and stuff, and one of the ideas was to put the features under the tooling/language organization.

/cc @eitsupi

bamurtaugh:
Thanks so much everyone for the brainstorming and reviews here- I really appreciate everyone providing their feedback and collaborating to find the best path forward.

@brendandburns is joining us as a spec maintainer, and the thinking here is his direct spec involvement, plus long-standing container and Haskell expertise, allows the spec to host this Feature (avoids the mono-repo / language expertise issues of vscode-dev-containers). 

I also think the suggestion to work with Haskell to host the Feature directly is quite interesting - I really appreciate folks sharing that idea. I've opened this issue in the Haskell repo: https://github.com/haskell/meta/issues/4, and we'll continue sharing updates in this PR.

samruddhikhandale (approved):
Looks great, thank you so much! ✨ ✨

src/haskell/install.sh
```diff
@@ -0,0 +1,42 @@
...
+echo https://downloads.haskell.org/~ghcup/${GHCUP_VERSION}/${architecture}-linux-ghcup-${GHCUP_VERSION} --output ${GHCUP_BIN}
+curl https://downloads.haskell.org/~ghcup/${GHCUP_VERSION}/${architecture}-linux-ghcup-${GHCUP_VERSION} --output ${GHCUP_BIN}
+# echo "25b7fc417c1a811dd7ff439b67ea647a59cf5b8d71b274f97e917d50b2150d5b ${GHCUP_BIN}" | sha256sum --check --status
+
+mv ${GHCUP_BIN} $GHCUP_DIR/ghcup
+chmod a+x $GHCUP_DIR/ghcup
+
+export GHCUP_INSTALL_BASE_PREFIX=${_REMOTE_USER_HOME}
+
+${GHCUP_DIR}/ghcup install ghc $BOOTSTRAP_HASKELL_GHC_VERSION
```
hasufell, line 29: I suggest to not do all this manually, but use the bootstrap script, which allows fine grained control via env vars.

E.g.

```sh
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | BOOTSTRAP_HASKELL_NONINTERACTIVE=1 BOOTSTRAP_HASKELL_CABAL_VERSION=$CABALVERSION BOOTSTRAP_HASKELL_INSTALL_HLS=1 BOOTSTRAP_HASKELL_ADJUST_BASHRC=1 sh
```

This will also try do adjust shell config to add `~/.ghcup/bin` to PATH and most importantly install stack installation hooks that make sure HLS actually works with stack (there are ABI issues). See https://www.haskell.org/ghcup/guide/#strategy-2-stack-hooks-new-recommended

Right now the bootstrap script assumes you want latest of stack and HLS, but I can add environment variables to control that as well.

src/haskell/install.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+set -e
+set -o xtrace
+
+GHCUP_VERSION="0.1.19.2"
```
hasufell, line 5: Yeah, please use the bootstrap script to always get the latest ghcup. Pinning ghcup version is not recommended.

test/haskell/custom_version.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+check "version" ~/.ghcup/bin/ghc --version | grep 9.2.6
+
```
hasufell, line 10: So is `~/.ghcup/bin` in PATH?

src/haskell/install.sh
hasufell: Oh and the bootstrap script also creates a `~/.ghcup/env` file that can be sourced to get the correct PATH.

src/haskell/devcontainer-feature.json
```diff
@@ -0,0 +1,53 @@
+{
+    "id": "haskell",
+    "version": "1.0.0",
+    "name": "Glasgow Haskell Compiler",
+    "documentationURL": "https://github.com/devcontainers/features/tree/main/src/haskell",
+    "description": "Installs the Glasgow Haskell Compiler",
+    "options": {
+        "version": {
+            "type": "string",
+            "proposals": [
```
hasufell, line 10: I'm not too familiar with how features work. My feeling is that this list can become outdated quickly. For HLS, it already is. Would it not be enough to choose between recommended and latest? Can the user enter manual strings regardless?

benz0li:
[`ghcr.io/devcontainers-extra/features/haskell:2`](https://github.com/devcontainers-extra/features/tree/da7016845383b02be07278e5d52bc8fcccb29c02/src/haskell) exists for over a year now.

Can this be closed?

Cross reference:

* https://github.com/haskell/docker-haskell/issues/123#issuecomment-3075127969

hasufell:
@benz0li are your containers affiliated with this project?

benz0li:
> @benz0li are your containers affiliated with this project?

No. It is Haskell via `ghcup`.

hasufell:
I mean, I guess the point of this specific project is to provide a set of docker containers with a consistent UX/quality. 

So I guess you would have to contribute them here and get them reviewed.

benz0li:
> I mean, I guess the point of this specific project is to provide a set of docker containers with a consistent UX/quality.

True. [Extra Dev Container Features](https://github.com/devcontainers-extra/features) are not officially associated with or endorsed by the official [Devcontainers](https://containers.dev/) project.

> So I guess you would have to contribute them here and get them reviewed.

[ghcr.io/devcontainers-extra/features/haskell:2](https://github.com/devcontainers-extra/features/tree/main/src/haskell) is not my project. Ping @koralowiec.

benz0li:
IMHO it is not worth investing any more time in this PR.

See also https://github.com/devcontainers/features/pull/470#issuecomment-1464809355.

benz0li:
@brendandburns Please close this PR.

Issue #477, opened 2023-03-08
[docker-in-docker] Latest docker (v23) BuildKit does not work in Codespaces


sinedied:
Here's my `devcontainer.json`:

```
{
	"name": "Node.js",
	"image": "mcr.microsoft.com/devcontainers/javascript-node:0-18",

	"features": {
		"ghcr.io/devcontainers/features/docker-in-docker:1": {
			"version": "latest",
			"moby": "false",
			"dockerDashComposeVersion": "v2"
		},
	},
}
```

With such a configuration, I get an error that BuildKit isn't available and should be installed whenever I try to run "docker build".
I could reproduce it with either v1 or v2 of the feature.

Docker v23 is currently installed using "version": "latest" and have this error. When forcing to an older version, like "v20" it's working correctly (I did not try v21 and v22).

JasonTheDeveloper:
I've found the same issue with `docker-outside-of-docker`/`docker-from-docker`. It seems like in docker `v23.0.0` they made some changes to `buildkit` and `buildx` https://docs.docker.com/engine/deprecated/.

PR #478, opened 2023-03-08
Docker fix - BuildKit is enabled but the buildx component is missing or broken

samruddhikhandale:
Fixes - https://github.com/devcontainers/features/issues/477

> DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
            
More details - https://docs.docker.com/engine/deprecated/#legacy-builder-fallback


joshspicer (dismissed):
Looks great to me!

src/docker-in-docker/install.sh
```diff
@@ -307,13 +305,27 @@ if [ -f "/usr/local/share/docker-init.sh" ]; then
...
-        groupadd docker
-    fi
+if ! cat /etc/group | grep -e "^docker:" > /dev/null 2>&1; then
+        groupadd -r docker
+fi
+
+usermod -aG docker ${USERNAME}
+
+if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
+    buildx_version="latest"
```
joshspicer, line 315: cc/ @Chuxel if he has any licensing concerns? I know we've had similar concerns in the past with docker.

joshspicer (approved):

PR #481, opened 2023-03-09
Remove duplicate env var sourcing for codespaces

eljog:
Codespaces already has the same mechanism built-in for sourcing secrets for login shells, from env-secrets file.
Removing the redundant and codespaces specific code from sshd feature.

### DIFF
--- a/src/sshd/install.sh
+++ b/src/sshd/install.sh
@@ -13,7 +13,6 @@ SSHD_PORT="${SSHD_PORT:-"2222"}"
 USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
 START_SSHD="${START_SSHD:-"false"}"
 NEW_PASSWORD="${NEW_PASSWORD:-"skip"}"
-FIX_ENVIRONMENT="${FIX_ENVIRONMENT:-"true"}"
 
 set -e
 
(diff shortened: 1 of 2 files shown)


dmgardiner25 (approved):

src/sshd/install.sh
```diff
@@ -93,35 +93,6 @@ sed -i -E "s/#*\s*Port\s+.+/Port ${SSHD_PORT}/g" /etc/ssh/sshd_config
 # Need to UsePAM so /etc/environment is processed
 sed -i -E "s/#?\s*UsePAM\s+.+/UsePAM yes/g" /etc/ssh/sshd_config
 
-# Script to store variables that exist at the time the ENTRYPOINT is fired
```
samruddhikhandale, line 96: Can you bump the patch version here? Thanks!  
https://github.com/devcontainers/features/blob/c9f622064b3c009b4101a7e9803148ff25d66ed3/src/sshd/devcontainer-feature.json#L3

src/sshd/install.sh
```diff
@@ -141,15 +112,6 @@ sudoIf()
 }
 
 EOF
-if [ "${FIX_ENVIRONMENT}" = "true" ]; then
```
samruddhikhandale, line 144: Can we remove this variable declaration at the top?

> FIX_ENVIRONMENT="${FIX_ENVIRONMENT:-"true"}"

eljog: done

src/sshd/install.sh
eljog: done

PR #487, opened 2023-03-11
Add pytest to Python Feature

dciborow:
The Python feature installs the Python Extension for VS Code, which enables testing functionality through `pytest`, but this package is not installed for the user.

### DIFF
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -26,7 +26,7 @@ CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN="${CONFIGUREJUPYTERLABALLOWORIGIN:-""}"
 # alongside PYTHON_VERSION, but not set as default.
 ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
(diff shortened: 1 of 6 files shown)


dciborow:
maybe this is not a good idea... when using the devcontainer python image, pytest needs to be installed in /usr/local/bin/python.

src/python/install.sh
```diff
@@ -26,7 +26,7 @@ CONFIGURE_JUPYTERLAB_ALLOW_ORIGIN="${CONFIGUREJUPYTERLABALLOWORIGIN:-""}"
 # alongside PYTHON_VERSION, but not set as default.
 ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
-DEFAULT_UTILS=("pylint" "flake8" "autopep8" "black" "yapf" "mypy" "pydocstyle" "pycodestyle" "bandit" "pipenv" "virtualenv")
```
samruddhikhandale, line 29: Can you bump the patch version in here 👇 ? Thanks
https://github.com/devcontainers/features/blob/3f2738f0217e9f6e5dc441fa105f2dca7f1e00d0/src/python/devcontainer-feature.json#L3

dciborow: Done!

samruddhikhandale (approved):
Nice, thanks! 🎉

PR #500, opened 2023-03-23
Add install support for Git and Rust feature

abergmeier:
Also add tests for Fedora 38


abergmeier:
Can currently not test this since devcontainer cli is not working on Ubuntu 22.04.

abergmeier:

PR #511, opened 2023-04-03
add a postCreateCommand to `git-lfs` Feature

joshspicer:
The postCreateCommand will trigger `git lfs pull` if lfs artifacts are detected in the repository.  

This is particularly useful for GitHub Codespaces, where the repo clone is performed before installing git-lfs.

### DIFF
--- a/src/git-lfs/devcontainer-feature.json
+++ b/src/git-lfs/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "git-lfs",
-    "version": "1.0.7",
+    "version": "1.0.8",
     "name": "Git Large File Support (LFS)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/git-lfs",
     "description": "Installs Git Large File Support (Git LFS) along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like git and curl.",
@@ -13,8 +13,14 @@
             ],
             "default": "latest",
             "description": "Select version of Git LFS to install"
+        },
+        "autoPull": {
+            "type": "boolean",
+            "default": true,
+            "description": "Automatically pull LFS files when creating the container. If false, manually running 'git lfs pull' will pull lfs files."
         }
     },
+    "postCreateCommand": "/tmp/pull-git-lfs-artifacts.sh",
     "installsAfter": [
         "ghcr.io/devcontainers/features/common-utils"
     ]
--- a/src/git-lfs/devcontainer-feature.json
+++ b/src/git-lfs/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "git-lfs",
-    "version": "1.0.8",
+    "version": "1.1.0",
     "name": "Git Large File Support (LFS)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/git-lfs",
     "description": "Installs Git Large File Support (Git LFS) along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like git and curl.",
@@ -17,7 +17,7 @@
         "autoPull": {
             "type": "boolean",
             "default": true,
-            "description": "Automatically pull LFS files when creating the container. If false, manually running 'git lfs pull' will pull lfs files."
+            "description": "Automatically pull LFS files when creating the container.  When false, running 'git lfs pull' in the container will have the same effect."
         }
     },
     "postCreateCommand": "/tmp/pull-git-lfs-artifacts.sh",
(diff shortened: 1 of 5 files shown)


test/git-lfs/scenarios.json
```diff
@@ -0,0 +1,20 @@
+{
+    "autoPullEnabled": {
+        "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+        "initializeCommand": "(git clone https://github.com/codspace/git-lfs-example /tmp/testRepo || true) && (cp -r /tmp/testRepo/.git* . && cp -r /tmp/testRepo/* .)",
```
joshspicer, line 4: Should I move this to the `devcontainers` org?   Also open to other ideas, but at the end of the day the best way to test this is to have a live git repo cloned in with an LFS artifact.

src/git-lfs/devcontainer-feature.json
```diff
@@ -13,8 +13,14 @@
...
             "default": "latest",
             "description": "Select version of Git LFS to install"
+        },
+        "autoPull": {
+            "type": "boolean",
+            "default": true,
+            "description": "Automatically pull LFS files when creating the container. If false, manually running 'git lfs pull' will pull lfs files."
         }
     },
+    "postCreateCommand": "/tmp/pull-git-lfs-artifacts.sh",
```
samruddhikhandale, line 23: Would adding this be an unexpected change for the user? 

1. Should we bump `major` version instead or 
2. Should we update `minor` version and set `autoPull` to `false` by default?

src/git-lfs/install.sh
```diff
@@ -185,6 +186,36 @@ if [ "${use_github}" = "true" ]; then
...
+# Check if repo is a git lfs repo.
+if ! git lfs ls-files > /dev/null 2>&1; then
+    echo "(!) Skipping automatic 'git lfs pull' because no git lfs files were detected"
+    exit 0
+fi
+
+git lfs pull
+EOF
+
+chmod 755 /tmp/pull-git-lfs-artifacts.sh
```
samruddhikhandale, line 217: Question, should this be `chowned` for the user (assuming it's not `root`)

samruddhikhandale (commented):
Exciting!!! Left some comments

test/git-lfs/scenarios.json
samruddhikhandale: > Should I move this to the devcontainers org?

I like that, I wonder if we should create a new `test` repo for this purpose and add scenarios to different branches. This way we don't have to worry about unwanted dependency on other repos. We'd like to avoid test failures just because `codspace/git-lfs-example` was deleted at some point.

src/git-lfs/devcontainer-feature.json
joshspicer: Good point, I shouldn't have bumped the breakfix version.  I think the minor version.  From what I understand, it's expected behavior to auto pull (if you have git lfs already installed on your computer and do a `git clone`, you'd end up with the artifacts auto pulled).    This manifests differently in Codespaces since we do the clone first _before_ installing git-lfs, so there's never a change for that pull to happen. That's why I set it the default to true

src/git-lfs/install.sh
joshspicer: good point, i'll add that and a test with a non-root remote user

joshspicer: well actually, these permissions let anyone read or execute so chowning is unnecessary.  I'll still add a remoteUser to validate that.

src/git-lfs/devcontainer-feature.json
joshspicer: aside, this is why in the `scenarios.json` I need `GIT_LFS_SKIP_SMUDGE=1` in front of the git clone.  GitHub Actions must have git-lfs installed, so in CI it was automatically doing the pull when I cloned, which of course we don't want for this test :)

src/git-lfs/install.sh
```diff
@@ -185,6 +186,36 @@ if [ "${use_github}" = "true" ]; then
     install_using_github
 fi
 
+# --- Generate a 'pull-git-lfs-artifacts.sh' script to be executed by the 'postCreateCommand' lifecycle hook
+
+tee /tmp/pull-git-lfs-artifacts.sh > /dev/null \
```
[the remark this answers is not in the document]
joshspicer: Ah yes i meant to have changed this!  Unless others have opinion, i'm going to do /usr/local/share/git-lfs-artifacts.sh, mirroring where we place sshd and docker-in-docker scripts.

samruddhikhandale (approved):
Looks great ⚡

PR #515, opened 2023-04-05
Removing forced gid 999 for docker

bigtallcampbell:
The group GID of 999 is a reserved ID on Azure VMs.  This PR removes the forced check for 999 and allows the rest of the feature to finish activating.

Log of file before this change:
```bash
[2023-04-05T18:53:18.249Z] 
#0 7.617 (*) Installing compose-switch as docker-compose...
#0 7.782 compose_switch_version=1.0.5
#0 7.949 (*) Updating docker group gid to 999...
#0 7.950 groupmod: GID '999' already exists
#0 7.951 ERROR: Feature "Docker (docker-outside-of-docker)" (ghcr.io/devcontainers/features/docker-outside-of-docker) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker for help troubleshooting this error.
------
[2023-04-05T18:53:18.249Z] failed to solve: process "/bin/sh -c chmod -R 0700 /tmp/build-features/docker-outside-of-docker_1 && cd /tmp/build-features/docker-outside-of-docker_1 && chmod +x ./devcontainer-features-install.sh && ./devcontainer-features-install.sh" did not complete successfully: exit code: 4

```

### DIFF
--- a/src/docker-outside-of-docker/install.sh
+++ b/src/docker-outside-of-docker/install.sh
@@ -261,11 +261,13 @@ if ! grep -qE '^docker:' /etc/group; then
     groupadd --system docker
 fi
 
+# Remarking this out to restore functionality in Azure VMs.  ID 999 is a reserved group ID
 # Ensure docker group gid is 999
-if [ "$(getent group docker | cut -d: -f3)" != "999" ]; then
-    echo "(*) Updating docker group gid to 999..."
-    groupmod -g 999 docker
-fi
+# if [ "$(getent group docker | cut -d: -f3)" != "999" ]; then
+#     echo "(*) Updating docker group gid to 999..."
+#     groupmod -g 999 docker
+# fi
+
 
 usermod -aG docker "${USERNAME}"
 


jkeech:
cc @felipecrs @samruddhikhandale Looks related to #443 that just merged.

src/docker-outside-of-docker/install.sh
```diff
@@ -261,11 +261,13 @@ if ! grep -qE '^docker:' /etc/group; then
     groupadd --system docker
 fi
 
+# Remarking this out to restore functionality in Azure VMs.  ID 999 is a reserved group ID
 # Ensure docker group gid is 999
-if [ "$(getent group docker | cut -d: -f3)" != "999" ]; then
-    echo "(*) Updating docker group gid to 999..."
-    groupmod -g 999 docker
-fi
+# if [ "$(getent group docker | cut -d: -f3)" != "999" ]; then
```
samruddhikhandale, line 266: Can you bump the patch version? Thanks! 

https://github.com/devcontainers/features/blob/4a92715458af1d23bdb6f24098e00af38ad77dd8/src/docker-outside-of-docker/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
> As part of the tests, I figured out an issue when using moby. It seems that the group which moby creates hasn't the gid of 999 which is the default docker gid. I added a fix for such cases as well.

The commented out code (in this PR) was added in https://github.com/devcontainers/features/pull/443 which was merged an hour ago.
The issue with moby existed even before it, in order to avoid failures for folks, we can comment this bit for now.

I'll create a followup PR to fix that edge case.

felipecrs:
I believe it is ok to revert this piece of code, but have in mind that the docker_init_moby tests will start to fail.

However, I don't know if that failure indicates a real world problem or not.

samruddhikhandale:
> I believe it is ok to revert this piece of code, but have in mind that the docker_init_moby tests will start to fail.

Strange, there were no test failures - all checks seem to have passed 🤔

src/docker-outside-of-docker/install.sh
samruddhikhandale: Pushed that change!

samruddhikhandale (approved):

samruddhikhandale:
Thanks for the fix! ❇️

PR #519, opened 2023-04-07
Speeds up Node Feature: Remove unnecessary "chmod" command

samruddhikhandale:
`chmod`ing the `$NVM_DIR/versions` folder takes a lot of time depending upon the size of this folder.
We already change the permissions of `NVM_DIR` in [here](https://github.com/devcontainers/features/blob/main/src/node/install.sh#L154). Hence, removing duplicates! 

Tested with the following scenario 👇 

```jsonc
"image": "mcr.microsoft.com/devcontainers/universal",
        "remoteUser": "codespace",
        "features": {
            "node": {
                "version": "16"
            }
        }
```

Left : Before the PR changes - Node Feature took 138.4s
Right: Post the PR changes - Node Feature took 5.6s 

Removing the duplicate `chmod` did not make a difference to the permissions of the `version` folder.

image

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.2.0",
+    "version": "1.2.1",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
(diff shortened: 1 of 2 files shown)


jkeech:
Can you double check that this is not needed? It looks like the `chmod` [here](https://github.com/devcontainers/features/blob/main/src/node/install.sh#L154) is not recursive, and it looks like the recursive `chmod` was added intentionally to fix a different issue in the past (although perhaps tests weren't written to cover that scenario). cc @Chuxel who helped make some of the earlier changes.

samruddhikhandale:
> The chmod here is needed for scenarios where the UID/GID is sync'd. Removing it will cause issues. IIRC, Actions is one of the scenarios where this happens.
> 
> Originally the way nvm untar'd contents, the sticky bit worked, but that stopped working so a full chmod was the only recourse AFAIK.

Makes sense, I'll close this PR. I'll look into another ways (if possible) to speed up the Feature.

PR #532, opened 2023-04-28
Adds httpProxy option to Python Feature

### LINKED ISSUE #530: Python: GPG ignores http_proxy environment variable
Hello, I'm trying to use this feature behind a corporate proxy. I can set the **http_proxy** and **https_proxy** environment variables in a custom dockerfile before the features are installed, as well as the apt proxy configs, but sadly gpg ignores those variables by default it seems, and its used for verifying the python source code in the feature.

Conditionally adding `http_proxy="${http_proxy}"` to the `--keyserver-args` fixes the issue.

I could raise a PR for this as I've got it working locally but am unsure how I'd test something like this without spinning up some fake proxy server and seeing if it gets used.

tweelix:
This option allows GPG to use a HTTP proxy to fetch keys from remote keyservers
fixes #530
As discussed in the issue this only sets the proxy variable for GPG rather than the full feature script

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.0.23",
+  "version": "1.1.0",
   "name": "Python",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/python",
   "description": "Installs the provided version of Python, as well as PIPX, and other common Python utilities.  JupyterLab is conditionally installed with the python feature. Note: May require source code compilation.",
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.0.22",
+  "version": "1.0.23",
   "name": "Python",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/python",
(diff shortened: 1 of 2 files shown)


tweelix:

src/python/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.0.22",
+  "version": "1.0.23",
```
samruddhikhandale, line 3:
```suggestion
  "version": "1.1.0",
```

As this PR adds a new Feature option, can we bump the minor version? thanks!

src/python/install.sh
```diff
@@ -84,6 +86,9 @@ receive_gpg_keys() {
         mkdir -p "$(dirname \"$2\")"
         keyring_args="--no-default-keyring --keyring $2"
     fi
+    if [ ! -z "${KEYSERVER_PROXY}" ]; then
+	    keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
```
samruddhikhandale, line 90: Can we format this and remove one tab to match the previous block?

samruddhikhandale (commented):
Thank you, left minor comments.

src/python/install.sh
tweelix: Good spot, interesting quirk of the github web editor there

samruddhikhandale (approved):
👏 👏

Facingreality (approved):
src/python/devcontainer-feature.json

PR #533, opened 2023-04-28
Add support for skipping `docker-compose` v1 operations during Docker feature install script

m-roberts:
Currently if someone wants to use Docker-in-Docker or Docker-outside-of-Docker features on arm64, `pipx` is invoked to install `docker-compose` v1, even if it is not wanted. This takes time and uses the image's Python configuration more than is desirable for my use-case (I am using a virtualenv, and `--user` does not play nicely).

To avoid these steps, I am currently using this workaround in my Dockerfile to pass the `type docker-compose` part of `install.sh` so that the aforementioned steps are not invoked:

```shell
RUN touch /usr/local/bin/docker-compose && chmod +x /usr/local/bin/docker-compose
```

This PR adds a "none" option for `dockerDashComposeVersion` to simply ignore the need for this binary.

Another possible way of doing this would be to only install v1 if it is selected.

### DIFF
--- a/test/docker-outside-of-docker/docker_build_no_compose.sh
+++ b/test/docker-outside-of-docker/docker_build_no_compose.sh
@@ -10,7 +10,6 @@ check "docker-buildx" docker buildx version
 check "docker-build" docker build ./
 
 check "not installing compose skips docker-compose v1 install" bash -c "! type docker-compose"
-check "not installing compose skips compose-switch" bash -c "[[ ! -f /usr/local/bin/compose-switch ]]"
 
 # Report result
 reportResults
--- a/test/docker-outside-of-docker/docker_dash_compose_v2.sh
+++ b/test/docker-outside-of-docker/docker_dash_compose_v2.sh
@@ -9,7 +9,7 @@ source dev-container-features-test-lib
 check "docker compose" bash -c "docker compose version | grep -E '2.[0-9]+.[0-9]+'"
 check "docker-compose" bash -c "docker-compose --version | grep -E '2.[0-9]+.[0-9]+'"
 
-check "installs compose-switch" bash -c "[[ -f /usr/local/bin/compose-switch ]]"
+check "installs compose-switch as docker-compose" bash -c "[[ -f /usr/local/bin/docker-compose ]]"
 
 # Report result
 reportResults


src/docker-in-docker/README.md
```diff
@@ -17,7 +17,7 @@ Create child containers *inside* a container, independent from the host's docker
 |-----|-----|-----|-----|
 | version | Select or enter a Docker/Moby Engine version. (Availability can vary by OS version.) | string | latest |
 | moby | Install OSS Moby build instead of Docker CE | boolean | true |
-| dockerDashComposeVersion | Default version of Docker Compose (v1 or v2) | string | v1 |
+| dockerDashComposeVersion | Default version of Docker Compose (v1 or v2 or none) | string | v1 |
```
samruddhikhandale, line 20: README files are autogenerated with https://github.com/devcontainers/action.

Can we revert the change to this file, and update 👇  instead? https://github.com/devcontainers/features/blob/89ac66a01556a2133da4cfd6a0608e80c0e42d2c/src/docker-in-docker/devcontainer-feature.json#L31

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -23,6 +23,7 @@
         "dockerDashComposeVersion": {
             "type": "string",
             "enum": [
+                "none",
```
samruddhikhandale, line 26: As we are allowing a new value for this Feature option, can we update the minor version in here?
https://github.com/devcontainers/features/blob/89ac66a01556a2133da4cfd6a0608e80c0e42d2c/src/docker-in-docker/devcontainer-feature.json#L3

src/docker-in-docker/install.sh
```diff
@@ -242,58 +242,61 @@ fi
...
-        export PIP_CACHE_DIR=/tmp/pip-tmp/cache
-        pipx_bin=pipx
-        if ! type pipx > /dev/null 2>&1; then
-            pip3 install --disable-pip-version-check --no-cache-dir --user pipx
-            pipx_bin=/tmp/pip-tmp/bin/pipx
-        fi
-        ${pipx_bin} install --pip-args '--no-cache-dir --force-reinstall' docker-compose
-        rm -rf /tmp/pip-tmp
+# If 'docker-compose' command is to be included
+if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "none" ]; then
```
samruddhikhandale, line 246: Can you help add a test to validate this change? thanks! 

Reference docs: https://github.com/devcontainers/cli/blob/main/docs/features/test.md

src/docker-outside-of-docker/README.md
```diff
@@ -19,7 +19,7 @@ Re-use the host docker socket, adding the Docker CLI to a container. Feature inv
 |-----|-----|-----|-----|
 | version | Select or enter a Docker/Moby CLI version. (Availability can vary by OS version.) | string | latest |
 | moby | Install OSS Moby build instead of Docker CE | boolean | true |
-| dockerDashComposeVersion | Compose version to use for docker-compose (v1 or v2) | string | v2 |
+| dockerDashComposeVersion | Compose version to use for docker-compose (v1 or v2 or none) | string | v2 |
```
samruddhikhandale, line 22: Can you address all the comments added for the `docker-in-docker` Feature for `docker-outside-of-docker` Feature as well?

samruddhikhandale (changes requested):
Thank you for opening this PR, left some comments.

src/docker-in-docker/install.sh
samruddhikhandale: Looks good to me, thanks.
If you could also add similar tests to `docker-outside-of-docker`, it would be great!

samruddhikhandale (approved):
Looks good, thanks!
left a follow-up comment.

samruddhikhandale:
@m-roberts Appreciate all the test coverage, looks like there are some test failures https://github.com/devcontainers/features/actions/runs/4859662615. Can you help take a look at it?

Issue #531, opened 2023-04-26
groupadd: GID '1000' already exists


raymond-chetty:
It looks like it is an error to set the GID to 1000 when installing the common-utils feature on a debian base in codespaces.
It's possible this is a codespaces issue... or maybe the base image includes a common-utils feature which creates a 1000 group. Once created the id isn't overridden in the codespaces devcontainer.json file that includes the feature. 

https://github.com/devcontainers/images/blob/main/src/base-debian/.devcontainer/devcontainer.json#L7
https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#L362

I created the devcontainer.json below using the vscode wizard in github codespaces....
F1 -> Add Dev Container Configuration Files... -> Create new configuration -> Debian -> bullseye -> Common utilities -> meson -> bazel -> android sdk -> android sdk platform tools -> ok -> configure options -> emulator -> ok -> 33 -> 33.0.2 -> v1.15.0 -> nonFreePackages -> ok -> devcontainer -> 1000 -> 1000

```
2023-04-26 20:08:02.133Z: #17 0.515 ===========================================================================
#17 0.515 Feature       : Common Utilities
#17 0.515 Description   : Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.
#17 0.515 Id            : ghcr.io/devcontainers/features/common-utils
#17 0.515 Version       : 2.0.8
#17 0.515 Documentation : https://github.com/devcontainers/features/tree/main/src/common-utils
#17 0.515 Options       :
#17 0.515     INSTALLZSH="true"
#17 0.515     CONFIGUREZSHASDEFAULTSHELL="false"
#17 0.515     INSTALLOHMYZSH="true"
#17 0.515     UPGRADEPACKAGES="true"
#17 0.515     USERNAME="devcontainer"
#17 0.515     USERUID="1000"
#17 0.515     USERGID="1000"
#17 0.515     NONFREEPACKAGES="true"
2023-04-26 20:08:02.241Z: #17 0.522 ===========================================================================
#17 0.587 Marker file found:
#17 0.590     PACKAGES_ALREADY_INSTALLED=true
#17 0.590     LOCALE_ALREADY_SET=true
#17 0.590     EXISTING_NON_ROOT_USER=vscode
#17 0.590     RC_SNIPPET_ALREADY_ADDED=true
#17 0.590     ZSH_ALREADY_INSTALLED=true
#17 0.623 
[...]

PR #534, opened 2023-04-28
Update NOTES.md

raymond-chetty:
#531: Quick document update about UID & GID using 1000 in devcontainer base images.

### DIFF
--- a/src/common-utils/NOTES.md
+++ b/src/common-utils/NOTES.md
@@ -2,7 +2,7 @@
 
 This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, RockyLinux, and Alpine Linux.
 
-## Base Images
+## Using with dev container images
 
 This Feature is used in many of the [dev container images](https://github.com/search?q=repo%3Adevcontainers%2Fimages+%22ghcr.io%2Fdevcontainers%2Ffeatures%2Fcommon-utils%22&type=code), as a result
 these images have already allocated UID & GID 1000. Attempting to add this Feature with  UID 1000 and/or GID 1000 on top of such a dev container image will result in an error when building the dev container.
--- a/src/common-utils/NOTES.md
+++ b/src/common-utils/NOTES.md
@@ -2,6 +2,12 @@
 
 This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, RockyLinux, and Alpine Linux.
 
+## Base Images
+
+This feature is used in many of the devcontainer [base images](https://github.com/devcontainers/images), as a result
+those base images have already allocated UID & GID 1000. Attempting to add this feature on a new configuration utilizing
+a devcontainer base image may result in an error when building if adding a new user and assiging UID or GID to 1000. 
+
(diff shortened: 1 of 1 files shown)


raymond-chetty:
Its possible to resolve this if we add a configuration to usermod a username as well as UID & GID. Possibly open a ticket for that.

src/common-utils/NOTES.md
```diff
@@ -2,6 +2,12 @@
 
 This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, RockyLinux, and Alpine Linux.
 
+## Base Images
```
samruddhikhandale, line 5:
```suggestion
## Using with dev container images
```

samruddhikhandale (changes requested):
Appreciate your help in improving our documentation, thank you! 

Left some minor comments.

samruddhikhandale:
> Looking at how the vscode wizard works, you can't manually type in a UID value only select from an option,

As mentioned in https://github.com/devcontainers/features/issues/531#issuecomment-1524168766,  I believe that's a bug with the wizard. Feel free to open an issue in https://github.com/microsoft/vscode-remote-release/issues

> Looking at how the vscode wizard works, you can't manually type in a UID value only select from an option, it might be better to assume that if UID 1000 exists it belongs to the base image user and that setting changes should all be applied to UID 1000, possibly including UID changes.

> Logic:
> if UID 1000 exists -> mod the user
> if UID X exists -> mod the user (no UID changes)
> else add the user
> 
> This way wizard users will do what they'd expect, modify the base user, assuming we leave the UID & GID 1000 option. Also, manual configurations will be possible with a variety of scenarios either adding or updating a user.

Thanks for sharing your thoughts, it's really helpful. 

We could make this change to the common-utils Feature, however, I am not sure if the logic could become confusing for users ; instead I think a clear error message could navigate them to use different UID/GID to unblock them.
I am happy to help out with the change if people think otherwise.

src/common-utils/NOTES.md
```diff
@@ -2,6 +2,12 @@
...
+
+This feature is used in many of the devcontainer [base images](https://github.com/devcontainers/images), as a result
+those base images have already allocated UID & GID 1000. Attempting to add this feature on a new configuration utilizing
+a devcontainer base image may result in an error when building if adding a new user and assiging UID or GID to 1000. 
+
```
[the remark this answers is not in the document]
raymond-chetty: Thank you for your suggestion. I approve of this correction on behalf of the people of the united states. @PresidentObama :)

src/common-utils/NOTES.md
raymond-chetty: approved!

samruddhikhandale: Same here!

nanis94:
devcontainers:main

nanis94:
devcontainers:main

nanis94:
devcontainers:main

nanis94 (commented):
devcontainers:main

samruddhikhandale (approved):
Thanks!

PR #535, opened 2023-04-28
Update devcontainer-feature.json

raymond-chetty:
#531 Modify proposed UID & GID so as not to conflict with base image.

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -40,7 +40,7 @@
         "userUid": {
             "type": "string",
             "proposals": [
-                "1000",
+                "1001",
                 "automatic"
             ],
             "default": "automatic",
@@ -49,7 +49,7 @@
         "userGid": {
             "type": "string",
             "proposals": [
-                "1000",
+                "1001",
                 "automatic"
             ],
             "default": "automatic",


src/common-utils/devcontainer-feature.json
```diff
@@ -40,7 +40,7 @@
         "userUid": {
             "type": "string",
             "proposals": [
-                "1000",
+                "1001",
```
samruddhikhandale, line 43: Can we update the patch version? Thanks! 

https://github.com/devcontainers/features/blob/4c7261225d5674773cf7f7bdebacf8dffbaf1a13/src/common-utils/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thank you. ✨ 

Left one minor comment.

src/common-utils/devcontainer-feature.json
raymond-chetty: I created a pull request #539.

raymond-chetty: It might be worth adding a `modifyUsername` parameter so that users can update the username instead of assume the intention is to create a new user when the username doesn't match the username of the base image.

samruddhikhandale (approved):

src/common-utils/devcontainer-feature.json
raymond-chetty: maybe it would be better to have added 1001 instead of modify the option... If you want I can add the original back as another option, an addition instead of substitution, so that the full list of options demonstrate the available capabilities of the feature.

PR #547, opened 2023-05-08
Create .config directory in common-utils

### LINKED ISSUE #529: common-utils: Add ~/.config directory for user
It can be common to mount (usually, subdirectories of) the user `~/.config` directory, either to persist the configs or to share them between containers.  Unfortunately, if the directory doesn't exist inside the container, mounting it causes it to be owned by root.  This interferes with mundane things like writing the run-once notice at the end of the script, but can cause additional subtler effects.

I'd be willing to make the changes for this, just wanting feedback, especially around whether there are any distro-specific issues here.

Clockwork-Muse:
Fixes #529

### DIFF
--- a/test/common-utils/config-subdirectory.sh
+++ b/test/common-utils/config-subdirectory.sh
@@ -6,8 +6,8 @@ set -e
 source dev-container-features-test-lib
 
 # Definition specific tests
-check "owned config directory" ls -ld ~/.config/subdirectory | awk '{print $3}' | grep 'devcontainer'
-check "owned config directory" ls -ld ~/.config | awk '{print $3}' | grep 'devcontainer'
+check "owned-config-sub-directory" bash -c "ls -ld ~/.config/subdirectory | awk '{print $3}' | grep 'devcontainer'"
+check "owned-config-directory" bash -c "ls -ld ~/.config | awk '{print $3}' | grep 'devcontainer'"
 
 # Report result
 reportResults
--- a/test/common-utils/config-subdirectory.sh
+++ b/test/common-utils/config-subdirectory.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+# check "owned config directory" bash -c "ls -ld .vscode | awk '{print $3}' | grep 'devcontainer'"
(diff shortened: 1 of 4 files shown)


test/common-utils/scenarios.json
```diff
@@ -115,6 +115,16 @@
             }
         }
     },
+    "config-subdirectory": {
```
samruddhikhandale, line 118: Looks like there is a test failure for this scenario, can you help fix it?
https://github.com/devcontainers/features/actions/runs/4910062272/jobs/8781347850?pr=547#step:4:14678

samruddhikhandale (changes requested):
Looks great, thank you! 
Left some minor comments.

test/common-utils/scenarios.json
Clockwork-Muse: Attempted to fix it, but I'm not sure what other directories there would be for me to mount.

test/common-utils/config-subdirectory.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+check "owned config directory" ls -ld ~/.config/subdirectory | awk '{print $3}' | grep 'devcontainer'
+check "owned config directory" ls -ld ~/.config | awk '{print $3}' | grep 'devcontainer'
```
samruddhikhandale, line 10:
```suggestion
check "owned-config-sub-directory" bash -c "ls -ld ~/.config/subdirectory | awk '{print $3}' | grep 'devcontainer'"
check "owned-config-directory" bash -c "ls -ld ~/.config | awk '{print $3}' | grep 'devcontainer'"
```

The tests should run if you pass it within `bash -c`

Clockwork-Muse: I've updated as suggested, however I'm not sure why it should make a difference, given that it runs fine locally (and I'm assuming the script runs _inside_ the container), and there are other scripts that do not use `bash -c`

samruddhikhandale: When I tested it locally, it did not run checks for me. Also, we need to wrap within `bash -c` as these commands are piped, looks like `check` function doesn't handle that well.

samruddhikhandale: The tests pass! 🎉

samruddhikhandale (approved):
Looks great, thank you so much!

PR #557, opened 2023-05-23
Fix: common-utils configureZshAsDefaultShell option not working on alpine linux image

krikchaip:
Hi, I was trying to set up a [devcontainer for my side-project](https://github.com/krikchaip/docker-kubernetes-course/blob/main/.devcontainer/devcontainer.json) using the base alpine image with `common-utils` feature. I knew that the image already included the feature, but I also wanted to set the default shell to `zsh`. According to [the documentation](https://github.com/devcontainers/features/tree/main/src/common-utils), I had to pass `"configureZshAsDefaultShell": "true"`, but it didn't seem to be working for me.

While reading error logs, I found that the problem came from `chsh` command, which was always asking for a password.

https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#LL439C9-L441C7
```bash
if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
    chsh --shell /bin/zsh ${USERNAME}

    # PASSWORD:
    # error -> chsh: PAM: Authentication failure
fi
```

After doing some research, I found [the solution](https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure). Basically, You need to modify the content in some file as shown on my PR.

`/etc/pam.d/chsh`
```bash
auth sufficient pam_shells.so
```

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -439,7 +439,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
         # Fixing chsh always asking for a password on alpine linux
         # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
-        if [ -f "/etc/pam.d/chsh" ] && grep -Eq "^auth(.*)pam_rootok\.so$" /etc/pam.d/chsh; then
+        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$"; then
             awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
         else
             echo "auth sufficient pam_rootok.so" > /etc/pam.d/chsh
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -439,7 +439,11 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
         # Fixing chsh always asking for a password on alpine linux
         # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
-        echo "auth sufficient pam_shells.so" > /etc/pam.d/chsh
+        if [ -f "/etc/pam.d/chsh" ] && grep -Eq "^auth(.*)pam_rootok\.so$" /etc/pam.d/chsh; then
+            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
+        else
+            echo "auth sufficient pam_rootok.so" > /etc/pam.d/chsh
+        fi
 
         chsh --shell /bin/zsh ${USERNAME}
     fi
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -439,7 +439,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
         # Fixing chsh always asking for a password on alpine linux
         # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
-        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$"; then
+        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$" /etc/pam.d/chsh; then
             awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
         else
             echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -439,10 +439,10 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
         # Fixing chsh always asking for a password on alpine linux
         # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
-        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$" /etc/pam.d/chsh; then
-            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
-        else
+        if [ ! -f "/etc/pam.d/chsh" ] || ! grep -Eq '^auth(.*)pam_rootok\.so$' /etc/pam.d/chsh; then
             echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh
+        elif [[ -n "$(awk '/^auth(.*)pam_rootok\.so$/ && !/^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$/' /etc/pam.d/chsh)" ]]; then
+            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
         fi
 
         chsh --shell /bin/zsh ${USERNAME}
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -442,7 +442,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
         if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$"; then
             awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
         else
-            echo "auth sufficient pam_rootok.so" > /etc/pam.d/chsh
+            echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh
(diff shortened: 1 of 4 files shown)


krikchaip:

src/common-utils/main.sh
```diff
@@ -437,6 +437,10 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     fi
 
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
+        # Fixing chsh always asking for a password on alpine linux
+        # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
+        echo "auth sufficient pam_shells.so" > /etc/pam.d/chsh
```
samruddhikhandale, line 442: I am not confident if changing auth permissions (to be able to switch shell) is a completely secured way. Wondering if there's another alternative.

@chrmarti @joshspicer would appreciate your thoughts here.

samruddhikhandale: If we end up adding this setting, then we should remove it after `chsh` command.

samruddhikhandale, line 442: Also, if the user already has `auth required pam_shells.so` in `/etc/pam.d/chsh`, then shouldn't that be replaced with `auth sufficient pam_shells` for `chsh` to work?

samruddhikhandale (commented):
Thanks for the PR, left some thoughts!

> The license/cla status has been queued for over a day. How can I make it green? 🤔

Mmm, I expect that it should pass now as the CLA is signed. I don't see an option to retry  the check, should have been automatic. 

Looking at https://github.com/microsoft/ContributorLicenseAgreement/issues/123 , probably closing & reopening the PR or pushing a new commit helps.

src/common-utils/main.sh
chrmarti: I see `auth            sufficient      pam_rootok.so` in `ubuntu:latest`. That only allows root to chsh (also for other users) without auth and should be safe to leave in.

krikchaip: Yes, that's correct. If that's the case then we can change it back after `chsh` command as you said. However, there are some considerations

* according to @chrmarti, for the `ubuntu` image, this file already existed with some content inside that he mentioned.
* for `alpine`, this file doesn't exist by default (from my observation so far). Thus, we need to create one with `auth sufficient ...` in order to make the `configureZshAsDefaultShell` work.

Not sure how we're gonna deal with this discrepancy though. Would appreciate your thoughts here.

chrmarti: You could check what the behavior is when both lines are present and whether the order matters. Maybe the last entry wins or maybe duplicate entries result in an error.

samruddhikhandale: Hi @krikchaip, I believe this PR is still needed? 
Can you help address this comment, then we could help review.  Let us know if you need help, thanks!

krikchaip: Yup. Sorry just got busy lately. I'm about to check what @chrmarti suggested. Will get back to you soon.

krikchaip: @samruddhikhandale how about checking whether if `/etc/pam.d/chsh` exists? If that's the case then we look for a line with `auth (...) pam_shells.so` and replace `(...)` with `sufficient`. What do you guys think?

samruddhikhandale: Sounds good to me, and if `auth (...) pam_shells.so ` does not exist? then, append `auth sufficient pam_shells.so` to end.

krikchaip:
I found something interesting inside `/etc/pam.d/chsh` of the devcontainer of this repo.

image

Seems like we should add `auth sufficient pam_rootok.so` instead of `pam_shells.so` 🤔 .

src/common-utils/main.sh
```diff
@@ -437,6 +437,14 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     fi
 
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
+        # Fixing chsh always asking for a password on alpine linux
+        # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
+        if [ -f "/etc/pam.d/chsh" ] && grep -Eq "^auth(.*)pam_rootok\.so$" /etc/pam.d/chsh; then
+            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
```
samruddhikhandale, line 443: Suggestion: Can add another condition so that we could avoid running this command when `/etc/pam.d/chsh` already contains `auth sufficient pam_rootok.so` ?


Something like 👇 

```

if [ -f "/etc/pam.d/chsh" ]; then
    if grep -vEq "^auth sufficient pam_rootok.so$"; then
            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
   fi
else
...

```

src/common-utils/main.sh
```diff
@@ -437,6 +437,14 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     fi
 
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
+        # Fixing chsh always asking for a password on alpine linux
+        # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
+        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$"; then
+            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
+        else
+            echo "auth sufficient pam_rootok.so" > /etc/pam.d/chsh
```
samruddhikhandale, line 445:
```suggestion
            echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh
```

src/common-utils/main.sh
samruddhikhandale: Looks like there's a bug, I tested with ubuntu where the `/etc/pam.d/chsh` file exists. It is still trying to add `auth sufficient pam_rootok.so`. Also, in this case, it is replacing the contents - which doesn't look correct

image

samruddhikhandale (changes requested):

krikchaip:
@samruddhikhandale hmmm. so weird. I've also tried the command with `/etc/pam.d/chsh` that has some content inside similar to you. It seemed to work pretty well AFAIK. 🤔


https://github.com/devcontainers/features/assets/23186625/f81c86ce-dc44-4bdf-8302-f6cd2692865b

samruddhikhandale:
@krikchaip In your case, you are running the conditions from the terminal, however, I tested it with a Feature test scenario (which resembles an actual Feature installation). Also, the install script is run with `user:root`, and in your case,  it's `node`. I believe these are the reasons why we both are seeing different results.

src/common-utils/main.sh
samruddhikhandale: With https://github.com/devcontainers/features/pull/557/commits/0ae570c72222c749911dda57af6962b8bca86187, instead of replacing contents, it is now appending to the end

image

src/common-utils/main.sh
```diff
@@ -437,6 +437,14 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
     if [ "${CONFIGURE_ZSH_AS_DEFAULT_SHELL}" == "true" ]; then
+        # Fixing chsh always asking for a password on alpine linux
+        # ref: https://askubuntu.com/questions/812420/chsh-always-asking-a-password-and-get-pam-authentication-failure.
+        if [ -f "/etc/pam.d/chsh" ] && grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$"; then
+            awk '/^auth(.*)pam_rootok\.so$/ { $2 = "sufficient" } { print }' /etc/pam.d/chsh > /tmp/chsh.tmp && mv /tmp/chsh.tmp /etc/pam.d/chsh
+        else
+            echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh
+        fi
+
         chsh --shell /bin/zsh ${USERNAME}
```
samruddhikhandale, line 448:
```suggestion
        CHSH_FILE_EXISTS="false"
        [ -f /etc/pam.d/chsh ] && CHSH_FILE_EXISTS="true"

        echo "auth sufficient pam_rootok.so" >> /etc/pam.d/chsh

        chsh --shell /bin/zsh ${USERNAME}

        if [ "${CHSH_FILE_EXISTS}" = "true" ]; then
            head -n -1 /etc/pam.d/chsh > /tmp/chsh-temp && mv /tmp/chsh-temp /etc/pam.d/chsh
        else
            rm -rf /etc/pam.d/chsh
        fi
```

@krikchaip what do you think of this approach? We are adding `auth sufficient pam_rootok.so`, switching the shell and reverting the changes. 

This way, we achieve our end goal of using `zsh` as default and don't have any lingering changes.

samruddhikhandale (changes requested):

krikchaip:
> in your case, you are running the conditions from the terminal, however, I tested it with a Feature test scenario (which resembles an actual Feature installation). Also, the install script is run with user:root, and in your case, it's node. I believe these are the reasons why we both are seeing different results.

@samruddhikhandale Did you mean the `configure_zsh_as_default_shell` scenario? If so, do you have any suggestion on how can I provide my own `/etc/pam.d/chsh` file?

src/common-utils/main.sh
krikchaip: You are basically trying to append the `auth sufficient pam_rootok.so` to the file right away. What will happen if there's already a line with `auth required pam_rootok.so`? Which one will take precedence? And the if statement at the bottom doesn't seem right to me.

src/common-utils/main.sh
krikchaip: I'm not sure how to reproduce this. Could you give me some guidance/instructions? Thx

samruddhikhandale:
@krikchaip Add echo statements to help debug the file changes to 👇 https://github.com/devcontainers/features/blob/0ae570c72222c749911dda57af6962b8bca86187/src/common-utils/main.sh#L439-L446 and run `BUILDKIT_PROGRESS=plain devcontainer features test -f common-utils --filter configure_zsh_as_default_shell --skip-autogenerated`

krikchaip:
@samruddhikhandale  Opsie.. my bad XD. Accidentally forgot that `grep` requires 2 parameters...

image

Please check again. It should be working rn.

image

samruddhikhandale:
> Can add another condition so that we could avoid running this command when /etc/pam.d/chsh already contains auth sufficient pam_rootok.so ?

Following up on https://github.com/devcontainers/features/pull/557#discussion_r1236009205

@krikchaip Can we update the logic to NOT make any changes to the file if `auth sufficient pam_rootok.so` exists? I don't think we should touch/make changes to a file if that's not needed. Saw this with `configure_zsh_as_default_shell` scenario, the diff is easily visible with respect to `spaces`

krikchaip:
@samruddhikhandale I believe the logic here seems to prevent it from touching the ` /etc/pam.d/chsh` file, but let me check again some time soon.

```bash
if [ -f "/etc/pam.d/chsh" ] &&
  grep -vEq "^auth[[:blank:]]+sufficient[[:blank:]]+pam_rootok\.so$" /etc/pam.d/chsh;
```

krikchaip:
@samruddhikhandale please recheck

samruddhikhandale (approved):
Looks great, thank you so much for taking the time to contribute this PR and reiterating on feedbacks. We appreciate it! ❇️

PR #561, opened 2023-05-25
chmod+x to all .sh files

AceHack:
I call this from Ubuntu, and I need them to already have the chmod +x attribute applied.


joshspicer:
Thanks for taking time to contribute! 

The dev container tooling will ensure that the execute bit is always set before running these scripts. While these Feature scripts may _mostly_ work if executed directly, it is not recommended to do so as they are designed specifically to work with the dev container CLI (or another supporting tool) and are subject to future changes diverging it from a 'simple shell script'.  For example, Specs [like this one](https://github.com/devcontainers/spec/blob/main/proposals/features-library.md), if implemented, will require the orchestrating CLI to source the requisite library scripts first.

May I ask what your use case is for running these scripts directly outside the tooling?  That may help me understand a gap in said tooling that we can improve.

PR #566, opened 2023-06-05
Change Ruby extension to ruby-lsp

rafaelfranca:
This is the extension that is recommended in the VSCode documentation now. 

https://code.visualstudio.com/docs/languages/ruby

### DIFF
--- a/src/ruby/devcontainer-feature.json
+++ b/src/ruby/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "ruby",
-    "version": "1.0.10",
+    "version": "1.1.0",
     "name": "Ruby (via rvm)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/ruby",
     "description": "Installs Ruby, rvm, rbenv, common Ruby utilities, and needed dependencies.",
@@ -21,7 +21,8 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "rebornix.Ruby"
+                "rebornix.Ruby",
+                "shopify.ruby-lsp"
             ]
         }
     },


src/ruby/devcontainer-feature.json
```diff
@@ -21,7 +21,7 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "rebornix.Ruby"
+                "Shopify.ruby-lsp"
```
samruddhikhandale, line 24:
```suggestion
                "rebornix.Ruby"
                "Shopify.ruby-lsp"
```

I wonder if we should keep "rebornix.Ruby" intact, but add "Shopify.ruby-lsp" as per the [VS Code docs recommendations](https://code.visualstudio.com/docs/languages/ruby). This way, we won't break scenarios for people who'd be expecting this extension.

@bamurtaugh any thoughts on this?

src/ruby/devcontainer-feature.json
```diff
@@ -21,7 +21,7 @@
     "customizations": {
```
samruddhikhandale, line 21: Can you help bump the minor version? thanks!

https://github.com/devcontainers/features/blob/0deca36b58024c8edd76d17c4df4db523b1ff01a/src/ruby/devcontainer-feature.json#L3

samruddhikhandale (commented):
Thank you, we appreciate your PR. 
Left some thoughts!

src/ruby/devcontainer-feature.json
rafaelfranca: That is a great question. [rebornix.Ruby isn't maintained anymore and it is recommending people to use `Shopify.ruby-lsp`](https://github.com/rubyide/vscode-ruby#deprecated).

I believe for most users they will not have any negative impact with this gem. Shopify.ruby-lsp tried to have feature parity with rebornix.Ruby since the first version.

Users that already have a container built, will just continue using rebornix.Ruby. Only new users that should get the new recommendation.

src/ruby/devcontainer-feature.json
rafaelfranca: Done

src/ruby/devcontainer-feature.json
bamurtaugh: @rebornix it'd be great to hear any insights you have on which Ruby extension we should use in our Ruby dev container Feature!

rafaelfranca: Any updates here? I want to avoid a situation where the Ruby community is left as second class citizen in the devcontainer ecosystem by getting as recommendation a extension that isn't maintained.

bamurtaugh: Hi folks! In chatting with @rebornix, we think it makes sense to include both extensions. 

We recommend Shopify’s Ruby extension in VS Code core, but Shopify’s extension doesn’t support Debugging, so we believe @rebornix's Ruby extension is still needed too.

samruddhikhandale (changes requested):

src/ruby/devcontainer-feature.json
samruddhikhandale, line 24:
```suggestion
                "shopify.ruby-lsp"
```

From 
image

samruddhikhandale: Probably, the `id` is case insensitive - but nice to be in sync.

rafaelfranca: > but Shopify’s extension doesn’t support Debugging

It does.

rafaelfranca: I added both just to move this forward, but I still believe recommending an unmaintained extension, when there is an alternative with the same feature set, will be configuring to users.

bamurtaugh: Thanks for clarifying! If it seems there aren't other gaps that'd be introduced by removing @rebornix's extension, it seems alright to remove it in this PR. I'd recommend adding a note in the Feature's readme and/or a comment in the devcontainer-feature.json to make users aware in case they are relying on the previous extension in some way.

rafaelfranca: I'm ok with waiting a few month to remove rebornix's extension as well so we don't cause any breaking change. Maybe get a larger set of users trying the Shopify's LSP and reporting gaps or missing features. Maybe aim to do it only in Q4 2023?

bamurtaugh: This sounds reasonable to me!

samruddhikhandale (approved):
Thank you! ✨

src/ruby/devcontainer-feature.json
Lobosque: Having both might confuse users (as it did to me). When you use the devcontainer feature, you get a popup asking you to disable `rebornix.Ruby` extension anyway, so why not remove it altogether?

Issue #560, opened 2023-05-24
dotnet feature fails to install on ubuntu:jammy


KIC:
using the dot net feature in a Dockerfile
```
FROM mcr.microsoft.com/vscode/devcontainers/base:jammy
LABEL maintainer="Bob codeql team"

# tzdata install needs to be non-interactive
ENV DEBIAN_FRONTEND=noninteractive

ARG USERNAME=vscode
ENV CODEQL_HOME /usr/local/codeql-home

# create user, install/update basics and python
# RUN adduser --home ${CODEQL_HOME} ${USERNAME} && \
RUN apt-get update && \
    apt-get upgrade -y && \
    apt-get install -y --no-install-recommends \
    	software-properties-common \
      nodejs \
    	vim \
    	curl \
    	wget \
    	git \
    	build-essential \
    	unzip \
    	apt-transport-https \
    	gnupg \
    	g++ \
    	make \
    	gcc \
    	apt-utils \
      rsync \
    	file \
      dos2unix \
    	gettext && \
      apt-get clean

# Install latest codeQL
RUN mkdir -p \
    ${CODEQL_HOME}/codeql-repo \
    /opt/codeql

# install codeql binary
RUN CODEQL_VERSION="v2.13.1" && \
    wget -q https://github.com/github/codeql-cli-binaries/releases/download/${CODEQL_VERSION}/codeql-linux64.zip -O /tmp/codeql_linux.zip && \
    unzip /tmp/codeql_linux.zip -d ${CODEQL_HOME} && \
    rm /tmp/codeql_linux.zip
ENV PATH="${CODEQL_HOME}/codeql:${PATH}"

# install language packs
RUN codeql pack download codeql/csharp-queries

```

and a devcontainer.json
```
	"build": {
		"dockerfile": "Dockerfile",
		"args": {
			"tag": "devcontainer"
		}
	},
	"features": {
		"ghcr.io/devcontainers/features/docker-in-docker:2": {},
		"ghcr.io/devcontainers/features/dotnet:1": { "version": "6.0" },
		"ghcr.io/devcontainers/features/kubectl-helm-minikube:1": {
			"version": "1.22",
			"helm": "3.7",
			"minikube": "none"
		},
		"ghcr.io/devcontainers/features/azure-cli:1": {},
		"ghcr.io/devcontainers/features/node:1": {}
	},

```

fails installing the dotnet feature
```
(*) Installing .NET CLI...
[...]

PR #571, opened 2023-06-13
Add `prerelease` extension tag

bamurtaugh:
Part of https://github.com/devcontainers/templates/issues/164

To use the C# Dev Kit today, users need the pre-release version of the C# extension.

The other main change to ensure the Feature works well in all setups is https://github.com/devcontainers/features/issues/560 (a fix is in-progress).

@timheuer would love to know if you had any other thoughts/feedback on this Feature. Thanks!

### DIFF
--- a/src/dotnet/devcontainer-feature.json
+++ b/src/dotnet/devcontainer-feature.json
@@ -34,7 +34,7 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "ms-dotnettools.csharp"
+                "ms-dotnettools.csharp@prerelease"
             ]
         }
     },


src/dotnet/devcontainer-feature.json
```diff
@@ -34,7 +34,7 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "ms-dotnettools.csharp"
+                "ms-dotnettools.csharp@prerelease"
```
samruddhikhandale, line 37: We have always added support for extensions which are production ready (ie. in release phase). I don't believe we have added any extension in a `prerelease` phase. 🤔 

> To use the C# Dev Kit today, users need the pre-release version of the C# extension.

I understand that its needed for `C# Dev Kit `, however, prerelease versions are generally used for pre-production/testing and might contain bugs. Not sure, if we should recommend them to users 🤔 

Wondering if we could simply wait until the changes reached prod? 
@bamurtaugh What do you think?

timheuer: Thanks for the tag -- yeah I agree...in 'official' dev container it probably isn't right to do this yet. We're trying to rapidly exit pre-release as best we can.  Target is end summer.

bamurtaugh: Thanks for the thoughts! I'll go ahead and close this and https://github.com/devcontainers/images/pull/619.

Issue #577, opened 2023-06-16
[debian bookworm] Python Feature fails to install with 'error: externally-managed-environment'


joshspicer:
ref: #576 

Test run: https://github.com/devcontainers/features/actions/runs/5293497423/jobs/9581722477#step:4:1608

```
[2023-06-16T19:24:31.534Z] #18 36.34 Processing triggers for libc-bin (2.36-9) ...
#18 36.39 Installing Python tools...
#18 36.42 Updating pip...

[2023-06-16T19:24:31.810Z] #18 36.77 error: externally-managed-environment
#18 36.77 
#18 36.77 × This environment is externally managed
#18 36.77 ╰─> To install Python packages system-wide, try apt install
#18 36.77     python3-xyz, where xyz is the package you are trying to
#18 36.77     install.
#18 36.77     
#18 36.77     If you wish to install a non-Debian-packaged Python package,
#18 36.77     create a virtual environment using python3 -m venv path/to/venv.
#18 36.77     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
#18 36.77     sure you have python3-full installed.
#18 36.77     
#18 36.77     If you wish to install a non-Debian packaged Python application,
#18 36.77     it may be easiest to use pipx install xyz, which will manage a
#18 36.77     virtual environment for you. Make sure you have pipx installed.
#18 36.77     
#18 36.77     See /usr/share/doc/python3.11/README.venv for more information.
[2023-06-16T19:24:31.810Z] 
#18 36.77 
#18 36.77 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
#18 36.77 hint: See PEP 668 for the detailed specification.
```

joshspicer:
Looks like the azure-cli (which attempts to install python) is effected the same way: https://github.com/devcontainers/features/actions/runs/5294360726/jobs/9583631989

joshspicer:
Good info on this change: https://pythonspeed.com/articles/externally-managed-environment-pep-668/#:~:text=If%20you%20wish%20to%20install,you%20have%20python3%2Dfull%20installed.

vermegi:
Affirmative on azure-cli install being effected by this change on Debian bookworm as well: https://github.com/Azure/azure-cli/issues/26737 and https://github.com/devcontainers/features/issues/589 

Reverting to -bullseye, this is (for now) working fine.

[...]

PR #578, opened 2023-06-16
Azure CLI: Compatibility with Debian `bookworm`

joshspicer:
closes https://github.com/devcontainers/features/issues/589
related https://github.com/devcontainers/features/issues/577

Changes needed to the Azure CLI Feature post Debian bookworm / [PEP 668](https://peps.python.org/pep-0668/#abstract)

When azure-cli binaries aren't available from `apt`, checks if  `pipx` if available in apt and if so, attempts to use that to install the Azure CLI.  From the base images I tested, `pipx` is only available in the `bookworm` apt repos (and that's the goal for this PR!).

### DIFF
--- a/src/azure-cli/install.sh
+++ b/src/azure-cli/install.sh
@@ -16,13 +16,17 @@ AZ_VERSION=${VERSION:-"latest"}
 AZ_EXTENSIONS=${EXTENSIONS}
 AZ_INSTALLBICEP=${INSTALLBICEP:-false}
 
-USERNAME="${USERNAME:-"${_REMOTE_USER:-"root"}"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 AZCLI_ARCHIVE_ARCHITECTURES="amd64"
 AZCLI_ARCHIVE_VERSION_CODENAMES="stretch buster bullseye bionic focal jammy"
 
 if [ "$(id -u)" -ne 0 ]; then
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    exit 1
+fi
+
+if [ -z "${_REMOTE_USER}" ]; then
+    echo -e 'Feature script must be executed by a tool that implements the dev container specification. See https://containers.dev/ for more information.'
     exit 1
 fi
 
@@ -152,7 +156,7 @@ install_with_pipx() {
     fi
 
     pipx install azure-cli${ver}
-    chown -hR ${USERNAME}:${USERNAME} "${REMOTE_USER_LOCAL_FOLDER}"   
+    chown -hR ${_REMOTE_USER}:${_REMOTE_USER} "${REMOTE_USER_LOCAL_FOLDER}"   
 }
 
 install_with_complete_python_installation() {
@@ -218,7 +222,7 @@ if [ ${#AZ_EXTENSIONS[@]} -gt 0 ]; then
     for i in "${extensions[@]}"
     do
         echo "Installing ${i}"
-        su ${USERNAME} -c "az extension add --name ${i} -y" || continue
+        su ${_REMOTE_USER} -c "az extension add --name ${i} -y" || continue
     done
 fi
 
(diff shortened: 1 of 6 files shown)


influencep (approved):

crazy4pi314:
Linking #577 
I think this is not going to be a great solution long term, as it creates a confusing situation for the user (where are packages getting installed) and is not following the best pratice guidance from the Python Language commitee. Installing Python in a venv is the reccomended path to ensure installs and packages are managed in the right scopes.

Problem tl;dr as show in the linked issue: OS provide Python distros can mix package management with os package managers with pip leading to confustion in where packages are. [PEP 668](https://peps.python.org/pep-0668/#abstract) was published as guidance for how to properly isolate the OS python install, and best pratice for a variety of situations where you might be trying to install Python.

crazy4pi314:
Linking #577 
I think this is not going to be a great solution long term, as it creates a confusing situation for the user (where are packages getting installed) and is not following the best pratice guidance from the Python Language commitee. Installing Python in a venv is the reccomended path to ensure installs and packages are managed in the right scopes.

Problem tl;dr as show in the linked issue: OS provide Python distros can mix package management with os package managers with pip leading to confustion in where packages are. [PEP 668](https://peps.python.org/pep-0668/#abstract) was published as guidance for how to properly isolate the OS python install, and best pratice for a variety of situations where you might be trying to install Python.

joshspicer:
@crazy4pi314 In this instance (installing the Azure CLI which need be installed system-wide), what changes would you recommend I make in this PR?  There are lots of instances in this repo where we're using pip to install tools globally, so having one good example of how to proceed would be useful.

influencep:
Yes

influencep (approved):

crazy4pi314:
Ok so for the azure-cli one, is there a need to have the full install of python as well as installing the CLI? If not, I could suggest just using pipx as provided from apt by putting this in `install_using_pip` where appropriate:
```
apt-get install pipx
pipx ensurepath
pipx install azure-cli
```
I tested this on "mcr.microsoft.com/devcontainers/base:bookworm" but would be good to try in the pipeline. 

Looking now more at the messier Python one!

crazy4pi314:
For the python one, I think there is a much more nuanced question there, there are a lot of ways to "install python" and that do different things such that not everyone will be happy. I think that's ok as this is an opnionated feature, and I would think that if I as the user wanted to "install python version x (or latest)" this would install python via wget-ing the appropriate build of python from Python.org, setting up symlinks, and setting up appropriate venvs and path defaults." This is not really at all what the exsisting feature does, it pretty heavily relies on using the system packaged python (the default even) in the ways the PEP is trying to prevent. 

The error message itself:
```
#18 36.77     If you wish to install a non-Debian-packaged Python package,
#18 36.77     create a virtual environment using python3 -m venv path/to/venv.
#18 36.77     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
#18 36.77     sure you have python3-full installed.
```
Would suggest that changing [`install_python`](https://github.com/devcontainers/features/blob/d934503a050ba84e6b42a006aacd891c4088eb62/src/python/install.sh#L328) to not allow "os-provided" or "system" and install python3-full, then use that to create a venv that then is set as default on path. This is similar to the last option suggested in [this blog](https://pythonspeed.com/articles/activate-virtualenv-dockerfile/):
```
ENV VIRTUAL_ENV=/opt/venv
RUN python3 -m venv $VIRTUAL_ENV
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```

I looked at how the Python docker image does it, which always wgets source tars from python.org and then builds. 

Idk of a short "fix" here that would have the behaviour I would expect, I would probably pick build from scratch/wget the prebuilt from python.org as default, installing create a venv and disable the "os-provided" option in version selecting (as that is what we are trying to prevent here.)

joshspicer:
> Ok so for the azure-cli one, is there a need to have the full install of python as well as installing the CLI?

One of the dependencies is a version of python: https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=script

src/azure-cli/install.sh
```diff
@@ -127,7 +130,37 @@ install_using_apt() {
...
+        ver="==${AZ_VERSION}"
+    fi
+
+    install_with_pipx "${ver}" || install_with_complete_python_installation "${ver}" || return 1
+}
+
+install_with_pipx() {
+    local ver="$1"
+
+    local REMOTE_USER_LOCAL_FOLDER="${_REMOTE_USER_HOME}/.local"
```
samruddhikhandale, line 148: FYI - [Bug: _REMOTE_USER_HOME is empty when executing a Feature where common-utils configures the user](https://github.com/devcontainers/cli/issues/331)

samruddhikhandale (approved):
Looks good to me.

src/azure-cli/install.sh
joshspicer: I might want to put it somewhere global anyway 🤔 .  I just noticed that the existing pip branch here writes to a global location.

samruddhikhandale (approved):
Looks good!

Issue #581, opened 2023-06-19
[Terraform] tflint update causes builds using terraform feature to fail


MathBouma:
The latest release of tflint (0.47.0) has broken terraform feature.

All devcontainers using terraform latest are failing as a result of a change in tflint.
The container script relies on a checksums.txt.sig file but this wasn't present in the latest release.
I have opened a ticket with tflint:
https://github.com/terraform-linters/tflint/issues/1783

In case the issue needs to be resolved there, but if not a change will be required to this script to workaround this.

kf6kjg:
And they've closed the issues saying the `checksums.txt.sig` was deprecated and now removed, and there's a newer more correct way. Meaning it's now a problem for this project.

joshspicer:
Thanks for raising! I've also run into this, found your closed issue on the tflint project, and eventually made my way back here 😄

Looking into a fix. Thanks @cjam for sharing the workaround!

joshspicer:
Version `1.3.1` of the `terraform` Feature is pinned to tflint `0.46.1` by default (can override with any version you'd like via the Feature's options).  This should unblock folks that are using Feature defaults.

Keeping this issue open while a forward fix is worked on 👍

samruddhikhandale:
@joshspicer should we close this issue as https://github.com/devcontainers/features/pull/586 is merged?

PR #586, opened 2023-06-20
Terraform: Add cosign integrity check for TFLint

joshspicer:
ref: https://github.com/devcontainers/features/issues/581

Updates the `terraform` Feature to use [cosign](https://github.com/sigstore/cosign) to validate signatures.  This change also unpins the temporary mitigation added in https://github.com/devcontainers/features/pull/583

### DIFF
--- a/src/terraform/devcontainer-feature.json
+++ b/src/terraform/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "terraform",
-    "version": "1.3.1",
+    "version": "1.3.2",
     "name": "Terraform, tflint, and TFGrunt",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/terraform",
     "description": "Installs the Terraform CLI and optionally TFLint and Terragrunt. Auto-detects latest version and installs needed dependencies.",
@@ -24,8 +24,8 @@
                 "0.47.0",
                 "0.46.1"
             ],
-            "default": "0.46.1",
-            "description": "Tflint version (Default value temporarily pinned to version 0.46.1: https://github.com/devcontainers/features/issues/581)"
+            "default": "latest",
+            "description": "Tflint version (https://github.com/terraform-linters/tflint/releases)"
         },
         "terragrunt": {
             "type": "string",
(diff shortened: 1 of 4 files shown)


src/terraform/install.sh
```diff
@@ -158,6 +157,26 @@ check_packages() {
...
+        curl -L "https://github.com/sigstore/cosign/releases/latest/download/cosign_${LATEST_COSIGN_VERSION}_${architecture}.deb" -o /tmp/cosign_${LATEST_COSIGN_VERSION}_${architecture}.deb
+        
+        dpkg -i /tmp/cosign_${LATEST_COSIGN_VERSION}_${architecture}.deb
+        rm /tmp/cosign_${LATEST_COSIGN_VERSION}_${architecture}.deb
+    fi
+    if ! type cosign > /dev/null 2>&1; then
+        echo "(!) Failed to install cosign."
+        exit 1
+    fi
+    cosign version
```
samruddhikhandale, line 177: Wondering if this is intentional, or was used for debugging.

src/terraform/devcontainer-feature.json
```diff
@@ -24,8 +24,8 @@
                 "0.47.0",
                 "0.46.1"
             ],
-            "default": "0.46.1",
```
samruddhikhandale, line 27: Double checking, does verifying integrity with co-signing works with versions <= 0.46.1 ?

samruddhikhandale (approved):
Great, thanks! Left two clarification comments.

src/terraform/devcontainer-feature.json
samruddhikhandale: Wondering if we should add a test scenario for validation.

joshspicer: very good point, I shouldn't have removed the previous implementation.

src/terraform/install.sh
joshspicer: I kept it in to have the version in the logs

src/terraform/devcontainer-feature.json
joshspicer: Appears like it wasn't too necessary to do this since they've been publishing the cosign signatures for a long time. Nonetheless added in the fallback 👍

samruddhikhandale (approved):
🚀

PR #590, opened 2023-06-26
Dotnet: Fix Feature failures on ubuntu:jammy (when installing using apt)

samruddhikhandale:
Closes: https://github.com/devcontainers/features/issues/560

Fixes `dotnet` Feature failures for `ubuntu:jammy`:
- Installs incorrect apt package (eg. `dotnet-sdk-6.0-source-built-artifacts` instead of `dotnet-sdk-6.0`)
- Fails to install package from Microsoft feed
    - Ref: https://learn.microsoft.com/en-us/dotnet/core/install/linux-package-mixup?pivots=os-linux-ubuntu#i-need-a-version-of-net-that-isnt-provided-by-my-linux-distribution
    - https://github.com/dotnet/core/issues/7699
- Fails to install from Ubuntu feed
    - (eg: .NET 6 is available in jammy feeds, however, it fails to install & falls backs to MSFT feeds)

### DIFF
--- a/src/dotnet/install.sh
+++ b/src/dotnet/install.sh
@@ -194,11 +194,13 @@ install_using_apt() {
         curl -sSL ${MICROSOFT_GPG_KEYS_URI} | gpg --dearmor > /usr/share/keyrings/microsoft-archive-keyring.gpg
         echo "deb [arch=${architecture} signed-by=/usr/share/keyrings/microsoft-archive-keyring.gpg] https://packages.microsoft.com/repos/microsoft-${ID}-${VERSION_CODENAME}-prod ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/microsoft.list
 
-        cat << 'EOF' > /etc/apt/preferences
+        # See https://github.com/dotnet/core/issues/7699
+        cat << 'EOF' >> /etc/apt/preferences
 Package: *
 Pin: origin "packages.microsoft.com"
 Pin-Priority: 1001
 EOF
+
         apt-get update -y
     fi
 
--- a/src/dotnet/install.sh
+++ b/src/dotnet/install.sh
@@ -147,9 +147,9 @@ apt_cache_package_and_version_soft_match() {
     . /etc/os-release
     local architecture="$(dpkg --print-architecture)"
 
+    apt-get update -y
     major_minor_version="$(echo "${requested_version}" | cut -d "." --field=1,2)"
-    package_name="$(apt-cache search "${partial_package_name}-[0-9].[0-9]" | awk -F" - " '{print $1}' | grep -m 1 "${partial_package_name}-${major_minor_version}")"
-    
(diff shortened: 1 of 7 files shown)


samruddhikhandale:
Failures for `debian:12 (bookworm)` are expected as there are no upstream .NET releases yet. 
See https://learn.microsoft.com/en-us/dotnet/core/install/linux-debian & https://github.com/dotnet/core/issues/8481

src/dotnet/install.sh
```diff
@@ -147,9 +147,9 @@ apt_cache_package_and_version_soft_match() {
     . /etc/os-release
     local architecture="$(dpkg --print-architecture)"
 
+    apt-get update -y
     major_minor_version="$(echo "${requested_version}" | cut -d "." --field=1,2)"
-    package_name="$(apt-cache search "${partial_package_name}-[0-9].[0-9]" | awk -F" - " '{print $1}' | grep -m 1 "${partial_package_name}-${major_minor_version}")"
-    
+    package_name="$(apt-cache search "${partial_package_name}-[0-9].[0-9]" | awk -F" - " '{print $1}' | grep -v "source-built-artifacts" | grep -m 1 "${partial_package_name}-${major_minor_version}")"
```
joshspicer, line 152: I feel like we could update the `${partial_package_name}-[0-9].[0-9]` regex to capture the expected name of the package better. Maybe something like this (untested):

```suggestion
    package_name="$(apt-cache search "${partial_package_name}-[0-9].[0-9]$" | awk -F" - " '{print $1}' | grep -m 1 "${partial_package_name}-${major_minor_version}")"
```

src/dotnet/install.sh
```diff
@@ -193,6 +193,14 @@ install_using_apt() {
         # Import key safely and import Microsoft apt repo
         curl -sSL ${MICROSOFT_GPG_KEYS_URI} | gpg --dearmor > /usr/share/keyrings/microsoft-archive-keyring.gpg
         echo "deb [arch=${architecture} signed-by=/usr/share/keyrings/microsoft-archive-keyring.gpg] https://packages.microsoft.com/repos/microsoft-${ID}-${VERSION_CODENAME}-prod ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/microsoft.list
+
+        # See https://github.com/dotnet/core/issues/7699
```
joshspicer, line 197: Do you think we should remove this after the install completes? I'm not sure if this will cause issues in subsequent `apt` operations.

src/dotnet/install.sh
samruddhikhandale: Nice 💡 
Udpated

src/dotnet/install.sh
samruddhikhandale: Yep, we can clean this! Updated

joshspicer (dismissed):
If you have the time, it might be worth quickly updating the dotnet/test.sh script here for bookworm.  From a glance it appears like it's installing fine but `latest` for bookworm is dotnet 8 instead of 7 (and test is validating for 7)

https://github.com/devcontainers/features/actions/runs/5407307268/jobs/9825172418?pr=590#step:4:452

joshspicer (approved):

PR #594, opened 2023-06-27
Use `\$` as the final character of `PS1`

casaroli:
If the effective UID is `0`, display `#`, otherwise, display `$` in the end of `PS1`

Since we use double quotes `"`, we should use `\\`.

I believe this was the intended behavior.

Current:
current

With the change applied:
new

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.0.10",
+    "version": "2.0.11",
     "name": "Common Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",


joshspicer:
would you mind sharing some screenshots of both the bug you're fixing with this and the result of the change?

casaroli:
> would you mind sharing some screenshots of both the bug you're fixing with this and the result of the change?

I updated the PR description, please check it.

Thank you.

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -18,7 +18,7 @@ __bash_prompt() {
         fi`'
     local lightblue='\[\033[1;34m\]'
     local removecolor='\[\033[0m\]'
-    PS1="${userpart} ${lightblue}\w ${gitbranch}${removecolor}\$ "
+    PS1="${userpart} ${lightblue}\w ${gitbranch}${removecolor}\\$ "
```
samruddhikhandale, line 21: Looks good to me! However, @Chuxel double checking if this wasn't intentional.

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -18,7 +18,7 @@ __bash_prompt() {
         fi`'
```
samruddhikhandale, line 18: Can we bump the patch version in here, thanks! 
https://github.com/devcontainers/features/blob/c408d345524651c8c0224520e193454c18b88a81/src/common-utils/devcontainer-feature.json#L3

casaroli: I added the commit. Please let me know if something is wrong.

casaroli:

PR #595, opened 2023-06-28
docker-in-docker: Bookworm support

joshspicer:


src/docker-in-docker/install.sh
```diff
@@ -370,28 +372,40 @@ dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAU
...
+    }
+
+    # Set cgroup nesting, retrying if necessary
+    set +e
+    set_cgroup_nesting
+    if [ $? -ne 0 ]; then
+        echo >&2 "cgroup v2: failed to enable nesting, retrying once..."
+        set -e
+        set_cgroup_nesting
     fi
```
samruddhikhandale, line 397: Suggestion: Should we prepend one tab for all the lines in between `set +e` and `set -e`?

src/docker-in-docker/install.sh
```diff
@@ -406,6 +420,21 @@ dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAU
 
     # Start docker/moby engine
     ( dockerd $CUSTOMDNS $DEFAULT_ADDRESS_POOL > /tmp/dockerd.log 2>&1 ) &
+
+    # Wait for API to be listening on unix socket.
+    echo "Waiting for 'docker-in-docker' to be ready..."
```
samruddhikhandale, line 425: Question: Is this change specifically needed for docker to work with bookworm? or is it docker in general?

src/docker-in-docker/install.sh
```diff
@@ -370,28 +372,40 @@ dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAU
...
+            # enable controllers
+            sed -e 's/ / +/g' -e 's/^/+/' < /sys/fs/cgroup/cgroup.controllers \
+                > /sys/fs/cgroup/cgroup.subtree_control
+        fi
+    }
+
+    # Set cgroup nesting, retrying if necessary
+    set +e
+    set_cgroup_nesting
+    if [ $? -ne 0 ]; then
```
samruddhikhandale, line 393: > Question: Is this change specifically needed for docker to work with bookworm? or is it docker in general?

Same question here!

joshspicer: This was to make sure that the tests now passed

src/docker-in-docker/install.sh
joshspicer: And i suspect could also be the cause of some failures in the wild

PR #605, opened 2023-07-05
Pin version of TFLint's GPG key URL

wata727:
See also https://github.com/terraform-linters/tflint/pull/1800

The GPG key is scheduled to be removed, so pin to the last used version.

### DIFF
--- a/src/terraform/install.sh
+++ b/src/terraform/install.sh
@@ -27,7 +27,7 @@ TFSEC_SHA256="${TFSEC_SHA256:-"automatic"}"
 TERRAFORM_DOCS_SHA256="${TERRAFORM_DOCS_SHA256:-"automatic"}"
 
 TERRAFORM_GPG_KEY="72D7468F"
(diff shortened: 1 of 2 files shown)


src/terraform/install.sh
```diff
@@ -27,7 +27,7 @@ TFSEC_SHA256="${TFSEC_SHA256:-"automatic"}"
 TERRAFORM_DOCS_SHA256="${TERRAFORM_DOCS_SHA256:-"automatic"}"
```
samruddhikhandale, line 27: Can we bump the patch version in here? thanks!

https://github.com/devcontainers/features/blob/79a042923039bb6c974a78ffed9db264ff5e4d57/src/terraform/devcontainer-feature.json#L3

samruddhikhandale (approved):
Thank you for the heads up, we appreciate it! 
Left a minor comment.

src/terraform/install.sh
wata727: Thanks. Bumped to 1.3.3.

samruddhikhandale (approved):
Thank you!

Issue #601, opened 2023-06-29
Ruby Feature: Fails if the dev container has the same ruby version installed


samruddhikhandale:
Repro:

.devcontainer.json
```jsonc
{
  "image": "mcr.microsoft.com/devcontainers/universal:2.4.0",
  "features": {
    "ghcr.io/devcontainers/features/ruby:1": {}
  }
}
```

Error: 
```
#17 529.2 Updating /etc/bash.bashrc and /etc/zsh/zshrc...
#17 529.2 ln: failed to create symbolic link '/usr/local/rvm/gems/default/bin/ruby': File exists
#17 529.2 ERROR: Feature "Ruby (via rvm)" (ghcr.io/devcontainers/features/ruby) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/ruby for help troubleshooting this error.
```

PR #609, opened 2023-07-10
Ruby Feature [bug fix] : Fails if the dev container has the same ruby version installed

samruddhikhandale:
Closes https://github.com/devcontainers/features/issues/601


test/ruby/scenarios.json
```diff
@@ -13,5 +13,13 @@
         "features": {
             "ruby": {}
         }
+    },
+    "image_and_feature_installs_same_ruby_version": {
+        "image": "mcr.microsoft.com/devcontainers/universal:2.4.0",
```
joshspicer, line 18: Is it necessary to use the universal image here?  If someone runs these tests locally (or if they're run in the pipeline) this large image will need to be pulled first)

src/ruby/install.sh
```diff
@@ -177,9 +177,11 @@ fi
 find_version_from_git_tags RUBY_VERSION "https://github.com/ruby/ruby" "tags/v" "_"
 
 # Just install Ruby if RVM already installed
-if [ $(rvm --version) != "" ]; then
+if rvm --version > /dev/null; then
     echo "Ruby Version Manager already exists."
-    if [ "${RUBY_VERSION}" != "none" ]; then
+    if [[ "$(ruby -v)" = *"${RUBY_VERSION}"* ]]; then
```
joshspicer, line 182: At this point I believe `RUBY_VERSION` is a fully expanded version (from line 177), right?

src/ruby/install.sh
```diff
@@ -273,8 +278,10 @@ if [ "${SKIP_RBENV_RBUILD}" != "true" ]; then
             ln -s /usr/local/share/ruby-build /home/${USERNAME}/.rbenv/plugins/ruby-build
         fi
 
-        ln -s /usr/local/rvm/rubies/default/bin/ruby /usr/local/rvm/gems/default/bin 
-        
+        if [ ! -f /usr/local/rvm/gems/default/bin/ruby ]; then
+            ln -s /usr/local/rvm/rubies/default/bin/ruby /usr/local/rvm/gems/default/bin
```
joshspicer, line 282: Do you know why we do this?

Princeprince559 (approved):

test/ruby/scenarios.json
samruddhikhandale: This issue only occurs with the universal image, as it uses this Ruby Feature to install a specific version of ruby. 😓 @joshspicer wondering, do you know of any other image which does that? 

I understand the pain point here, but this is the exact same scenario which is currently breaking.

src/ruby/install.sh
samruddhikhandale: Yes, that's correct.

image

src/ruby/install.sh
samruddhikhandale: Oryx expects ruby to be installed at  `/usr/local/rvm/gems/default/bin`, else it breaks the oryx flow of automatic installing gems. Hence, this bit was added. Added a comment!

test/ruby/scenarios.json
samruddhikhandale: The action is [running out of disk space](https://github.com/devcontainers/features/actions/runs/5627383080/job/15249820817#step:4:6294) (this is a manual run with trace logs which surfaced the cause) and failing due to the added test ^

Removing the test, validated locally that it works as expected 👇 : 

image

joshspicer (approved):

PR #614, opened 2023-07-16
Common utils: Add config to remove zsh rc files from

naturedamends:
My dotfiles adds `.zshrc` itself via the `postCreateCommand`.

plz squash on merge.

## Proposal 
Add flag that disables overwriting the user(s) `$HOME/.zshrc` with a file containing the devcontainers default zsh theme


src/common-utils/main.sh
```diff
@@ -471,6 +472,9 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
             cp -rf "${user_rc_file}" "${oh_my_install_dir}" /root
             chown -R ${USERNAME}:${group_name} "${oh_my_install_dir}" "${user_rc_file}"
         fi
+        if [ "$INSTALL_OH_MY_ZSH_CONFIG" != "true" ]; then
+            rm -f "${user_rc_file}" "/root/.zshrc"
```
samruddhikhandale, line 476: Question: Instead of configuring the `user_rc_file` and then deleting it based on `INSTALL_OH_MY_ZSH_CONFIG`, why not create and configure the file only if `INSTALL_OH_MY_ZSH_CONFIG` is true?

samruddhikhandale (commented):
I like the idea of adding the ability to control the config for oh-my-zsh 👍 (// cc @joshspicer @Chuxel for any other thoughts)

Thanks for opening the PR and being willing to iterate based on feedbacks.

> I will update the README etc or anything else if you give me an approval nod. Need this to complete setup

Thank you, few things which should be added to this PR -

1. Adding this ability in the form of [Feature option](https://containers.dev/implementors/features/#options-property)
2. README.md is auto-generated from devcontainer-feature.json & NOTES.md. Hence, you won't need to update the README file
3. Adding [test scenarios](https://github.com/devcontainers/cli/blob/main/docs/features/test.md)
4. Bumping the minor version - as this PR is adding in a new Feature option

src/common-utils/main.sh
naturedamends: Erm, it was just easier for me to create the PR, without moving around any code, its just succinct

naturedamends: I was more worried with is the json configuration more cumbersome than the rm command, but suppose its worth in the for UI configuration prompt.

Just need user configured default templates now to auto add this to all my dev-containers.

naturedamends: @samruddhikhandale I'm having a trouble: 

- running the tests, you don't know of a way I can run docker build from inside a dev container, reading from the container filesystem? I made a rule that I would run everything inside a virtual machine and keep my main computer free of dependencies.
- The change I have just committed is technically major? Since it not does not override the users zshrc if it already exists, from the base container, or copied from the /etc/skel folder. I can change this if requested? There dons't seem to be a any tests that fail. Just hoping it runs in the github actions.

naturedamends:

src/common-utils/main.sh
```diff
@@ -455,20 +457,27 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
-        echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
-        sed -i -e 's/ZSH_THEME=.*/ZSH_THEME="devcontainers"/g' ${user_rc_file}
 
         # Add Dev Containers theme
         mkdir -p ${oh_my_install_dir}/custom/themes
         cp -f "${FEATURE_DIR}/scripts/devcontainers.zsh-theme" "${oh_my_install_dir}/custom/themes/devcontainers.zsh-theme"
         ln -s "${oh_my_install_dir}/custom/themes/devcontainers.zsh-theme" "${oh_my_install_dir}/custom/themes/codespaces.zsh-theme"
 
+        # Attempt adding devcontainers .zshrc, does nothing given an existing file
+        if [[ (! -f "$user_rc_file") && ("$INSTALL_OH_MY_ZSH_CONFIG" = "true") ]]; then
```
samruddhikhandale, line 467: > (! -f "$user_rc_file") 

Question: Provided `installOhMyZshConfig:true`, shouldn't the contents be simply overwrriten by this Feature? If a user does not want that behavior, they can always pass `installOhMyZshConfig:false`

samruddhikhandale, line 467: > (! -f "$user_rc_file") 

Question: Provided `installOhMyZshConfig:true`, shouldn't the contents be simply overwrriten by this Feature? If a user does not want that behavior, they can always pass `installOhMyZshConfig:false`

samruddhikhandale: We'd also need to update the description https://github.com/devcontainers/features/pull/614/files#diff-6dcca292e999f6e647eaaf8dec68633d7f1631c75140eeaf8b89683f0edb2bdcR26 (if we remove this condition)

samruddhikhandale (changes requested):
Thank you! Appreciate the changes ✨ 

Can we add test scenarios to validate this change with `installOhMyZshConfig ` set to `true` and `false`? Let me know if you need any help with it. 

See https://github.com/devcontainers/cli/blob/main/docs/features/test.md for reference.

src/common-utils/main.sh
samruddhikhandale: Have you tried using the `devcontainer features test` command? You can run that from within a dev container with docker-in-docker installed. You can simply open this repo with dev containers extension in VS Code or use GitHub codespaces.

samruddhikhandale: > The change I have just committed is technically major? Since it not does not override the users zshrc if it already exists, from the base container, or copied from the /etc/skel folder. I can change this if requested?

Yep, added a comment. See https://github.com/devcontainers/features/pull/614#discussion_r1278159617

src/common-utils/main.sh
naturedamends: The reason it now does not override is to support base dockerfiles that have an existing .Ashcroft

src/common-utils/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.0.10",
+    "version": "2.1.10",
```
samruddhikhandale, line 3:
```suggestion
    "version": "2.1.0",
```

src/common-utils/devcontainer-feature.json
```diff
@@ -20,6 +20,11 @@
             "default": true,
             "description": "Install Oh My Zsh!?"
         },
+        "installOhMyZshConfig": {
+            "type": "boolean",
+            "default": true,
+            "description": "Allow installing the default dev conatiner .zshrc templates?"
```
samruddhikhandale, line 26:
```suggestion
            "description": "Allow installing the default dev container .zshrc templates?"
```

src/common-utils/main.sh
samruddhikhandale: If that's the case, can the user pass `installOhMyZshConfig:false` to get around it?

My major concern with this change is that the Feature will start behaving differently than before (which might not be expected and could lead to unwanted changes). To make a change like this, we'd need to bump major version of this Feature. We avoid bumping major versions (unless necessary), as it's bit slow for everyone to update their dev containers, people will be left behind and lose important upcoming updates.

We can open an issue and add this change in when we are going to bump major version. @naturedamends what do you think?

samruddhikhandale: https://github.com/devcontainers/features/pull/614/commits/ab94a9760bb8962ee6113330d080e75308916198 Thanks! ❇️

samruddhikhandale (changes requested):
Looks great, left some minor comments. Thanks!

src/common-utils/devcontainer-feature.json
```diff
@@ -20,6 +20,11 @@
             "default": true,
             "description": "Install Oh My Zsh!?"
         },
+        "installOhMyZshConfig": {
```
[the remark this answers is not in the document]
naturedamends: I have added some tests. But I can't run them. I've still not got round to getting `act` command working in my workspace dev container.

Can you run the ci actions?

samruddhikhandale (commented):
@naturedamends Looks like some of the zsh tests are failing, mind taking a look? Thanks! 

image

https://github.com/devcontainers/features/actions/runs/5805829101/job/15753628087?pr=614#step:4:17389

naturedamends:
> @naturedamends Looks like some of the zsh tests are failing, mind taking a look? Thanks! 
> 
> 
> 
> image
> 
> 
> 
> https://github.com/devcontainers/features/actions/runs/5805829101/job/15753628087?pr=614#step:4:17389

Done

samruddhikhandale:
Hmm, looks like they are still failing. @naturedamends do they pass locally for you? Let me know, I'd be happy to take a look. 

https://github.com/devcontainers/features/actions/runs/5811584487/job/15755869319?pr=614#step:4:17303

naturedamends:
> Hmm, looks like they are still failing. @naturedamends do they pass locally for you? Let me know, I'd be happy to take a look. 
> 
> 
> 
> https://github.com/devcontainers/features/actions/runs/5811584487/job/15755869319?pr=614#step:4:17303

I'll hopefully get the act cli working with docker outside next time. Then should be able run tests on my local machine for any repository without having 5 docker daemons running.

Feel free to merge your own or, fork this... if you want it faster.

Was sure that would work, but not read all the code yet. root user home path is not ~, is my first though on issue.

I'll mark work in process

naturedamends:
@samruddhikhandale Tests now pass for me on my fork. Can I have some free build mins haha.

Defo worth you looking over tho.

samruddhikhandale (approved):
Looks amazing~ 🚢 🎉 

Thank you for reiterating over all the feedbacks! 🙏

PR #618, opened 2023-07-19
Fix: Docker-in-docker for python - Arm64

samruddhikhandale:
Ref: https://github.com/devcontainers/features/issues/616 & https://github.com/yaml/pyyaml/issues/601#issuecomment-1638509577

Helps unblock docker-in-docker Feature installation for Python >= 3.10 on ARM64 machines.

Note: `pyyaml` cannot be installed with `pipx`, hence need to install `docker-compose` via `pip3`

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -265,7 +265,21 @@ if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "none" ]; then
                 pip3 install --disable-pip-version-check --no-cache-dir --user pipx
                 pipx_bin=/tmp/pip-tmp/bin/pipx
             fi
-            ${pipx_bin} install --pip-args '--no-cache-dir --force-reinstall' docker-compose
+
+            set +e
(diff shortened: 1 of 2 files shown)


src/docker-in-docker/install.sh
```diff
@@ -265,7 +265,21 @@ if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "none" ]; then
...
+                exit_code=$?
+            set -e
+
+            if [ ${exit_code} -ne 0 ]; then
+                # Temporary: https://github.com/devcontainers/features/issues/616
+                # See https://github.com/yaml/pyyaml/issues/601
+                echo "(*) Failed to install docker-compose via pipx. Trying via pip3..."
+
+                export PYTHONUSERBASE=/usr/local
+                pip3 install --disable-pip-version-check --no-cache-dir --user "Cython<3.0" pyyaml wheel docker-compose --no-build-isolation
```
samruddhikhandale, line 280: Note: `pyyaml` cannot be installed with `pipx`, hence need to install `docker-compose` via `pip3`

chrmarti (approved):

Issue #620, opened 2023-07-19
Docker-in-docker: Use v2 as default for docker compose version


samruddhikhandale:
> Important
> 
> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.
> 
> Compose V2 is included with all currently supported versions of Docker Desktop. For more information, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate).
> 
> Docker’s documentation refers to and describes Compose V2 functionality.

> source: 
https://docs.docker.com/compose/install/standalone/

I agree, and I think we should switch the default for `dockerDashComposeVersion` Feature option to `v2`, thanks for pointing that out.

_Originally posted by @samruddhikhandale in https://github.com/devcontainers/features/issues/619#issuecomment-1642480651_

samruddhikhandale:
proposing to switching from using pip to install, to Docker's instructions https://docs.docker.com/compose/install/standalone/ 

 > https://github.com/devcontainers/features/compare/main...jacobwoffenden:devcontainer-features:use-compose-v2

_Originally posted by @jacobwoffenden in https://github.com/devcontainers/features/issues/619#issuecomment-1642512468_

samruddhikhandale:
> https://github.com/devcontainers/features/compare/main...jacobwoffenden:devcontainer-features:use-compose-v2

> >  Thanks @jacobwoffenden, we appreciate all the help. Would you be willing to open the PR on this repo? As we are switching the  `dockerDashComposeVersion`'s default to `v2`, we'd need to bump the major version of the Feature. Hence, we need to be careful with the migration process as it's a breaking change.

_Originally posted by @samruddhikhandale in https://github.com/devcontainers/features/issues/619#issuecomment-1642554643_

dslatkin:
Does #873 also close this issue?

samruddhikhandale:
Yep, closing as fixed with https://github.com/devcontainers/features/pull/873

PR #621, opened 2023-07-19
Use v2 as default for Docker Compose version

jacobwoffenden:
Resolves https://github.com/devcontainers/features/issues/620

- Switches to use Docker Compose v2, [v1 is deprecated](https://docs.docker.com/compose/install/standalone/#:~:text=Install%20Compose%20standalone-,Important,-From%20July%202023)
- Installs binary from GitHub instead of using Pip ([source](https://docs.docker.com/compose/install/standalone/))
- Bump to v3


jacobwoffenden:
@samruddhikhandale I wasn't sure if there was an automated process for bumping versions, updating the README, so I did it manually

jacobwoffenden:
Regarding tests:

* [PR - Test Updated Features / test (docker-in-docker, debian:12)](https://github.com/devcontainers/features/actions/runs/5603716575/jobs/10250757682?pr=621)
  * I can see `bookworm` in https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/shared/settings.env#L57

test/docker-in-docker/docker_build.sh
```diff
@@ -9,8 +9,7 @@ source dev-container-features-test-lib
 check "docker-buildx" docker buildx version
 check "docker-build" docker build ./
 
-check "installs docker-compose v1 install" bash -c "type docker-compose"
-check "installs compose-switch" bash -c "[[ -f /usr/local/bin/compose-switch ]]"
```
samruddhikhandale, line 13: You would be able to revert these changes after you add back compose switch installation!

test/docker-in-docker/docker_build.sh
```diff
@@ -9,8 +9,7 @@ source dev-container-features-test-lib
...
-
+check "installs docker-compose v2 install" bash -c "type docker-compose"
+check
```
samruddhikhandale, line 13: nit: remove?

test/docker-in-docker/docker_build_no_compose.sh
```diff
@@ -9,8 +9,7 @@ source dev-container-features-test-lib
 check "docker-buildx" docker buildx version
 check "docker-build" docker build ./
 
-check "not installing compose skips docker-compose v1 install" bash -c "! type docker-compose"
-check "not installing compose skips compose-switch" bash -c "[[ ! -f /usr/local/bin/compose-switch ]]"
```
samruddhikhandale, line 13: Same here! ^

src/docker-in-docker/install.sh
```diff
@@ -244,73 +244,25 @@ echo "Finished installing docker / moby!"
...
+        find_version_from_git_tags docker_compose_version "https://github.com/docker/compose" "refs/tags/v"
     else
-        update-alternatives --set docker-compose /usr/local/bin/compose-switch
+        docker_compose_version="${DOCKER_DASH_COMPOSE_VERSION}"
     fi
+
+    echo "(*) Installing docker-compose ${docker_compose_version}..."
+
+    curl --location https://github.com/docker/compose/releases/download/v${docker_compose_version}/docker-compose-linux-${dockerComposeArch} \
+        --output /usr/local/bin/docker-compose
```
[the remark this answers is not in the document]
samruddhikhandale: Hmmm, now that I think about it --> ^ would fail if anyone passes `v2` (including test scenarios) 

I wonder if we should deprecate passing `v2` as option? 🤔 One can pass semvers instead (eg. 2 or 2.19 or 2.19.0). This would allow us to sync with other Feature options which does not contain the `v` prefix.

As we are bumping major versions, we could sneak this change in  (if we think that's needed, else we can condition on the `"refs/tags/v"` bit! ) 😬 

// cc @Chuxel @craiglpeters @bamurtaugh for thoughts on this change.

src/docker-outside-of-docker/README.md
```diff
@@ -19,7 +19,7 @@ Re-use the host docker socket, adding the Docker CLI to a container. Feature inv
 |-----|-----|-----|-----|
```
samruddhikhandale, line 19: Can we make similar changes as per my previous comments for `docker-in-docker` Feature? Thanks!

test/docker-outside-of-docker/scenarios.json
```diff
@@ -76,26 +77,6 @@
         },
         "containerUser": "vscode"
     },
-    "docker_dash_compose_v1": {
```
samruddhikhandale, line 79: > You would be able to revert these changes after you add back compose switch installation! 

Can we add a test scenario which validates specific version/semver changes? Thanks!

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -24,11 +24,10 @@
...
             "enum": [
                 "none",
-                "v1",
-                "v2"
+                "latest"
             ],
-            "default": "v1",
-            "description": "Default version of Docker Compose (v1 or v2 or none)"
+            "default": "latest",
+            "description": "Default version of Docker Compose (latest, none or specific version number e.g. 2.20.2)"
```
samruddhikhandale, line 30:
```suggestion
            "description": "Select or enter a Docker Compose version."
```

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -24,11 +24,10 @@
             "type": "string",
             "enum": [
```
samruddhikhandale, line 25:
```suggestion
            "proposals": [
```

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -24,11 +24,10 @@
...
                 "none",
-                "v1",
-                "v2"
+                "latest"
```
samruddhikhandale, line 27:
```suggestion
                "latest",
                "2.20.1",
                "2.20",
                "2"
```

samruddhikhandale (commented):
Thanks you so much for taking the time to contribute this PR, we really appreciate the support! ✨ 

I have left some thoughts/comments, thanks again!

src/docker-in-docker/devcontainer-feature.json
samruddhikhandale: Depends on https://github.com/devcontainers/features/pull/621#discussion_r1270090974

src/docker-in-docker/install.sh
```diff
@@ -244,73 +244,25 @@ echo "Finished installing docker / moby!"
...
+        dockerComposeArch="x86_64"
+    elif [ "${architecture}" == "arm64" ]; then
+        dockerComposeArch="aarch64"
     fi
 
-    # Install docker-compose switch if not already installed - https://github.com/docker/compose-switch#manual-installation
-    current_v1_compose_path="$(which docker-compose)"
-    target_v1_compose_path="$(dirname "${current_v1_compose_path}")/docker-compose-v1"
-    if ! type compose-switch > /dev/null 2>&1; then
-        echo "(*) Installing compose-switch..."
```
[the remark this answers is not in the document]
jacobwoffenden: I'm not sure I understand the need for `compose-switch`, based on it's description

>Compose Switch is a replacement to the Compose V1 docker-compose (python) executable. It translates the command line into Compose V2 docker compose then run the latter.

During the installation, it already installs `docker-compose-plugin` from APT which enables `$ docker compose`, the part I'm updating installs https://github.com/docker/compose a.k.a Docker Compose v2 into `/usr/local/bin/docker-compose`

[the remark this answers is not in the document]
samruddhikhandale: > During the installation, it already installs docker-compose-plugin from APT which enables $ docker compose, the part I'm updating installs https://github.com/docker/compose a.k.a Docker Compose v2 into /usr/local/bin/docker-compose

Installing docker compose v2 is perfect and that's what is required. In this context of the comment, it provides the `docker compose` command (notice the space here)

> Can we keep the installation for docker compose-switch?

With the PR changes, tools/scripts which still use the `docker-compose` command (which was provided by v1) will fail. We don't have to install v1 to serve this scenario, however, we need to make sure that `docker-compose` command (notice the hyphen here) works. 

We can achieve this by either installing the docker-compose switch or by adding a new alias. Even v2 Docker Desktop supports the use of `docker-compose` command (they use aliasing, so aliasing or compose switch anything is fine). So it feels right to add this support in the Feature as well.

Feel free to raise any questions/concerns!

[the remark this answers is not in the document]
jacobwoffenden: Sorry, maybe I didn't articulate my point properly in my previous comment. I still don't think compose-switch is required here.

During installation, either `docker-compose-plugin` or `moby-compose` is installed from APT depending on `if [ "${USE_MOBY}" = "true" ]`, this enables the use of `$ docker compose`

But also during installation, `if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "none" ]`, then it installs https://github.com/docker/compose, which to my understanding is actually the same binary as the package from APT, and this is installed into `/usr/local/bin/docker-compose`, so they are functionally the same, the difference being installation method

Packaged: https://docs.docker.com/compose/install/linux/#install-using-the-repository

VS

Manual: https://docs.docker.com/compose/install/standalone/

Which actually raises the question, is this step even required? Instead, is the right approach just to add an alias `alias docker-compose='docker compose'`?

Note:

Running the installation script with defaults results in

```bash
root ➜ /workspace $ which docker
/usr/bin/docker
root ➜ /workspace $ docker compose version
Docker Compose version 2.20.2+azure-1
root ➜ /workspace $ which docker-compose
/usr/local/bin/docker-compose
root ➜ /workspace $ docker-compose version
Docker Compose version v2.20.2
```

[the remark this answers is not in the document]
samruddhikhandale: Thanks for the explanation. 

> Even v2 Docker Desktop supports the use of docker-compose command (they use aliasing, so aliasing or compose switch anything is fine).

> Which actually raises the question, is this step even required? Instead, is the right approach just to add an alias alias docker-compose='docker compose'?

Yep, aliasing should be fine as well 👍

eitsupi:
Is there any reason why this was not merged?

it seems that docker-outside-of-docker already defaults to compose V2.

https://github.com/devcontainers/features/blob/0cf690ee14279d9e7ae5e1308950d6067d3aba81/src/docker-outside-of-docker/devcontainer-feature.json#L23-L32

PR #625, opened 2023-07-21
Remove deprecated Python settings

luabud:
Hi there!

The `python.linting` and `python.formatting` settings have been deprecated as in the latest Python extension pre-release version, as we're migrating our tooling support to extensions ([see here for more details](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions)). 

We figured we'd submit PRs to remove them from devcontainers configs 😊


src/python/devcontainer-feature.json
```diff
@@ -65,15 +65,7 @@
         "ms-python.vscode-pylance"
       ],
       "settings": {
-        "python.defaultInterpreterPath": "/usr/local/python/current/bin/python",
-        "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
```
samruddhikhandale, line 69: https://github.com/devcontainers/features/blob/2c069d33e7caa3d3dff0a44d0a3c7d164fd02ba8/src/python/install.sh#L29

I believe we are installing utils ^^, so that the Python extension can make use of it. @luabud wondering, are there any replacement extensions that we could use for these deprecated settings? 

I agree we should clean the deprecated settings, trying to understand the impact here. As most of them are disabled by default, probably the impact shouldn't be breaking? 

@bamurtaugh would appreciate your thoughts here 👀

samruddhikhandale (commented):
Thank you for taking the time to contribute this PR, we appreciate it! ✨

src/python/devcontainer-feature.json
```diff
@@ -65,15 +65,7 @@
...
-        "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
```
samruddhikhandale, line 70: Looks like we'd need similar changes to the images repo https://github.com/search?q=repo%3Adevcontainers%2Fimages+python.formatting+python.linting&type=code

src/python/devcontainer-feature.json
luabud: Sorry, I just noticed I sent the wrong link in the opening comment, apologies for that. This is the fixed one: https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions

> As most of them are disabled by default, probably the impact shouldn't be breaking?
I agree!

There are extensions we could add to replace a few of them! I'm just not sure if it's valuable adding them all given once they're enabled, they all start working immediately, which could be an overwhelming experience (especially when it comes to the linters). I guess one important thing I forgot to note is that these extensions ship a version of the tools alongside them, so it's not required to install the linters or formatters in the environment as they come out of the box. 

In any case, here's the extensions ids of the extensions we have:

- [ms-python.pylint](https://marketplace.visualstudio.com/items?itemName=ms-python.pylint)
- [ms-python.flake8](https://marketplace.visualstudio.com/items?itemName=ms-python.flake8)
- [ms-python.mypy-type-checker](https://marketplace.visualstudio.com/items?itemName=ms-python.mypy-type-checker)
- [ms-python.black-formatte](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter)
- [ms-python.autopep8](https://marketplace.visualstudio.com/items?itemName=ms-python.autopep8)
- [ms-python.isort](https://marketplace.visualstudio.com/items?itemName=ms-python.isort)

If we decide to include them, we might want to set the "fromEnvironment" setting for each extension, and then the "path" setting to point to the tools that are installed in the container. This way the extensions won't use the version of the tools that are shipped, and will instead use the ones installed in the container 😊

(we're deprecating support for yapf, pydocstyle, pycodestyle and bandit giving they have way too low usage to justify the maintenance cost, so unfortunately there's not much of a workaround for that).

src/python/devcontainer-feature.json
luabud: yes, good point! I'll send a PR there too 😊

src/python/devcontainer-feature.json
bamurtaugh: > I believe we are installing utils ^^, so that the Python extension can make use of it.

Maybe this is a silly question, but if we were installing these utils for the Python extension and the Python extension has now deprecated some or all of them, would we want to just no longer install them then? Or is the issue that these utils were being passed to the Python extension but can also be used independent of it, so we'll still need to keep them?

samruddhikhandale: I had the same question, even I am not sure if we only install these utils for the extension. Let's say we did, then would it be possible that folks started using it for other purpose and started relying on its installation? If not, then we could clean the utils in this PR as well. 

@joshspicer wondering if you might know or have any thoughts on this? 

> In any case, here's the extensions ids of the extensions we have:

@luabud thank you for sharing this information. @bamurtaugh what do you think about this one? Should we simply remove the deprecated settings or should we replace them? I agree, adding so many extensions might feel too overwhelming 😅  So probably removing the settings and reiterating based on feedback might make more sense 🤔

bamurtaugh: > So probably removing the settings and reiterating based on feedback might make more sense 🤔

I think this is a great idea! 👍 if they're no longer supported in the Python extension, I think it's good evidence we can try removing them and see if folks start sharing any feedback/issues.

luabud: I agree with everything! Happy to then clean up the utils in this PR as well if you'd like 😊

luabud:
Took the liberty to remove the linters and formatters from DEFAULT_UTILS, but happy to revert if it not appropriate! 😊

luabud:
@samruddhikhandale friendly ping 😊

src/python/devcontainer-feature.json
samruddhikhandale: > I had the same question, even I am not sure if we only install these utils for the extension.

I am still not confident if the extensions are used by folks for any other purpose. If they are, then we shouldn't risk breaking their dev config. We could clean up the utils in a PR which bumps the major version. Hence, I think we should hold on for now? @bamurtaugh what do you think?

That said, and if Brigit agrees, @luabud can you help revert https://github.com/devcontainers/features/pull/625/commits/57c7bec14dca54fde4b4df2580ed429484180882 ? Thanks!

src/python/devcontainer-feature.json
```diff
@@ -65,15 +65,7 @@
...
-        "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
-        "python.linting.flake8Enabled": false,
-        "python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
-        "python.linting.mypyEnabled": false,
-        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint",
-        "python.linting.pylintEnabled": false
+        "python.defaultInterpreterPath": "/usr/local/python/current/bin/python"
```
samruddhikhandale, line 68: Can we bump the minor version? Thanks! 

https://github.com/devcontainers/features/blob/57c7bec14dca54fde4b4df2580ed429484180882/src/python/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Thanks for the ping, apologies to have missed this!
Left some comments.

luabud:
Last commit reverted 😊

luabud:
@samruddhikhandale friendly ping!

luabud:
@samruddhikhandale oh  I'm so sorry! I had totally misunderstood that this would be done in a follow up clean up PR (which I now realize was about bumping the major version, instead of minor 🤦‍♀️)

 Just bumped the version to it to 1.2.1 :)  thanks a lot!

samruddhikhandale (approved):
Looks great, thank you so much! 🎉

momohnasirumuri (approved):

Issue #464, opened 2023-02-27
[dotnet] Simply script to internally use `https://dot.net/v1/dotnet-install.sh`


joshspicer:
The existing `dotnet-install.sh` script maintained by the dotnet team is really reliable and works well - we should refactor the dotnet script to simply use this script in all cases for simplicity

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script

This should also fix the goal of: ref: https://github.com/devcontainers/features/pull/297

Issue #389, opened 2023-01-09
Multiple .NET SDK versions?


justinyoo:
Hi, Team.

I was wondering if it's possible to install multiple .NET SDKs in one devcontainer settings. For example, I'd like to install both .NET 7 and .NET 6 in one devcontainer environment.

What I tried was:

1. In `Dockerfile`, select the base container image from `mcr.microsoft.com/dotnet/sdk:7.0`
2. In `devcontainer.json`, run a shell script, say `post-create.sh`, through `postCreateCommand`.
3. In `post-create.sh`, run the `dotnet-install.sh` script by following [this document](https://learn.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual#scripted-install) to install .NET 6.

However, it didn't go well as intended. Would you be able to provide some guidance to install multiple .NET SDK versions - STS and LTS?

Cheers,

samruddhikhandale:
**Option 1** - The [mcr.microsoft.com/devcontainers/universal:2](https://github.com/devcontainers/images/tree/main/src/universal) consists of .NET 7 & 6. Maybe you could use that as the base container image? 

**Option 2** - We have an experimental `additionalVersions` option for some of the Features to install more than one SDKs. The value specified for `version` option is set as default. 

Can you try adding the following to `.devcontainer.json`? 

```jsonc
"features": {
        "ghcr.io/devcontainers/features/dotnet:1": {
            "version": "7",
            "additionalVersions": "6"
        }
    }
```

justinyoo:
Oh, I might try the **Option 2**, then. Thanks!

justinyoo:
By the way, [devcontainer-feature.json](https://github.com/devcontainers/features/blob/main/src/dotnet/devcontainer-feature.json) doesn't define the `additionalVersions` option, but `install.sh` defines it. Have you noticed that?

samruddhikhandale:
> Option 2 - We have an experimental additionalVersions option for some of the Features to install more than one SDKs. The value specified for version option is set as default.

@justinyoo Yep, we had added it as an experimental option and hence, didn't document it. As we can see a request for it, we'd think of either formalizing that option or finding another better approach for this scenario.

kosalanuwan:
> > Option 2 - We have an experimental additionalVersions option for some of the Features to install more than one SDKs. The value specified for version option is set as default.
> 
[...]

PR #628, opened 2023-07-22
Use dotnet-install.sh in .NET feature

sliekens:
## Enhancements to .NET Feature and improved configuration options

This pull request builds upon the exceptional work done by @jungaretti in #467. It contains significant changes to the .NET feature to ensure better compatibility with multiple .NET versions (#389). Additionally, the installation process has been streamlined by exclusively utilizing [dotnet-install.sh](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) and discontinuing the support for installing .NET via apt. This change aims to simplify the installation process and remove the complexities associated with hybrid installations.

### Major changes:

1. **New configuration options:**
   - I've introduced the ability to specify a release in the form 'X.Y' or 'X.Y.Zxx'.
   - I've replaced `runtimeOnly` with `dotnetRuntimeVersions` and `aspNetCoreRuntimeVersions`
   - I've removed the option to specify a release in the form 'X' to align with the supported configurations of `dotnet-install.sh`.

2. **Valid feature configurations:**
   With these changes, the following `version` and `additionalVersions` configurations are now considered valid:
   - `latest` (default)
   - `lts` 
   - `none`
   - `3.0`
   - `3.1`
   - `6.0.3xx`
   - `6.0.413`
   - `8.0.100-preview.7.23376.3`

   Additionally you may request `dotnetRuntimeVersions` and `aspNetCoreRuntimeVersions`, which use a slightly different version scheme:

   - `latest`
   - `lts` 
   - `3.0`
   - `3.1`
   - `6.0.21`
   - `8.0.0-preview.7.23375.6`

   For a runtime-only installation, you can set the `version` to `"none"` to skip installing the SDK.

``` json
"features": {
    "ghcr.io/devcontainers/features/dotnet:2": {
        "version": "none",
        "dotnetRuntimeVersions": "latest, lts, 3.1",
        "aspnetCoreRuntimeVersions": "latest, lts, 3.1"
    }
}
```

### Reasoning behind changes:

1. **Removing ability to specify only major version ('X'):**
   I decided to remove the ability to specify only a major version ('X') since `dotnet-install.sh` does not directly support this configuration. Previously, the feature would fallback to the _oldest_ minor version ('X.0'), which was counterintuitive. To avoid confusion and maintain clarity, I chose to remove this option and look forward to a resolution of this behavior in the dotnet-install script itself.

2. **New configuration options for runtime-only installations:**
   I've replaced the `runtimeOnly` option with separate options for _dotnet_ and _aspnetcore_ runtime version for several reasons:
   - The previous setup didn't allow the installation of a mix of SDKs and Runtimes using `additionalVersions`, as users could either get all SDKs or all Runtimes.
   - The lack of an option to specify which runtime to install (dotnet or aspnetcore) limited user flexibility.

### Testing improvements:

For comprehensive testing, I've included additional test projects, each targeting a different framework version. These test scenarios progressively utilize newer language features, such as top-level statements, implicit usings, and raw strings, in each version. The tests ensure that the example projects compile and run smoothly under the _vscode_ remote user (except `test.sh`, which runs as root).

I believe these changes will significantly enhance the user experience and improve the overall stability and compatibility of the .NET feature. I welcome feedback and suggestions to continue refining and enhancing this crucial functionality. Thank you for your attention and support in reviewing this PR.

Closes https://github.com/devcontainers/features/issues/464
Fixes https://github.com/devcontainers/features/issues/293
Fixes https://github.com/devcontainers/features/issues/389


sliekens:

sliekens:
Note about using the `"latest"` configuration option: this will run  `dotnet-install.sh --version latest --channel STS` which is not the same as the current latest. When .NET 8 is released, the LTS channel will contain the current latest, until .NET 9 is released.

Another small note: it's not currently possible to pass a `quality` such as `'preview'` or `'daily'` when you specify a two-part version like `'8.0'`. The default quality `'GA'` is used. You can only get preview/daily builds by specifying the exact full version.

sliekens:
@jungaretti  I realized it's not clear from the changes that I used your script as a starting point, you can see the changes more clearly here https://github.com/devcontainers/features/compare/jungaretti/dotnet-2..sliekens:dotnet-install-script?diff=unified#diff-3dff5aa292fb61b47ab35a7440d0770b7188768ac421213739525e3aca615cdc

jungaretti:
> My thinking is rather than verifying the SHA256 digest, wouldn't it be better to include a copy of the script in the feature? That way, upstream changes won't break the feature.

Yes, this makes sense to me. Either way, we need to update the script to respond to upstream changes: updating the hash or patching the copied script. Patching the copied script is much less disruptive!

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
-            DOTNET_DOWNLOAD_HASH="$(echo "${dotnet_download_details}" | jq -r '.hash')"
-            DOTNET_DOWNLOAD_NAME="$(echo "${dotnet_download_details}" | jq -r '.name')"
-        else
-            err "Unable to find .NET binary for version ${DOTNET_DOWNLOAD_VERSION}"
-            exit 1
-        fi
+fetch_latest_sdk_version() {
+    local latest_STS=$(fetch_latest_sdk_version_in_channel "STS")
+    local latest_LTS=$(fetch_latest_sdk_version_in_channel "LTS")
+    if [[ "$latest_STS" > "$latest_LTS" ]]; then
```
samruddhikhandale, line 41: From https://github.com/devcontainers/features/pull/628#issuecomment-1646547165
> Running dotnet-install.sh without any arguments will always give you the latest LTS release. 

The default option of this Feature is `latest`, doesn't it mean it could either give LTS or STS depending on whichever is most recent? eg. The `latest` should provide `.NET 7` (which is STS) 


https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#cadence
image

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
+
+    echo "Installing version '$version'..."
+
+    # Quick options reminder for dotnet-install.sh:
+    #
+    # --version: 'latest' (default) or an exact version in the form 'A.B.C' like '6.0.412'
+    # --channel: 'LTS' (default), 'STS', a two-part version in the form 'A.B' like '6.0' or three-part form 'A.B.Cxx' like '6.0.1xx'
+    # --quality: 'daily', 'signed', 'validated', 'preview' or 'GA'
+    #
+    # Valid examples
```
samruddhikhandale, line 61: Awesome, this is so helpful. Thanks for adding these comments! 🎉

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
-APT_PREFERENCES_FILE_ADDED="false"
-APT_PREFERENCES_UPDATED="false"
+DOTNET_INSTALL_SCRIPT_URL='https://dot.net/v1/dotnet-install.sh'
+DOTNET_INSTALL_SCRIPT='/tmp/dotnet-install.sh'
+DOTNET_INSTALL_DIR='/usr/share/dotnet'
 
 set -e
 
-# Clean up
-rm -rf /var/lib/apt/lists/*
```
samruddhikhandale, line 49: Can we keep this?

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
-    tar -xzf "${TMP_DIR}/${DOTNET_DOWNLOAD_NAME}" -C "${DOTNET_INSTALL_PATH}" --strip-components=1
-
-    CURRENT_DIR="${TARGET_DOTNET_ROOT}/current"
-    if [[ ! -d "${CURRENT_DIR}" ]]; then
-        ln -s "${DOTNET_INSTALL_PATH}" "${CURRENT_DIR}" 
+        # Assume version is an exact version string like '6.0.412' or '8.0.100-rc.1.23371.5''
+        channel=""
     fi
 
-    # Give write permissions to the user.
```
samruddhikhandale, line 389: Won't we still need this?

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
-if [ "${APT_PREFERENCES_UPDATED}" = "true" ]; then
-    if [ "${APT_PREFERENCES_FILE_ADDED}" = "true" ]; then
-        rm -f /etc/apt/preferences
-    else
-        head -n -3 /etc/apt/preferences > /tmp/preferences-temp && mv /tmp/preferences-temp /etc/apt/preferences
-    fi
+    IFS=$OLD_IFS
 fi
 
-rm -rf /var/lib/apt/lists/*
```
samruddhikhandale, line 490: Same here, it would be nice to clean up later. Can we keep this ?

src/dotnet/install.sh
samruddhikhandale: I believe we would still need the code for updating user & installed dir permissions.

test/dotnet/scenarios.json
```diff
@@ -1,83 +1,50 @@
...
+        "image": "mcr.microsoft.com/devcontainers/base:ubuntu-20.04",
+        "remoteUser": "vscode",
         "features": {
             "dotnet": {
-                "version": "6"
+                "version": "3.1"
             }
         }
     },
-    "install_dotnet_latest": {
```
samruddhikhandale, line 59: Can we still keep the scenario which validates the `latest` option? I know that the test would fail when a new latest is available, but should be fine.

src/dotnet/install.sh
samruddhikhandale: I think the code is correctly implemented ^, thanks!  👍

samruddhikhandale (commented):
Hi 👋 

Thank you so much for taking the time to contribute this spectacular change. We really appreciate it! Good job @sliekens @jungaretti 🎉 

Left some thoughts/comments, let me know what you think or how I could help.

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
+        # When user input is form 'A.B' like '3.1'
+        # Then version=latest, channel=3.1
+        channel="$version"
+        version="latest"
+    elif [[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]xx$ ]]; then
+        # When user input is form 'A.B.Cxx' like '6.0.4xx'
+        # Then version=latest, channel=6.0.4xx
+        channel="$version"
+        version="latest"
     else
```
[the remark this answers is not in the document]
sliekens: 👌 I added the validation in https://github.com/devcontainers/features/pull/628/commits/e2ce4631647b2b1a459779413e07e5aad2c8c928. I made it so the additional versions are also validated before any real work is carried out.

src/dotnet/install.sh
sliekens: Yep, makes sense to me.

src/dotnet/install.sh
sliekens: Yep, makes sense to me.

test/dotnet/scenarios.json
sliekens: There is the autogenerated test case which defaults to `latest` and this is verified in `test.sh` in a non-breaking way by comparing it to `latest.version` files from the AzureCDN. Will that suffice?

samruddhikhandale: That was a smart thing to do! ✨ 

Based on https://github.com/devcontainers/features/pull/628#discussion_r1276647264, I am looking forward to validate the `fetch_latest_sdk_version()` function. Hence, validating the `latest` installed version with `7` will provide us an assurance that the function works as expected.

Hence, once .NET 8 is released, I expect the test to fail.

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,31 @@
...
+        "additionalVersions": {
+            "type": "string",
+            "default": "",
+            "description": "Comma-separated list of additional .NET SDK versions to install. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' for a specific release, or 'X.Y.Z' for a specific version."
         }
     },
     "containerEnv": {
-        "DOTNET_ROOT": "/usr/local/dotnet/current",
-        "PATH": "/usr/local/dotnet/current:${PATH}"
+        "DOTNET_ROOT": "/usr/share/dotnet",
```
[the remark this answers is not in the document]
sliekens: This change comes from a want to run a GitHub action inside a prebuilt Dev Container (https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container)

I noticed that the `setup-dotnet` GitHub action installs to `/usr/share/dotnet` by default
https://github.com/actions/setup-dotnet#environment-variables

If you want to use `setup-dotnet` in a Dev Container that already has dotnet installed in `/usr/local`, then you have to override the `DOTNET_INSTALL_DIR` env var of the action. Otherwise the existing installation will not be detected.

I recommend aligning the installation directory of the .NET feature with the default installation directory of the `setup-dotnet` action so they work together without requiring configuration of the `DOTNET_INSTALL_DIR` env var.

src/dotnet/install.sh
```diff
@@ -4,489 +4,133 @@
...
-Package: *
-Pin: origin "packages.microsoft.com"
-Pin-Priority: 1001
-EOF
-)"
+DOTNET_ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
-APT_PREFERENCES_FILE_ADDED="false"
-APT_PREFERENCES_UPDATED="false"
+DOTNET_INSTALL_SCRIPT_URL='https://dot.net/v1/dotnet-install.sh'
```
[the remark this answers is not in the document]
sliekens: I've never done this but I imagine a GitHub workflow could fetch the latest script and create a pull request if there are any changes?

src/dotnet/install.sh
sliekens: This question I don't know much about. All I can say is the scenario tests run under the `vscode` remote user and they pass. I think maybe the `dotnet-install.sh` script ensures correct permissions?

test/dotnet/scenarios.json
sliekens: So I should add an assertion that the installed SDK is a flavor of 7.0, expecting the test will break in November?

src/dotnet/install.sh
```diff
@@ -104,33 +112,57 @@ install_version() {
...
+    echo $1 | tr -d '[:space:]'
+}
+
 if [ "$(id -u)" -ne 0 ]; then
-    echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
+    err 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
 fi
 
+# For our own convenience, combine DOTNET_VERSION and DOTNET_ADDITIONAL_VERSIONS into a single 'versions' array
```
samruddhikhandale, line 134: 🎉 Nice!

test/dotnet/scenarios.json
samruddhikhandale: Yes, that would be great!

src/dotnet/install.sh
samruddhikhandale: > I think maybe the dotnet-install.sh script ensures correct permissions?

I am not too confident about that, but doesn't seem like the script does that. 👇 

image

In the past, we intentionally added privs - set user permissions, created specific gids/groups for each Feature and added sticky bits as there were permissions issues. I believe, we still need this. // cc @Chuxel @bamurtaugh

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
samruddhikhandale: Hmm, thanks for the explanation. The logic makes sense to me. 
I'll need to dig around for the actual reason of preferring this PATH. (@Chuxel wondering if you might know!) 

Note: We'd need to update https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/local-features/setup-user/install.sh#L73-L78 post this PR is merged.

src/dotnet/install.sh
sliekens: Do you recall what permission issues? I'm hesitant to potentially weaken security when currently all tests pass. I think `rwx` for root and `r-x` for non-root users is already an ideal situation.

test/dotnet/scenarios.json
sliekens: Done in dc7a0beda42c078afcb078a1e57612885558d77a

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,31 @@
...
+                "lts",
+                "8.0",
+                "7.0",
+                "6.0"
             ],
             "default": "latest",
-            "description": "Select or enter a dotnet CLI version. (Available versions may vary by Linux distribution.)"
+            "description": "Version of .NET SDK to install. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' for a specific release, or 'X.Y.Z' for a specific version."
         },
-        "runtimeOnly": {
```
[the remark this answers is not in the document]
sliekens: My 2¢ is that it would be better to offer the runtimes as separate features, as that would solve my concerns with offering it as a `runtimeOnly` boolean option.

- Separate features allow users to install a **precise** mix of SDKs and Runtimes
- It will be easier to handle differences between the SDK versioning and Runtime versioning
    - Eliminate the need for `if [ "${DOTNET_RUNTIME_ONLY}" = "true" ]; then ...` conditional code
    - [How .NET is versioned](https://learn.microsoft.com/en-us/dotnet/core/versions/)


``` json
{
    "features": {
        "ghcr.io/devcontainers/features/dotnet:2": { "version": "latest", "additionalVersions": [] },
        "ghcr.io/devcontainers/features/dotnet-runtime:1": { "version": "latest", "additionalVersions": [] },
        "ghcr.io/devcontainers/features/aspnetcore-runtime:1": { "version": "latest", "additionalVersions": [] },
    }
}
```

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,31 @@
...
+            "type": "string",
+            "default": "",
+            "description": "Comma-separated list of additional .NET SDK versions to install. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
         }
     },
     "containerEnv": {
-        "DOTNET_ROOT": "/usr/local/dotnet/current",
-        "PATH": "/usr/local/dotnet/current:${PATH}"
+        "DOTNET_ROOT": "/usr/share/dotnet",
+        "PATH": "${DOTNET_ROOT}:${PATH}"
```
sliekens, line 28: Specifying PATH here might be unnecessary, the install script already does this unless `--no-path` is used.

I am not sure if it is better to remove this config, or keep it and run the script with `--no-path`

image

sliekens: Removing PATH from containerEnv breaks the test 🤔

```
dotnet_helpers.sh: line 28: dotnet: command not found
```

sliekens: Oh because the script only exports it for the current session... so I guess better to use `--no-path`

Note: also need to add `${DOTNET_ROOT}/tools` to PATH for global tools.

sliekens: Fixed in 2d0fb73697461a28b2628b69962263043d6873dd

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: If this is too problematic, I think it is also fine to keep the old directory and add `DOTNET_INSTALL_DIR` to the `containerEnv`.

``` json
"containerEnv": {
    "DOTNET_ROOT": "/usr/local/dotnet/current",
    "DOTNET_INSTALL_DIR": "$DOTNET_ROOT",
    "PATH": "$PATH:$DOTNET_ROOT:~/.dotnet/tools"
}
```

This might be the best way to ensure maximum compatibility, although I find the `current` directory a bit weird and unnecessary beyond maintaining compatibility.

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,31 @@
...
-            "type": "boolean",
-            "default": true,
-            "description": "If true, it installs using apt instead of the release URL"
+        "additionalVersions": {
+            "type": "string",
+            "default": "",
+            "description": "Comma-separated list of additional .NET SDK versions to install. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
         }
     },
     "containerEnv": {
```
sliekens, line 26: DOTNET_USE_POLLING_FILE_WATCHER should be enabled for Docker mounted volumes, otherwise `dotnet watch` won't work.

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: Also check out the docs for DOTNET_ROOT

> Specifies the location of the .NET runtimes, if they are not installed in the default location. The default location on Windows is C:\Program Files\dotnet. The default location on macOS is /usr/local/share/dotnet. The default location on Linux varies depending on distro and installment method. **The default location on Ubuntu 22.04 is /usr/share/dotnet (when installed from packages.microsoft.com) or /usr/lib/dotnet (when installed from Jammy feed)**. 

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_root-dotnet_rootx86

jungaretti:
Closes https://github.com/devcontainers/features/issues/464
Fixes https://github.com/devcontainers/features/issues/293
Fixes https://github.com/devcontainers/features/issues/389

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
jungaretti: I agree with @sliekens that the runtime is best suited for its own feature. I recognize that both features would use this `dotnet-install.sh` script and this overall structure, but the runtime is certainly a distinct tool than the SDK.

This reminds me of the situation with our Python and Jupyter features @samruddhikhandale. We discontinued the Jupyter feature in favor of an option on the Python feature. In hindsight, I think most of us agree that was a confusing decision.

src/dotnet/install.sh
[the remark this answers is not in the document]
jungaretti: Can you add a simple script that downloads the latest version? For example, `src/dotnet/scripts/fetch-latest-dotnet-install.sh`

```bash
pushd src/dotnet/scripts
wget https://dot.net/v1/dotnet-install.sh -O dotnet-install.sh
```

The `scripts` directory can hold `dotnet-install.sh` and `fetch-latest-dotnet-install.sh`

We can run it manually for now, or add a simple automation in the future.

src/dotnet/install.sh
jungaretti: Relevant issue regarding permissions: https://github.com/devcontainers/features/issues/293

The bug reveals itself when you run `dotnet restore`. @sliekens Do any of your tests use `dotnet restore`? Seems like no. Could we add a project with a simple dependency to make sure that works? Here's an example from my original PR: https://github.com/devcontainers/features/pull/467/files#diff-5d8e34cd2b389901f42c076a13c3f622cec4aa490c1eb3c07acb32a324343527

sliekens: @jungaretti there is an implicit restore behind `dotnet run` in the tests. Indeed, I borrowed your example project with the Newtonsoft.Json dependency. Executing `dotnet restore` before `dotnet run` is not strictly needed unless there is a specific reason.

src/dotnet/install.sh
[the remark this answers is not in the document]
sliekens: @jungaretti done in 52008539b5ffcaf8fcee7314d5bf30103a676aa2

sliekens:
@samruddhikhandale @jungaretti  what do you think about adding `DOTNET_RUNNING_IN_CONTAINER(S)` to the `containerEnv`?

image

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables

src/dotnet/install.sh
jungaretti: Rock on, what a legendary PR!

I confirmed that your feature fixes #293. I made a simple project that consumes this feature. You can make a codespace for my branch here: https://github.com/codespaces/new/?skip_quickstart=true&repo=449816349&ref=sliekens-dotnet&devcontainer_path=.devcontainer%2Fdevcontainer.json

samruddhikhandale: Thanks @jungaretti for providing and validating the `dotnet restore` scenario. Also, this bit was added so that a non-root user don't require `sudo` and there were issues with `dotnet nuget` command.

Looping in @bderusha who first contributed this .NET Feature (when it was known as scripts) to [microsoft/vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/dotnet-debian.sh). 

@bderusha would highly appreciate your review on this PR and your thoughts on removing the permissions bit which was added in https://github.com/microsoft/vscode-dev-containers/pull/1210

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
samruddhikhandale: >  "DOTNET_ROOT": "/usr/share/dotnet",

Discussed this with the team, as `"/usr/share/dotnet"` syncs with GitHub Actions, and also with the [official dotnet image](https://hub.docker.com/_/microsoft-dotnet), it would be great to sync the PATH. Thanks for the change! 

As PATH changes can be highly breaking, we do need to signal it with major version bump. As this PR already does that, it seems fine. 👍

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
samruddhikhandale: Discussed this with the team, historically `runtimeOnly` option is used by folks who simply need to run .NET tools/programs and who uses other programming languages for development. Installing the runtime can save some disk space compared to installing the SDKs. 

I like the idea of isolating it to a new Feature, however, if `runtimeOnly` Feature opiton's usage is pretty low, not sure if we should invest time in creating or maintaining this new Feature. 

Looping in @bderusha who originally added this Feature option, any thoughts about removing this option from the the Feature? or would appreciate your insights on it's usage/use cases. Thank you!

samruddhikhandale:
> @samruddhikhandale @jungaretti what do you think about adding `DOTNET_RUNNING_IN_CONTAINER(S)` to the `containerEnv`?
> 
> image
> https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables

Adding `DOTNET_RUNNING_IN_CONTAINER=true` seems fine, I can see it set in [.NET sdk images](https://hub.docker.com/_/microsoft-dotnet-sdk/). Would be nice to sync as much as possible 👍 

image

@sliekens Out of curiosity, how/who uses this env variable? From a quick search, looks like this only signals a developer that .NET is running in the container and doesn't have any other use case (could be wrong)?

sliekens:
@samruddhikhandale I found an example where it is used to redirect build output to a different folder based on whether `dotnet build` was executed on the host or inside the container.

https://github.com/dotnet/dotnet-docker/blob/main/samples/Directory.Build.props

The reason for this is stretching the limits of my MSBuild knowledge, but, I think this configuration makes it so you can switch between running `dotnet build` on the host or inside the container without having to do `dotnet clean` in between builds.

Other than that, I think perhaps .NET itself does some things differently when this environment variable is set.

There is an open issue https://github.com/dotnet/dotnet-docker/issues/3997 to document the purpose of DOTNET_RUNNING_IN_CONTAINER and I think currently @richlander is the best person to answer this question.

sliekens:
@samruddhikhandale maybe a bit nitpicky but I would like to be able to use JSON booleans in `containerEnv`, validation is failing because only strings are allowed.

``` json
{
    "containerEnv": {
        "DOTNET_RUNNING_IN_CONTAINER": true
    }
}
```

image

samruddhikhandale:
> @samruddhikhandale maybe a bit nitpicky but I would like to be able to use JSON booleans in `containerEnv`, validation is failing because only strings are allowed.

Thanks for the suggestion, opened https://github.com/devcontainers/spec/issues/286 for tracking.

sliekens:
Thanks!

> @sliekens Out of curiosity, how/who uses this env variable?

I found it's used by ASP.NET Core DataProtection to log a warning when running in a container without a persistent volume for the generated keys.

https://github.com/dotnet/aspnetcore/blob/125033fd2f7a46ffe41e9e52af911c7027d310d4/src/DataProtection/DataProtection/src/Internal/ContainerUtils.cs#L86

https://github.com/dotnet/aspnetcore/blob/125033fd2f7a46ffe41e9e52af911c7027d310d4/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs#L36-L40

It's also used for some telemetry stuff in https://github.com/dotnet/interactive and https://github.com/dotnet/upgrade-assistant. That's about all I could find from Microsoft.

richlander:
> DOTNET_RUNNING_IN_CONTAINER

I think you've reverse-engineered the intent.

- Simple indicator that code is running in containers. We tried but could not identify a reliable way to do this ourselves, algorithmically.
- Using the value to enable volume mounting scenarios were container and local builds might stomp each other. That's what the D.B.p is doing.

I think of the ENV as nice-to-have. If you add it, cool. If you don't add it, also fine.

sliekens:
Thanks for all the feedback so far. How can I regenerate the README from the updated devcontainer-feature.json, @samruddhikhandale?

test/dotnet/scenarios.json
```diff
@@ -1,84 +1,58 @@
...
-    "install_dotnet_lts": {
-        "image": "ubuntu:focal",
+    "install_dotnet_multiple_versions": {
+        "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+        "remoteUser": "vscode",
         "features": {
             "dotnet": {
-                "version": "lts"
+                "version": "8.0.100-preview.6.23330.14",
+                "additionalVersions": [
```
sliekens, line 44: How does this test pass when `additionalVersions` should have been a comma-separated string instead of an array?

sliekens: Cool, there is an automatic conversion somewhere from array to comma-separated strings.

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,33 @@
...
-            "type": "boolean",
-            "default": false,
-            "description": "Install just the dotnet runtime if true, and sdk if false."
-        },
-        "installUsingApt": {
-            "type": "boolean",
-            "default": true,
-            "description": "If true, it installs using apt instead of the release URL"
+        "additionalVersions": {
+            "type": "string",
```
sliekens, line 21: @samruddhikhandale is it possible to change this typing to suggest an array instead of a string? I discovered there is an automatic conversion from array to comma-separated strings.

This works
``` json
{
  "additionalVersions": "6.0, 5.0, 3.1"
}
```

But this also works and feels more intuitive.
``` json
{
  "additionalVersions": ["6.0", "5.0", "3.1"]
}
```

sliekens: Seems it's currently not officially supported by the spec

image

https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-features.md#the-options-property

The fact that an array _does_ work might be accidental?

samruddhikhandale:
> Thanks for all the feedback so far. How can I regenerate the README from the updated devcontainer-feature.json, @samruddhikhandale?

@sliekens Once this PR is merged, https://github.com/devcontainers/action will run automatically ; which opens a PR that  updates the README file based on `devcontainer-feature.json` and `NOTES.md`.

src/dotnet/devcontainer-feature.json
samruddhikhandale: > But this also works and feels more intuitive.
``` json
{
  "additionalVersions": ["6.0", "5.0", "3.1"]
}
```

I do like the idea of having the `value` as an `array` and it's good to know that it works out of the box.

> Seems it's currently not officially supported by the spec

I am not confident as of why we currently don't support `array` as a type. However, one thing I can think of --> It would definitely need some tweaking to the Dev Containers extension for the `Add dev container Configuration files...` command (for accepting values in the form of an array). See 👇 

image

@joshspicer any thoughts?

src/dotnet/install.sh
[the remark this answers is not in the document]
samruddhikhandale: 💡 @sliekens Would you be interested in adding the workflow file? It would be very similar to https://github.com/devcontainers/features/blob/main/.github/workflows/update-documentation.yml

Instead of the `devcontainers/action`, you could use the commands from [fetch-latest-dotnet-install.sh](https://github.com/devcontainers/features/blob/e39d87bb202a1704e5d15377be14804479e2aebc/src/dotnet/scripts/fetch-latest-dotnet-install.sh) to open a PR (if there are any updates to `dotnet-install.sh`). I can provide help with the changes!

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,33 @@
...
+            "default": "",
+            "description": "Comma-separated list of additional .NET SDK versions to install. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
         }
     },
     "containerEnv": {
-        "DOTNET_ROOT": "/usr/local/dotnet/current",
-        "PATH": "/usr/local/dotnet/current:${PATH}"
+        "DOTNET_ROOT": "/usr/share/dotnet",
+        "PATH": "$PATH:$DOTNET_ROOT:~/.dotnet/tools",
```
samruddhikhandale, line 28: > ~/.dotnet/tools

From https://github.com/devcontainers/cli/issues/391, I don't expect ^ to work? 🤔

src/dotnet/install.sh
```diff
@@ -4,44 +4,13 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 #
-# Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/dotnet.md
+# Docs: https://github.com/devcontainers/features/tree/main/src/dotnet
 # Maintainer: The VS Code and Codespaces Teams
```
samruddhikhandale, line 8:
```suggestion
# Maintainer: The Dev Container spec maintainers
```

src/dotnet/devcontainer-feature.json
jungaretti: Where's the automatic conversion taking place? That's interesting

src/dotnet/devcontainer-feature.json
sliekens: Hmm 🙃, what did you mean by "that doesn't work if a shell is not invoked."?

I wrote the following test to make sure running a tool from this directory works in bash.

https://github.com/devcontainers/features/blob/3d48f7749e8e74477d26e23dc9a6e4f4f8ae77e3/test/dotnet/install_dotnet_global_tool.sh#L16-L25

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
bderusha: I can't speak to how the maintainers should weigh the user value VS maintenance cost of adding dotnetcore and ASPcore runtime features, which I agree is the best user experience option.

Keeping the old boolean is not ideal, but it is not the only alternative. IMHO it would be straight forward to offer support for a version prefix like `dotnetcore#3.1`.  Split on the special character, normalize it, and pass the value to via the `--runtime` parameter and let the install script manage the rest. This might strike a better balance between maintenance and feature support.

It would be great if you could pass the raw prefix to `--runtime`, but the magic values would need to be `aspnetcore` and `dotnet`. The latter being not descriptive enough for me, personally.

src/dotnet/install.sh
```diff
@@ -53,440 +22,145 @@ err() {
...
-    dotnet_channel_version="$(curl -s "${DOTNET_CDN_FEED_URI}/dotnet/release-metadata/releases-index.json" | jq -r --arg channel_version "${dotnet_patchless_version}" '[."releases-index"[]] | sort_by(."channel-version") | reverse | map( select(."channel-version" | startswith($channel_version))) | first | ."channel-version"')"
-    set -e
-
-    # Construct the releases URL using the official channel-version if one was found.  Otherwise make a best-effort using the user input.
-    if [ -n "${dotnet_channel_version}" ] && [ "${dotnet_channel_version}" != "null" ]; then
-        dotnet_releases_url="${DOTNET_CDN_FEED_URI}/dotnet/release-metadata/${dotnet_channel_version}/releases.json"
+# Installs a version of .NET using the DOTNET_INSTALLER_SCRIPT
+install_version() {
+    local version="$1"
+    local channel="STS"
```
bderusha, line 58: It seems like this default value is never used. Consider setting it to `channel=""` to make that more clear

src/dotnet/devcontainer-feature.json
jungaretti: Cannot find executable in dotnet/tools

This doesn't seem to work in a codespace. Link to create one: https://github.com/codespaces/new/?skip_quickstart=true&repo=449816349&ref=sliekens-dotnet&devcontainer_path=.devcontainer%2Fdevcontainer.json

src/dotnet/install.sh
bderusha: I contributed this feature a while ago, so forgive the fuzzy memory, but I believe we added this to support running the CLI as non-root without `sudo`.

We also copied the majority of the lines from the [Go install script](https://github.com/devcontainers/features/blob/e39d87bb202a1704e5d15377be14804479e2aebc/src/go/install.sh#L138-L144) so it is certainly possible that some of the lines required for that script made it into ours unnecessarily.

bderusha (commented):
This is great work! I left some comments for your consideration.

src/dotnet/scripts/dotnet-install.sh
```diff
@@ -0,0 +1,1746 @@
+#!/usr/bin/env bash
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
```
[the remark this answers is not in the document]
sliekens: I went ahead and moved the script to a `vendor` directory with a README to alert users not to make changes here, will that suffice? I'd rather not modify the actual script in any way.

src/dotnet/scripts/fetch-latest-dotnet-install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
```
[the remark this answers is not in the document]
sliekens: I wrote this script, but didn't give too much thought to the header. I changed it to that of the other Dev Container script.

``` bash
#-------------------------------------------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
#-------------------------------------------------------------------------------------------------------------
#
# Docs: https://github.com/devcontainers/features/tree/main/src/dotnet
```

src/dotnet/install.sh
```diff
@@ -53,440 +22,145 @@ err() {
...
+    # dotnet-install.sh [--version latest] --channel 8.0 --quality daily
+    # dotnet-install.sh --version 6.0.412
+    #
+    # The channel option is only used when version is 'latest' because an exact version overrides the channel option
+    # The quality option is only used when channel is 'A.B' or 'A.B.Cxx' because it can't be used with STS or LTS
+    #
+    # This script aims to reduce these combinations of options to a single 'version' input
+    # Currently this script does not make it possible to request a version in the form 'A.B' or 'A.B.Cxx' and a quality other than 'GA'
+    if [[ "$version" == "latest" ]]; then
+        # Fetch the latest version manually, because dotnet-install.sh does not support it directly
```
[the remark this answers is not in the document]
sliekens: Yea, so the problem that `fetch_latest_sdk_version` solves is that STS is not by definition the latest version. When .NET 8.0 becomes available, the LTS channel will have the latest version while the STS channel will continue to point at .NET 7.0.

src/dotnet/install.sh
```diff
@@ -53,440 +22,145 @@ err() {
...
-    set +e
-    dotnet_channel_version="$(curl -s "${DOTNET_CDN_FEED_URI}/dotnet/release-metadata/releases-index.json" | jq -r --arg channel_version "${dotnet_patchless_version}" '[."releases-index"[]] | sort_by(."channel-version") | reverse | map( select(."channel-version" | startswith($channel_version))) | first | ."channel-version"')"
-    set -e
-
-    # Construct the releases URL using the official channel-version if one was found.  Otherwise make a best-effort using the user input.
-    if [ -n "${dotnet_channel_version}" ] && [ "${dotnet_channel_version}" != "null" ]; then
-        dotnet_releases_url="${DOTNET_CDN_FEED_URI}/dotnet/release-metadata/${dotnet_channel_version}/releases.json"
+# Installs a version of .NET using the DOTNET_INSTALLER_SCRIPT
+install_version() {
+    local version="$1"
```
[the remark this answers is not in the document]
sliekens: Indeed, this is better

src/dotnet/install.sh
```diff
@@ -53,440 +22,145 @@ err() {
...
+    #
+    # The channel option is only used when version is 'latest' because an exact version overrides the channel option
+    # The quality option is only used when channel is 'A.B' or 'A.B.Cxx' because it can't be used with STS or LTS
+    #
+    # This script aims to reduce these combinations of options to a single 'version' input
+    # Currently this script does not make it possible to request a version in the form 'A.B' or 'A.B.Cxx' and a quality other than 'GA'
+    if [[ "$version" == "latest" ]]; then
+        # Fetch the latest version manually, because dotnet-install.sh does not support it directly
+        version=$(fetch_latest_sdk_version)
+        channel=""
```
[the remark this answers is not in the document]
sliekens: Updated to always set version, then channel.

src/dotnet/install.sh
```diff
@@ -53,440 +22,145 @@ err() {
...
+        # When user input is 'lts'
+        # Then version=latest, channel=LTS
+        channel="LTS"
+        version="latest"
+    elif [[ "$version" =~ ^[0-9]+\.[0-9]+$ ]]; then
+        # When user input is form 'A.B' like '3.1'
+        # Then version=latest, channel=3.1
+        channel="$version"
+        version="latest"
+    elif [[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]xx$ ]]; then
```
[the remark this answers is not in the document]
sliekens: From what I understand, these numbers are incremented with Visual Studio service releases. So for 7.0.4xx you need to update your VS2022 when coming from 7.0.3xx or lower.

I am not sure if this is at all relevant to someone using VS Code in a Linux container, so I am okay with removing this case. I am not entirely sure why the option exists, that's why I included it as a safe default.

src/dotnet/install.sh
sliekens: Very true, fixed.

src/dotnet/scripts/fetch-latest-dotnet-install.sh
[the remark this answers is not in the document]
sliekens: Marking as resolved but let me know if there is still a problem

src/dotnet/install.sh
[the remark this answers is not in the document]
sliekens: Marking as resolved but let me know if there is still a problem

src/dotnet/devcontainer-feature.json
sliekens: I'm not 100% sure why `which` doesn't see it, but...

image

sliekens: Another example

image

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: Or maybe something like this? Naming was never my strong suit but you get the picture.

``` json
{
    "features": {
        "ghcr.io/devcontainers/features/dotnet:2": {
            "version": "latest",
            "additionalVersions": "lts",
            "dotnetRuntimeVersions": "3.1,5.0",
            "aspnetcoreRuntimeVersions": "3.1,5.0"
        }
    }
}
```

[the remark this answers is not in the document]
bderusha: Yes, I like the explicit nature of this rather than relying on documented magic strings. Looking at your example I am wondering if it might be cleanest to just have three options:

```
{
    "features": {
        "ghcr.io/devcontainers/features/dotnet:2": {
            "dotnetSdkVersions": "lts, 5.0",
            "dotnetcoreRuntimeVersions": "3.1",
            "aspnetcoreRuntimeVersions": "latest"
        }
    }
}
```
And signal the most common case by making `dotnetSdkVersions` default to `latest` and the rest default to empty string.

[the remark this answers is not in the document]
bderusha: It's my expectation that your example above would **not** disable installation of the SDK. If `dotnetSdkVersions` defaults to `latest` and it is left unspecified (per your example) the resulting environment variables would be:

```
DOTNETSDKVERSIONS="latest"
DOTNETCOREVERSIONS=""
ASPNETCOREVERSIONS="3.1"
```

A user wanting to install only that single runtime would supply:
```
{
    "features": {
        "ghcr.io/devcontainers/features/dotnet:2": {
             "dotnetSdkVersions": "",
             "aspnetcoreRuntimeVersions": "3.1"
        }
    }
}
```
That feels pretty intuitive so long as that default is documented.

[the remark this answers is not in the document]
sliekens: Is it possible for the install script to distinguish missing config from a configured empty string?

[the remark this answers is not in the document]
bderusha: Check out the docs for how dev containers handles [options resolution](https://containers.dev/implementors/features/#option-resolution-example). TL;DR: The install script doesn't distinguish between empty VS missing, but the tool that parses the feature object and generated the ENV VARs does and will apply the configured defaults for missing options.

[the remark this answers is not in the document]
sliekens: Ah, so actually this code at the top is not good? As it would override a configured empty string.

``` sh
DOTNET_VERSION="${VERSION:-"latest"}"
```

src/dotnet/install.sh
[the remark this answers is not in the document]
sliekens: @samruddhikhandale  I can try but how would I test it? Run it against my fork?

sliekens:
Regarding the C# for Visual Studio Code extension:

Recently it has taken a dependency on [.NET Install Tool for Extension Authors](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime) which now causes a version of .NET Runtime to be downloaded and installed automatically.

image

I find this behavior slightly annoying

- the Output window steals focus
- there is no caching of the installation, the runtime is downloaded each time VSCode is restarted
- you end up with multiple installations of .NET CLI/Runtime
    -  `/home/vscode/.vscode-server/data/User/globalStorage/ms-dotnettools.vscode-dotnet-runtime/.dotnet/7.0.10/dotnet`
    -  `/usr/share/dotnet/dotnet`

It is possible to configure an `existingDotnetPath`, which solves the second and third issues. However I am not confident that this is a good default. I am not sure if this will work well if the user configures an older .NET version.

``` json
{
  "customizations": {
    "vscode": {
      "settings": {
        "dotnetAcquisitionExtension.existingDotnetPath": [
          {
            "extensionId": "ms-dotnettools.csharp",
            "path": "/usr/share/dotnet/dotnet"
          }
        ]
      }
    }
  }
}
````

Also, the output window will still steal focus.

image

File: src/dotnet/devcontainer-feature.json (hunk 15 above)
Comment (sliekens) [reply src/dotnet/devcontainer-feature.json:21]: Good question 🤷‍♂️, I analyzed the call stack starting from here
https://github.com/devcontainers/cli/blob/413c5f790f3a4972bf6f05b71d71ac946c87bc7c/src/spec-node/containerFeatures.ts#L300

I didn't find any explicit conversions or clues that hint at an implicit conversion

File: src/dotnet/devcontainer-feature.json (hunk 11 above)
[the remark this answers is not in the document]
Comment (sliekens) [reply src/dotnet/devcontainer-feature.json:19]: Removed in 2b2e683da219ca23b892c49376b0449aeb38af84

Now I'm waiting for a decision on what to do with runtime

1. Introduce separate features for each runtime
    - offers the best UX
    - ⚠️ bigger maintenance cost
2. Introduce new options to configure runtime versions
    - also acceptable UX
    - reduced maintenance cost compared to separate features
3. Keep the `runtimeOnly` option as it was before
    - ⚠️ no mixing of SDK and Runtimes possible
    - ⚠️ no ability to specify `dotnet` or `aspnetcore` runtime 
    - ⚠️ more complicated input validation code
4. Discontinue runtime only installations
    - the SDK includes all runtimes anyway
    - ⚠️ bigger image size might not be justifiable if you only need the runtime

I think 2) is clearly ahead in the race so I would like to propose this configuration schema.

``` js
{
    "version": "latest",
    "runtime": undefined | 'dotnet' | 'aspnetcore', // undefined for SDK install (default)
    "additionalSdkVersions": "5.0, 6.0, 7.0",
    "additionalDotnetRuntimeVersions": "5.0, 6.0, 7.0",
    "additionalAspNetCoreRuntimeVersions": "5.0, 6.0, 7.0",
};
```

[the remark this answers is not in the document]
bderusha: Is there a reason to prefer the idea of "additional" versions to just using a single list of versions (for each type)?

```
{
    "sdkVersions": "5.0, 6.0, 7.0",  // default to "latest"
    "dotnetRuntimeVersions": "5.0, 6.0, 7.0", // default to ""
    "aspNetCoreRuntimeVersions": "5.0, 6.0, 7.0", // default to ""
};
```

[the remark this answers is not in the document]
sliekens: I'd keep `version` because it is special cased by the CLI to make shorthand configuration possible.

image
https://containers.dev/implementors/features/#devcontainer-json-properties

Special treatment in the CLI:

https://github.com/devcontainers/cli/blob/413c5f790f3a4972bf6f05b71d71ac946c87bc7c/src/spec-configuration/containerFeaturesConfiguration.ts#L1303C3-L1305

src/dotnet/install.sh
sliekens: So what is the verdict? Personally I don't think it's needed anymore. I couldn't find anything that is blocked by permissions, all tests pass with the current permission set. I'm not comfortable putting these lines of code in without a failure scenario.

jungaretti: Agreed. I feel confident we don't need these lines, especially with the codespace repro that proves the original problem is solved.

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
bderusha: Ah yes, keeping that shorthand is nice. Perhaps an approach that could work to keep `version` but eliminate the need for `kind` is to take a similar approach here as is done for java:

That is, allowing a version to be `"none"` (See java example [here](https://github.com/devcontainers/images/blob/main/src/java/.devcontainer/devcontainer.json#L7-L9))

The use case for Java is that the feature allows you to easily add other binaries like Maven and Gradle, but if you are starting from the base Java devcontainer image you don't want the feature to reinstall java. Since we treat `version` as a shorthand for `SDKversion`, `"none"` is a reasonable choice of SDK for anyone wanting to install standalone runtimes.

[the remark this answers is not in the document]
sliekens: Perfect! I think consistency in configuration between features is very good.

Updated proposal for runtime-only configuration:

``` js
{
    "version": "none",
    "dotnetRuntimeVersions": "5.0, 6.0, 7.0",
    "aspNetCoreRuntimeVersions": "5.0, 6.0, 7.0"
}
```

SDK only

``` js
{
    "version": "latest",
    "additionalVersions": "lts"
}
```

Mix

``` js
{
    "version": "latest",
    "additionalVersions": "lts",
    "dotnetRuntimeVersions": "5.0",
    "aspNetCoreRuntimeVersions": "5.0"
}
```

(I'm not sure about empty string for `version` - assume `latest` or `none`?)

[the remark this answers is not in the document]
bderusha: My preference would be to treat empty string the same as `none`, but there's a reasonable argument to be made for either. As long as it's documented I would be supportive of whichever you choose.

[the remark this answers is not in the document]
samruddhikhandale: The proposal in https://github.com/devcontainers/features/pull/628/files#r1296984833 looks amazing! 🎉  Thank you! 

> (I'm not sure about empty string for version - assume latest or none?)

I think `latest` makes the most sense to be in sync with other Features.

src/dotnet/install.sh
```diff
@@ -4,44 +4,10 @@
...
-DOTNET_ARCHIVE_VERSION_CODENAMES="buster bullseye bionic focal hirsute jammy"
-# Feed URI sourced from the official dotnet-install.sh
-# https://github.com/dotnet/install-scripts/blob/1b98b94a6f6d81cc4845eb88e0195fac67caa0a6/src/dotnet-install.sh#L1342-L1343
-DOTNET_CDN_FEED_URI="https://dotnetcli.azureedge.net"
-# Ubuntu 22.04 and on do not ship with libssl1.1, which is required for versions of .NET < 6.0
-DOTNET_VERSION_CODENAMES_REQUIRE_OLDER_LIBSSL_1="buster bullseye bionic focal hirsute"
-
-# Comma-separated list of dotnet versions to be installed
-# alongside DOTNET_VERSION, but not set as default.
-ADDITIONAL_VERSIONS=${ADDITIONALVERSIONS:-""}
```
samruddhikhandale, line 34: Can we add this back? It makes the variable readable.

From earlier conversations, I understand the reason behind this change https://github.com/devcontainers/features/commit/2b2e683da219ca23b892c49376b0449aeb38af84. Can we update it to 👇  instead?

```
ADDITIONAL_VERSIONS=${ADDITIONALVERSIONS}
```

src/dotnet/install.sh
```diff
@@ -4,44 +4,10 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 #
-# Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/dotnet.md
-# Maintainer: The VS Code and Codespaces Teams
-
-DOTNET_VERSION="${VERSION:-"latest"}"
```
samruddhikhandale, line 10: Same here, can we add back this variable? 

```
DOTNET_VERSION="${VERSION}"
```

samruddhikhandale:
> Regarding the C# for Visual Studio Code extension:
> 
> Recently it has taken a dependency on [.NET Install Tool for Extension Authors](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime) which now causes a version of .NET Runtime to be downloaded and installed automatically.
> 
> image
> 
> I find this behavior slightly annoying
> 
> * the Output window steals focus
> * there is no caching of the installation, the runtime is downloaded each time VSCode is restarted
> * you end up with multiple installations of .NET CLI/Runtime
>   
>   * `/home/vscode/.vscode-server/data/User/globalStorage/ms-dotnettools.vscode-dotnet-runtime/.dotnet/7.0.10/dotnet`
>   * `/usr/share/dotnet/dotnet`
> 
> It is possible to configure an `existingDotnetPath`, which solves the second and third issues. However I am not confident that this is a good default.
> 
> ```json
> {
>   "customizations": {
>     "vscode": {
>       "settings": {
>         "dotnetAcquisitionExtension.existingDotnetPath": [
>           {
>             "extensionId": "ms-dotnettools.csharp",
>             "path": "/usr/share/dotnet/dotnet"
>           }
>         ]
>       }
>     }
>   }
> }
> ```
> 
> I am not sure if this will work well as a default if the user has configured an older .NET version (which could be incompatible with the C# extension?). Also, the output window will still steal focus.
> 
> image
> 
> Perhaps the config can be added as a code snippet to NOTES.md, and let users decide if they want to add this config to their devcontainer.json? @samruddhikhandale
> 
> Alternatively, I can configure `/usr/share/dotnet/dotnet` as a default and explain in NOTES.md how to override it by blanking out the array in `.vscode/settings.json`
> 
> image

I see, thanks for the detailed explanation. C# extension is very helpful for .NET development and is a great add-on for the dotnet Feature. That said, this new dependency could potentially add a strange and unpleasant experience for the users. However, I don't think if we should be removing this extension. 

Looping in @bamurtaugh for her thoughts (@bamurtaugh should we loop in .NET folks for their insights and check if there's a way to improve user experience?)

sliekens:
FWIW a new major version of the .NET runtime installer extension is coming. Any config we add for the current version is likely to stop working in the future.
https://github.com/dotnet/vscode-dotnet-runtime/issues/1240#issuecomment-1680759233

It's probably fine to only add a note to the README with instructions how to configure `existingDotnetPath` in `devcontainer.json`?

src/dotnet/install.sh
sliekens: Done

src/dotnet/install.sh
sliekens: Done

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: @bderusha @samruddhikhandale this is now implemented and documented

I tried to organize helper functions into separate files, but I'm not sure about the approach. The files are sourced in `install.sh` where I call their functions. There is a little bit of shared state that is not passed as an argument. (`DOTNET_INSTALL_SCRIPT` is defined in `install.sh` but only used in the helper functions.) I'm not sure how to make that less ugly or if I should just pass everything as function arguments.

src/dotnet/devcontainer-feature.json
sliekens: (Marking resolved because I think this works as intended)

src/dotnet/install.sh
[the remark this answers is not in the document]
sliekens: @bderusha I identified one area where this does matter: `dotnet test` uses [VSTest](https://github.com/microsoft/vstest) and updating it may cause issues like https://github.com/microsoft/vstest/issues/4657. So it can be favorable to stay on an older feature band like 7.0.3xx without pinning to an exact patch version.

I think that justifies keeping it and marking as resolved.

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
bderusha: It is not clear to me what the custom scripts are doing that isn't already handled by `vendor/dotnet-install.sh`. I think that the needs of this feature work well with the defaults to allow you to just pass the version directly into `vendor/dotnet-install.sh` for all cases except `lts`, which can be handled.

```sh
for version in "${versions[@]}"; do
  if [ "${version]" == "lts" ]; then
    args="--version latest --channel LTS --install-dir $DOTNET_INSTALL_DIR --no-path"
  else
    args="--version ${version} --install-dir $DOTNET_INSTALL_DIR --no-path"
  fi

  "$DOTNET_INSTALL_SCRIPT" "${args}"
done

## Repeat for dotnetRuntimeVersions and aspNetCoreRuntimeVersions with appropriate --runtime arg as well
```

Would something like this work or did I miss something? It certainly reduces the footprint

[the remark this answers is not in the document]
sliekens: @bderusha to clarify, the vendor script usage is really not as simple as you (and I) expected:

* `--version` is only used for an exact version like `6.0.412` 
* `--channel` is used for everything else (and optionally a `--quality` but I chose not to implement it)

So, a lot of the complexity comes from deciding when to use `--channel` instead of `--version`

Examples

* `--channel LTS`
* `--channel STS` (not implemented, we offer `latest` instead) (I don't think STS means latest, needs confirmation)
* `--channel 6.0` for pinning major.minor version
* `--channel 6.0.1xx` for pinning major.minor version and feature band 1 (SDK only)

More complexity comes from figuring out which channel contains the latest version, STS or LTS. This complexity could be removed if STS does actually always point to the latest version, but I have not found any indication of that being true.

In the end, I chose to handle the cases when I know for sure you need to use the `--channel` parameter, and only use `--version` as a fallback.

src/dotnet/NOTES.md
```diff
@@ -1,4 +1,71 @@
+## Configuration examples
```
samruddhikhandale, line 1: Nice 🎉

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
samruddhikhandale: > I tried to organize helper functions into separate files, but I'm not sure about the approach. The files are sourced in install.sh where I call their functions. There is a little bit of shared state that is not passed as an argument. (DOTNET_INSTALL_SCRIPT is defined in install.sh but only used in the helper functions.) I'm not sure how to make that less ugly or if I should just pass everything as function arguments.

Looking at all three files (`install-aspnetcore-runtime.sh`, `install-dotnet-runtime.sh`, `install-dotnet-sdk.sh`), seems like the functions look way too similar, and seems redundant. 

Suggestion: Adding a new file instead of adding functions to the `install.sh` is fine and acceptable. However, can we have one file instead of three?
- eg

```
fetch_latest_aspnetcore_runtime_version_in_channel() {
    local channel="$1"
    wget -qO- "https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$channel/latest.version"
}
```

Can be refactored like 👇 

```
fetch_latest_version_in_channel() {
    local channel="$1"
    local type="$2"
    wget -qO- "https://dotnetcli.azureedge.net/dotnet/$type/$channel/latest.version"
}
```

Similarly, can we update other two functions as well?

src/dotnet/install.sh
```diff
@@ -4,47 +4,22 @@
...
-
-MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
-DOTNET_ARCHIVE_ARCHITECTURES="amd64"
-DOTNET_ARCHIVE_VERSION_CODENAMES="buster bullseye bionic focal hirsute jammy"
-# Feed URI sourced from the official dotnet-install.sh
-# https://github.com/dotnet/install-scripts/blob/1b98b94a6f6d81cc4845eb88e0195fac67caa0a6/src/dotnet-install.sh#L1342-L1343
-DOTNET_CDN_FEED_URI="https://dotnetcli.azureedge.net"
-# Ubuntu 22.04 and on do not ship with libssl1.1, which is required for versions of .NET < 6.0
-DOTNET_VERSION_CODENAMES_REQUIRE_OLDER_LIBSSL_1="buster bullseye bionic focal hirsute"
+ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
```
samruddhikhandale, line 10:
```suggestion
ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS}"
```

I understand that you are trying to follow the old semantics, however, updating all such options would help sync the `ADDITIONAL_VERSIONS` var with the Feature defaults.

eg. 

```
"additionalVersions": {
            "type": "string",
            "default": "",
            "description": "Enter additional .NET SDK versions, separated by commas. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
        },
```

Hence, the CLI always sets `ADDITIONALVERSIONS` to "".  See https://containers.dev/implementors/features/#options-property Hence, this is no longer needed and its adds unwanted maintainence. eg. Updating default value of `additionalVersions` would always need a change to this line.

src/dotnet/install.sh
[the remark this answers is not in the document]
samruddhikhandale: > Run it against my fork?

Yes. However, feel free to contribute the change in a separate PR,. I don't think it's necessary/needed for this PR to be merged. 👍

src/dotnet/NOTES.md
sliekens: Thanks :)

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: Absolutely, taking care of it

src/dotnet/install.sh
sliekens: @samruddhikhandale does this comment also apply to `DOTNET_VERSION="${VERSION:-"latest"}"`? That's where added it back to make sure an empty string behaves the same as the default.

src/dotnet/devcontainer-feature.json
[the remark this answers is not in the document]
sliekens: @samruddhikhandale check again, I think you will be pleased with the result

src/dotnet/install.sh
```diff
@@ -34,128 +43,103 @@ check_packages() {
...
+        err "Unsupported ASP.NET Core Runtime version '${version}'. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
+        exit 1
+    fi
+done
+
 # Install .NET versions and dependencies
 # icu-devtools includes dependencies for .NET
 check_packages wget ca-certificates icu-devtools
 
+if [ "${versions[0]}" != "none" ]; then
```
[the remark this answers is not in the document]
sliekens: Updated, much better indeed

bderusha (approved):
Love the changes! This looks great. Nice work 🎉

jungaretti:
Agreed, looking good!

src/dotnet/devcontainer-feature.json
```diff
@@ -1,35 +1,44 @@
 {
     "id": "dotnet",
-    "version": "1.1.4",
+    "version": "2.0.0",
     "name": "Dotnet CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/dotnet",
-    "description": "Installs the .NET CLI. Provides option of installing sdk or runtime, and option of versions to install. Uses latest version of .NET sdk as defaults to install.",
+    "description": "This feature installs the latest .NET SDK, which includes the .NET CLI and the shared runtime. Options are provided to choose a different version or additional versions.",
```
samruddhikhandale, line 6:
```suggestion
    "description": "This Feature installs the latest .NET SDK, which includes the .NET CLI and the shared runtime. Options are provided to choose a different version or additional versions.",
```

samruddhikhandale (dismissed):
Super cool ✨ 

Thank you so much for contributing this PR and reiterating based on feedbacks, we appreciate it 🚀

src/dotnet/install.sh
samruddhikhandale: > The CLI does not use the default for version in this case.

I think that's the correct way ; whatever the user passes should be taken as input and we should only consider defaults when the option is omitted.

sliekens: Right, just to be clear, I can keep the fallback to `latest` for the version option?

``` sh
DOTNET_VERSION="${VERSION:-"latest"}"
```

Otherwise if you specify empty string for the version option then you get a technical error.

``` sh
12.58 Executing scripts/vendor/dotnet-install.sh --version  --channel  --install-dir /usr/share/dotnet --no-path
16.67 scripts/vendor/dotnet-install.sh: line 1253: effective_versions[$link_index]: unbound variable
16.68 ERROR: Feature "Dotnet CLI" (Unknown) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/dotnet for help troubleshooting this error.
```

samruddhikhandale: Yes, I think this is fine for now 👍

samruddhikhandale (dismissed):

sliekens:
@samruddhikhandale didn't see you were trying to merge, sorry. I added the workflow for updating the vendor script  37c220fb6cde72ebf5fee46e67d6c47bc0b8165e. I tested it against my fork, it created this PR 🔗https://github.com/sliekens/devcontainers-features/pull/3.

.github/workflows/update-dotnet-install-script.yml
```diff
@@ -0,0 +1,48 @@
+name: "Update dotnet-install script"
```
samruddhikhandale, line 1:
```suggestion
name: "Updates vendor 'dotnet-install' script"
```

src/dotnet/install.sh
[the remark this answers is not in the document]
sliekens: Done in 37c220fb6cde72ebf5fee46e67d6c47bc0b8165e

.github/workflows/update-dotnet-install-script.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Run fetch-latest-dotnet-install.sh
+        run: src/dotnet/scripts/fetch-latest-dotnet-install.sh
+
+      - name: Create a PR for dotnet-install.sh
+        id: push_image_info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
samruddhikhandale, line 19:
```suggestion
          GITHUB_TOKEN: ${{ secrets.PAT }}
```

Unfortunately, at org level we have the setting for "Allow GitHub to create and merge PRs" option disabled.

.github/workflows/update-dotnet-install-script.yml
```diff
@@ -0,0 +1,48 @@
...
+          echo "Start."
+
+          # Configure git and Push updates
+          git config --global user.email github-actions@github.com
+          git config --global user.name github-actions
+          git config pull.rebase false
+
+          branch=automated-script-update-$GITHUB_RUN_ID
+          git checkout -b $branch
+          message='Automated dotnet-install script update'
```
samruddhikhandale, line 31:
```suggestion
          message='[Updates] Automated vendor 'dotnet-install' script'
```

samruddhikhandale (commented):
No worries, left some minor comments. Thanks for adding the workflow

samruddhikhandale (approved):

.github/workflows/update-dotnet-install-script.yml
```diff
@@ -0,0 +1,48 @@
...
+          git config --global user.name github-actions
+          git config pull.rebase false
+
+          branch=automated-script-update-$GITHUB_RUN_ID
+          git checkout -b $branch
+          message='[Updates] Automated vendor 'dotnet-install' script'
+
+          # Add / update and commit
+          git add src/dotnet/scripts/vendor/dotnet-install.sh
+          git commit -m 'Automated dotnet-install script update [skip ci]' || export NO_UPDATES=true
```
sliekens, line 35: @samruddhikhandale it just occurred to me - maybe this commit should go through all the automated validations? (i.e. Remove `[skip ci]`.) Unlikely, but an update to the vendor script could potentially break the dotnet feature.

samruddhikhandale: That's a very good point! I agree, let's remove `[skip ci]` bit?

samruddhikhandale (approved):
Thanks! Let me know when you are ready, I can help merge this PR if necessary.

sliekens:
I'm pretty happy with the state of this branch, thanks to everyone who reviewed and gave feedback!

Issue #623, opened 2023-07-20
Custom .zshrc


j2udev:
I was looking to use the [common-utils](https://github.com/devcontainers/features/tree/main/src/common-utils) feature to setup zsh/ohmyzsh and noticed that there doesn't seem to be a way to use a custom .zshrc?  My Dockerfile is copying it into the remote user's home directory, but it gets overwritten.  The [install script for ohmyzsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/install.sh#L333-L336) has a block to account for existing .zshrc files (if the --keep-zshrc flag is set), but it doesn't appear to have made it into the [common-utils feature](https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#L458)

This is pretty important for doing any sort of customization

bamurtaugh:
Thanks for opening! @joshspicer @samruddhikhandale would be great to get your insights here.

samruddhikhandale:
Hi 👋 

We have https://github.com/devcontainers/features/pull/614 open which proposes to add a new Feature option (`installOhMyZshConfig`) to control the config (ie zsh rc files). Unless I have misunderstood the ask here, @j2udev would `installOhMyZshConfig:false` help, once available?  

Feel free to add your thoughts on the PR.

naturedamends:
In the meantime, you can add your dotfiles (folder or repository) via Dockerfile, via the dev-container dotfiles repository config option, via the mounts dev-container config option,  or via a volume in the `docker-compose.yaml` as seen in the config below. 

Then in the postInstallCommand you can rm the existing file: here is one of my devcontainer.json(s)

```json
{
	"name": "Python 3",

	"dockerComposeFile": [
		"docker-compose.yml"
	],
	"service": "development",

	"workspaceFolder": "/workspace",

	"customizations": {
		"vscode": {
			"settings": {
				"terminal.integrated.shell.*": "/usr/bin/zsh"
			},
			"extensions": [
				"firsttris.vscode-jest-runner",
				"esbenp.prettier-vscode"
			]
		}
	},
	
  "postCreateCommand": "mv /home/vscode/.zshrc /home/vscode/.zshrc.code; DOTFILES_STARTUP=\"/home/vscode/repos/dotfiles/machines/env/empty/\" DOTFILES_DOTFILES=\"/home/vscode/repos/dotfiles\" DOTFILES_ENV=\"empty\" bash -i /home/vscode/repos/dotfiles/machines/env/install.sh ",
[...]

PR #630, opened 2023-07-28
WIP: Common utils: config for oh-my-zsh repository location.

naturedamends:
Seems in the spirit of using your own repository for omz, if your features deviate from theirs.

Can merge this in with other pull requests for a more general update of the oh-my-zsh features, if you don't want to isolate them


samruddhikhandale:
@naturedamends Curious, what are some other popular (/widely used) repositories for configuring oh-my-zsh? It's great to add it in new Feature options and provide customization options. However, I would like to learn how useful this Feature option would turn out to be. 

Sometimes, I feel that providing too many options can confuse a user 😅 Would like to learn some other use cases for this one. Thank you! 😄 

Let me know what you think!

Also, looping in @j2udev would appreciate your thoughts as well. Ref: https://github.com/devcontainers/features/issues/623

naturedamends:
> @naturedamends Curious, what are some other popular (/widely used) repositories for configuring oh-my-zsh? It's great to add it in new Feature options and provide customization options. However, I would like to learn how useful this Feature option would turn out to be. 
> 
> 
> 
> Sometimes, I feel that providing too many options can confuse a user 😅 Would like to learn some other use cases for this one. Thank you! 😄 
> 
> 
> 
> Let me know what you think!
> 
> 
> 
> Also, looping in @j2udev would appreciate your thoughts as well. Ref: https://github.com/devcontainers/features/issues/623

Not sure about popular variants, but maintaining your own copy and updates?

Nvm feel free to close

j2udev:
> @naturedamends Curious, what are some other popular (/widely used) repositories for configuring oh-my-zsh? It's great to add it in new Feature options and provide customization options. However, I would like to learn how useful this Feature option would turn out to be.
> 
> Sometimes, I feel that providing too many options can confuse a user 😅 Would like to learn some other use cases for this one. Thank you! 😄
> 
> Let me know what you think!
> 
> Also, looping in @j2udev would appreciate your thoughts as well. Ref: #623

IMHO, the shell setup is a little too subjective to belong in something like "common-utils" in the first place.  To me, common utils should be things like setting up a non-root user (already doing that so 👍), syncing dev-container time with host time, setting the USER (==USERNAME) env var as many terminal prompts seem to expect that, setting up  man pages, etc.

While OhMyZsh is popular, there are other very popular solutions as well like [OhMyPosh](https://ohmyposh.dev/) (which I'm personally trying out atm and quite like it), [Starship](https://starship.rs/), and [Powerlevel10k](https://github.com/romkatv/powerlevel10k) to name a few. All of this is to explain why I feel an opinionated shell setup is probably too subjective to belong in common-utils.

Personal opinions aside, if OhMyZsh had its own feature instead of being coupled to common-utils, I think this would be fine (although there are really only two places that make sense IMO... either where their install script puts it: $HOME/.oh-my-zsh, or somwhere in $XDG_CONFIG_HOME).  While OhMyZsh setup lives in common-utils, I tend to think that the config added here is perhaps a little excessive.  My opinion might change if it didn't live in common-utils though :)

PR #631, opened 2023-07-29
new option to install SessionManagerPlugin

sikaoyuan:
Session Manager is useful to use aws-cli. There is simple patch to allow user install Session Manager Plugin when they use aws-cli feature.


yuanyi-wang:

samruddhikhandale (commented):
Hi 👋 

Thank you so much for opening the PR, we appreciate it. ✨ 

I understand the general use of session manager, however, I would like to understand how useful it would be for the community. @yuanyi-wang can you help me understand your use case in little more depth? Thanks!

yuanyi-wang:
when I need use Session manager to ssh from local machine to EC2 that does not have internet IP address, I need use aws-cli ssm plugin; also, when I need connect to database hosted on AWS using sql developer from local machine, I still need this plugin.
currently, I have to manually install it after initial devcontainer with aws-cli feature. if we can have this patch, I can install this plugin more easy.

samruddhikhandale:
Thanks @yuanyi-wang for providing the reasoning, makes sense to me! 

@bamurtaugh What do you think? I'd shortly post this to the [#devcontainers-community](https://github.com/orgs/devcontainers/discussions/3) slack channel for getting some opinions/gather interest from the community.

bamurtaugh:
Sharing this in the community Slack channel makes sense to me, thank you!

Also thanks for the context on the scenario. I agree it generally makes sense to me, though it'd be good to get a broader set of feedback (from Slack) too.

Issue #634, opened 2023-08-02
Docker-in-docker: Add retry mechanism into the docker init script (Failed to connect to Docker)


samruddhikhandale:
Sometimes docker fails to start within a container with the following error 👇 

> Failed to connect to Docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

As manually running `/usr/local/share/docker-init.sh` fixes this issue, add some retry mechanism into the docker-init script .

AndriiTsok:
For the last two weeks we have been experiencing the same issue. Running the script recovers the docker instance.

samruddhikhandale:
Thanks for reporting! Opened https://github.com/devcontainers/features/pull/637 which adds a retry logic to address the issue ^

AndriiTsok:
@samruddhikhandale Thank you for your fast reply!

We tried to create more than 30 codespaces during the last few hours to try to reproduce the issue. At the moment we are not able to produce `(*) Failed to start docker, retrying in 5s...` in creation logs. Earlier this morning the docker step was failing almost in every codespace. 

Today, we made sure that not the prebuilt image was used and that the codespaces are created for new branches. Also, we explicitly set 2.3.0 for the D-in-D feature.

We will keep an eye on the logs and the stability and will create another issue in case it is reproducible.

AndriiTsok:
@samruddhikhandale we have just managed to catch the error again. Here is the creation log: https://gist.github.com/AndriiTsok/1a62138fca79da47cb8d90db1b87ca9f

Failed to start docker line: https://gist.github.com/AndriiTsok/1a62138fca79da47cb8d90db1b87ca9f#file-gistfile1-txt-L284

Line with Docker daemon error: https://gist.github.com/AndriiTsok/1a62138fca79da47cb8d90db1b87ca9f#file-gistfile1-txt-L307

We simply trying to check the clusters and remove them for cases when we rebuilding the codespaces:
```bash
# Get the list of all existing clusters
clusters=$(k3d cluster list -o json | jq -r '.[].name')

# Iterate over each cluster and delete it
for cluster in $clusters; do
  echo "Deleting cluster $cluster"
  k3d cluster delete $cluster
done

# Create a new cluster with the given configuration
k3d cluster create --config .devcontainer/k3d/config.yaml --kubeconfig-update-default
```

Fails trying to get clusters.


P.S.
I have observed that the Docker daemon is now failing less frequently than it was just three days ago.

[...]

PR #637, opened 2023-08-04
Docker-in-docker: Add retries to the docker-init script until docker daemon starts

samruddhikhandale:
Closes: https://github.com/devcontainers/features/issues/634

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -5,7 +5,7 @@
 #-------------------------------------------------------------------------------------------------------------
 #
 # Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/docker-in-docker.md
-# Maintainer: The VS Code and Codespaces Teams
+# Maintainer: The Dev Container spec maintainers
 
 
 DOCKER_VERSION="${VERSION:-"latest"}" # The Docker/Moby Engine + CLI should match in version
@@ -347,7 +347,7 @@ fi
 
 tee /usr/local/share/docker-init.sh > /dev/null \
 << EOF
-#!/bin/sh
+#!/usr/bin/env bash
 #-------------------------------------------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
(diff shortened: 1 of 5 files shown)


AndriiTsok:
@samruddhikhandale Thank you! 

I hope this one will be prioritized. We experiencing a big dev experience issue working with inner loop env with a k8s cluster within the Codespaces. Every 3rd run fails, and we required to do full container rebuild.. sometimes multiple.

src/docker-in-docker/install.sh
```diff
@@ -347,7 +347,7 @@ fi
 
 tee /usr/local/share/docker-init.sh > /dev/null \
 << EOF
-#!/bin/sh
+#!/usr/bin/env bash
```
jkeech, line 350: This might be a breaking change? I'm pretty sure there are some other places where bash is required/assumed, but just calling this out in case folks were running on an image without bash installed previously.

jkeech (commented):
Can you add a test that verifies that the retries are working as expected? Left one comment about a breaking change. At minimum, this feels like a minor version bump to me since adding retries on the docker start is a new feature.

src/docker-in-docker/install.sh
samruddhikhandale: I agree, reverted that change!

samruddhikhandale:
> Can you add a test that verifies that the retries are working as expected? Left one comment about a breaking change. At minimum, this feels like a minor version bump to me since adding retries on the docker start is a new feature.

@jkeech Thanks for the review! Added a test which tests the retry logic and bumped the minor version. Also, see https://github.com/devcontainers/features/pull/637#discussion_r1286088476

test/docker-in-docker/test-scripts/docker-test-init.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+
+retry_count=0
```
jkeech, line 7: It would be good to add a comment saying this needs to be kept up-to-date with the retry logic in the install.sh script. A future improvement would be to pull this out into a helper lib that can be used within the install.sh and invoked from the test as well to reduce duplication and ensure the test is exercising the correct logic.

jkeech (approved):
Approved with one suggestion

test/docker-in-docker/test-scripts/docker-test-init.sh
samruddhikhandale: Makes sense, I will shortly create a follow up PR -- avoiding another commit to this PR, else it would dismiss your approval 🤦‍♀️

Issue #569, opened 2023-06-08
Node Feature fails when zsh is configured as default shell on Ubuntu dev container


oliverlabs:
I am using a base Ubuntu (latest) image: `"image": "mcr.microsoft.com/devcontainers/base:ubuntu"`.

I am specifying the node as a feature like so:

```
"features": {
    "ghcr.io/devcontainers/features/node:1": {}
}
```

This fails my devcontainer build:

Here's an error message:

```
#21 [dev_containers_target_stage  9/14] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=node_6,target=/tmp/build-features-src/node_6     cp -ar /tmp/build-features-src/node_6 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/node_6  && cd /tmp/dev-container-features/node_6  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/node_6
#21 0.677 ===========================================================================#21 0.677 Feature       : Node.js (via nvm), yarn and pnpm#21 0.677 Description   : Installs Node.js, nvm, yarn, pnpm, and needed dependencies.#21 0.677 Id            : ghcr.io/devcontainers/features/node#21 0.677 Version       : 1.2.0#21 0.677 Documentation : https://github.com/devcontainers/features/tree/main/src/node#21 0.677 Options       :
#21 0.677     VERSION="lts"
#21 0.677     NODEGYPDEPENDENCIES="true"
#21 0.677     NVMINSTALLPATH="/usr/local/share/nvm"
#21 0.677     NVMVERSION="0.39.2"
#21 0.677 ===========================================================================
#21 0.882 Get:1 http://security.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
#21 0.886 Get:2 https://packages.microsoft.com/repos/azure-cli jammy InRelease [[357](https://github.com/oliverlabs/azdevcont/actions/runs/5212210292/jobs/9405513410#step:4:358)5 B]
#21 0.893 Get:3 https://dl.yarnpkg.com/debian stable InRelease [17.1 kB]
#21 0.936 Get:4 https://packages.microsoft.com/repos/azure-cli jammy/main amd64 Packages [786 B]
#21 0.948 Get:5 https://packages.microsoft.com/repos/azure-cli jammy/main all Packages [1096 B]
[...]

PR #645, opened 2023-08-09
Avoid unexpected shell expansion

shikanime:
Some variables may contain symbols such as wildcards, which can be expanded by other shell interpreters such as ZSH.

Fix: https://github.com/devcontainers/features/issues/569

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.2.0",
+    "version": "1.2.1",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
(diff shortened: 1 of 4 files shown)


shikanime:

samruddhikhandale (changes requested):
Looks great, thanks for taking the time to contribute this PR.
Left a minor comment.

samruddhikhandale:
> ERROR: failed to solve: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: write /usr/local/bin/git: no space left on device

Looks like the tests are failing due to ^ disk space issues , see https://github.com/devcontainers/features/actions/runs/5813161318/job/15762628752?pr=645#step:4:3293

@shikanime can you help remove the "install_node_on_universal_image" scenario? https://github.com/devcontainers/features/blob/5128a41391094f5423c781c8ccd4fe6d4ae094fc/test/node/scenarios.json#L40-L48 

The universal image is ~ 12.5 GB, we shouldn't be using such a big image which will result in slow (local) test runs. Also, it would help fix the disk space issue in the runners.

I'll create an issue for myself to re-add a (slim) test scenario as a replacement.

samruddhikhandale:
Opened https://github.com/devcontainers/features/issues/646

samruddhikhandale (approved):
Looks great, thanks!

Issue #647, opened 2023-08-10
common-utils not installing ZSH correctly if devcontainer is based on pre-build without ZSH


ankudinov:
I'm testing devcontainers/ci for my builds currently.
The plan is:

- build base (Python slim, low size, no ZSH)
- build dev (from base with ZSH, d-in-d, etc.)

When base is pre-build with "installZsh": false and "installOhMyZsh": false,  
building dev with both parameters set to true will result in no ZSH installed.

Same if I'm trying to build dev locally from pre-build base.

Here is the marker file for the base container:

```bash
$ cat /usr/local/etc/vscode-dev-containers/common
    PACKAGES_ALREADY_INSTALLED=true
    LOCALE_ALREADY_SET=true
    EXISTING_NON_ROOT_USER=vscode
    RC_SNIPPET_ALREADY_ADDED=true
    ZSH_ALREADY_INSTALLED=
```

I can see the following log for the dev build:

```
  #22 0.074 ===========================================================================
  #22 0.074 Feature       : Common Utilities
  #22 0.074 Description   : Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.
  #22 0.074 Id            : ghcr.io/devcontainers/features/common-utils
  #22 0.074 Version       : 2.0.11
  #22 0.074 Documentation : https://github.com/devcontainers/features/tree/main/src/common-utils
  #22 0.074 Options       :
  #22 0.074     INSTALLZSH="true"
  #22 0.074     CONFIGUREZSHASDEFAULTSHELL="true"
  #22 0.074     INSTALLOHMYZSH="true"
  #22 0.074     UPGRADEPACKAGES="false"
  
  
  [2023-08-10T13:03:09.443Z] #22 0.074     USERNAME="vscode"
  #22 0.074     USERUID="1000"
  #22 0.074     USERGID="1000"
  #22 0.074     NONFREEPACKAGES="false"
  #22 0.074 ===========================================================================
  #22 0.086 Marker file found:
  #22 0.088     PACKAGES_ALREADY_INSTALLED=true
  #22 0.088     LOCALE_ALREADY_SET=true
  #22 0.088     EXISTING_NON_ROOT_USER=vscode
  #22 0.088     RC_SNIPPET_ALREADY_ADDED=true
  #22 0.088     ZSH_ALREADY_INSTALLED=
  #22 0.114 chsh: Warning: /bin/zsh does not exist
  
  
[...]

PR #649, opened 2023-08-11
[common-utils]: Bug fix: Install zsh on an image previously built with "installZsh:false"

samruddhikhandale:
Closes: https://github.com/devcontainers/features/issues/647

Testes locally with an image provided in https://github.com/devcontainers/features/issues/647#issuecomment-1674386348. Works as expected! 

Unfortunately, I wasn't able to write a test for such a complex scenario, feedbacks on how to write a test are welcome 🙏


naturedamends:
Please also see https://github.com/devcontainers/features/pull/653

joshspicer (approved):
Looks good to me! Most of this is refactor right? The important bit is that installing `zsh` won't be blocked if `PACKAGES_ALREADY_INSTALLED`?

samruddhikhandale:
> The important bit is that installing zsh won't be blocked if PACKAGES_ALREADY_INSTALLED?

@joshspicer Yes, that's correct!

PR #652, opened 2023-08-12
Dind optional login

jameswinegar:
I am running `devpod` on an AKS cluster and am running into authentication issues even though the cluster has `AcrPush` and `AcrPull` permissions. Since the build is happening inside of a container instead of on the node, the system identity can be used, but a `docker login` command needs to be executed, and it will leverage the [Managed identity for Azure resources](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#authentication-options).

When looking into solutions that did not involve janky practices, I decided it made sense to put `login` into the docker installation process. 

This PR adds login as an optional parameter and a supporting test case.

### DIFF
--- a/src/docker-in-docker/devcontainer-feature.json
+++ b/src/docker-in-docker/devcontainer-feature.json
@@ -45,6 +45,11 @@
             "type": "boolean",
             "default": true,
             "description": "Install Docker Buildx"
+        },
+        "login": {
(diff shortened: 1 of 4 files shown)


jameswinegar:

jameswinegar:
seems not the right thing to do after testing.

PR #653, opened 2023-08-13
Common-utils: various cleaning things from other pr.

naturedamends:
## Proposal

Enable usage of the `installOhMyZshConfig` flag within Microsoft base dev container images.

Since images may have had previous installs which come with commom-utils. This allows the config option to function as stated.

## failed proposal

This is now just cleaning up the file and fixes some issues with the flag for non root usernames.


naturedamends:
Tests are failing locally due to a permissions issue writing to `~/.oh-my-zsh owned by `vscode` as user `root` to

src/common-utils/main.sh
```diff
@@ -531,6 +545,7 @@ echo -e "\
     LOCALE_ALREADY_SET=${LOCALE_ALREADY_SET}\n\
     EXISTING_NON_ROOT_USER=${EXISTING_NON_ROOT_USER}\n\
     RC_SNIPPET_ALREADY_ADDED=${RC_SNIPPET_ALREADY_ADDED}\n\
-    ZSH_ALREADY_INSTALLED=${ZSH_ALREADY_INSTALLED}" > "${MARKER_FILE}"
+    ZSH_ALREADY_INSTALLED=${ZSH_ALREADY_INSTALLED}\n
```
samruddhikhandale, line 548:
```suggestion
    ZSH_ALREADY_INSTALLED=${ZSH_ALREADY_INSTALLED}\n\
```

src/common-utils/main.sh
```diff
@@ -531,6 +545,7 @@ echo -e "\
...
+    MARKED_INSTALL_OH_MY_ZSH_CONFIG=${MARKED_INSTALL_OH_MY_ZSH_CONFIG}" > "${MARKER_FILE}"
```
samruddhikhandale, line 549:
```suggestion
    INSTALL_OH_MY_ZSH_ALREADY_CONFIGURED=${INSTALL_OH_MY_ZSH_ALREADY_CONFIGURED}" > "${MARKER_FILE}"
```

Refactor to follow previous conventions?

src/common-utils/main.sh
```diff
@@ -426,7 +426,24 @@ if [ "${RC_SNIPPET_ALREADY_ADDED}" != "true" ]; then
     RC_SNIPPET_ALREADY_ADDED="true"
 fi
 
+echo "USERNAME: $USERNAME"
+ls -la ~
+ls -la "${user_home}"
```
samruddhikhandale, line 431: Looks like this was used for debugging? Can we remove it?

src/common-utils/main.sh
```diff
@@ -426,7 +426,24 @@ if [ "${RC_SNIPPET_ALREADY_ADDED}" != "true" ]; then
     RC_SNIPPET_ALREADY_ADDED="true"
```
samruddhikhandale, line 426: Can we bump the patch version? Thanks! 
https://github.com/devcontainers/features/blob/6e635a6ea0afee640bf51adcb5a758e39c7dcbea/src/common-utils/devcontainer-feature.json#L3

test/common-utils/scenarios.json
```diff
@@ -126,6 +126,16 @@
             }
         }
     },
+    "configure_zsh_no_template_first_step": {
+        "image": "debian:bullseye",
+        "remoteUser": "vscode",
```
samruddhikhandale, line 131:
```suggestion
        "remoteUser": "devcontainer",
```

I wonder if this would fix https://github.com/devcontainers/features/actions/runs/5849488185/job/15890491045?pr=653#step:4:15987

test/common-utils/scenarios.json
```diff
@@ -126,6 +126,16 @@
...
+        "features": {
+            "common-utils": {
+                "installZsh": true,
+                "installOhMyZshConfig": false
+            }
+        }
+    },
```
samruddhikhandale, line 138: Looks like there is a failing check, see https://github.com/devcontainers/features/actions/runs/5849488185/job/15890491045?pr=653#step:4:14862

samruddhikhandale (commented):
Thanks for contributing the PR, left some comments!

naturedamends:
Testing the marked flag I have is a little hard since I would need to build a base image within a test, since the avaliable base images have old version of the source code, and thus no flag.

src/common-utils/main.sh
```diff
@@ -468,27 +478,31 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
-        # Copy to non-root user if one is specified
-        if [ "${USERNAME}" != "root" ]; then
-            copy_to_user_files=("${oh_my_install_dir}")
-            [ -f "$user_rc_file" ] && copy_to_user_files+=("$user_rc_file")
+        # Copy to alternate user if one is specified
+        if [ "${USERNAME}" != "root" ] && (( ${#copy_to_user_files[@]} != 0 )); then
             cp -rf "${copy_to_user_files[@]}" /root
-            chown -R ${USERNAME}:${group_name} "${oh_my_install_dir}" "${user_rc_file}"
+            root_files=("/root/.oh-my-zsh")
+            [ -f /root/.zshrc ] && root_files+=("/root/.zshrc")
```
naturedamends, line 504: Given root already has .zshrc installed ie previous step was `installOhMyZshConfig true`
When a new step is ran with `installOhMyZshConfig true`
Then don't copy .zshrc?

samruddhikhandale:
Looks like the recent commits are causing multiple test failures. See https://github.com/devcontainers/features/actions/runs/5868233855?pr=653

naturedamends:
Do you understand this -- error with the coreutils cp? @samruddhikhandale 

```
cp -f /tmp/dev-container-features/common-utils_0/scripts/devcontainers.zsh-theme /home/vscode/.oh-my-zsh/custom/themes/devcontainers.zsh-theme
#14 0.396 + cp -f /home/vscode/.oh-my-zsh/custom/themes/devcontainers.zsh-theme /home/vscode/.oh-my-zsh/custom/themes/codespaces.zsh-theme
#14 0.397 cp: '/home/vscode/.oh-my-zsh/custom/themes/devcontainers.zsh-theme' and '/home/vscode/.oh-my-zsh/custom/themes/codespaces.zsh-theme' are the same file
```

test/common-utils/test.sh
```diff
@@ -13,6 +13,8 @@ check "zsh" zsh --version
 check "ps" ps --version
 check "Oh My Zsh! theme" test -e $HOME/.oh-my-zsh/custom/themes/devcontainers.zsh-theme
 check "zsh theme symlink" test -e $HOME/.oh-my-zsh/custom/themes/codespaces.zsh-theme
+check "zsh theme filename" test -e $HOME/.oh-my-zsh/custom/themes/codespaces.zsh-theme
```
naturedamends, line 16: I would take the meaning of -e and symlink to mean xor... not sure @samruddhikhandale

src/common-utils/main.sh
```diff
@@ -427,7 +427,27 @@ if [ "${RC_SNIPPET_ALREADY_ADDED}" != "true" ]; then
...
+omz_source_dirname=".oh-my-zsh"
+user_rc_file="${user_home}/${omz_rc_filename}"
+user_omz_install_dir="${user_home}/${omz_source_dirname}"
+template_path="${user_omz_install_dir}/templates/zshrc.zsh-template"
+
+# Allow upstream steps to use installOhMyZshConfig false
+# Given previous step configured ~/.zshrc,
+# When installOhMyZshConfig is false, done before INSTALL_ZSH since
+#  Or where installOhMyZshconfig false, and installZsh false
+# Then remove the file
```
samruddhikhandale, line 440: With the PR changes `installOhMyZshConfig` would behave completely different as compared to other Feature options.

For eg. Let's say an image (say `X`) installs common-utils as follows -

```jsonc
"features": {
            "common-utils": {
                "installZsh": true,
            }
        }
```

Then a user creates their `devcontainer.json` as 👇 

```
{
        "image": "X",
        "features": {
            "common-utils": {
                "installZsh": false
            }
        }
    }
```

> "installZsh": false

Does this mean the Feature should remove the `zsh` installation? I don't think that the Features should remove an existing functionality, however, should only update or add new tools/configs/themes.

When a user add a Feature to their `devcontainer.json`, it is assumed that the `"installZsh": false` option would NOT install `zsh` but it doesn't mean that it should remove the already existing `zsh`. This behavior could create confusions and removing something is always a bit risky. I think we should avoid this approach.

@naturedamends Let me know what you think!

naturedamends: Maybe, remove the file, if zshrc exists and installZshConfig false and installZsh true?

Or do you not like the file being removed at all?

Should I remove the source of zshrc added in tests? Such that it's not executed within within build runtime. Then again.. it's a container

samruddhikhandale: > Or do you not like the file being removed at all?

This ☝️ 
You never know if the zsh config has been modified by the dockerfiles (or something else), hence, it's best to avoid removing anything. Also, removing config makes sense only if the Feature had an option like  `removeZshConfig `, but doesn't seem like that's needed.

naturedamends: Hose about linking ~/.zshrc to .zshrc.devcontainer.

Them downstream can handle, or ln -sf?

src/common-utils/main.sh
```diff
@@ -450,57 +461,60 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
-        # Copy to non-root user if one is specified
         if [ "${USERNAME}" != "root" ]; then
-            copy_to_user_files=("${oh_my_install_dir}")
-            [ -f "$user_rc_file" ] && copy_to_user_files+=("$user_rc_file")
-            cp -rf "${copy_to_user_files[@]}" /root
-            chown -R ${USERNAME}:${group_name} "${oh_my_install_dir}" "${user_rc_file}"
+            # Copy files to alternate user if one is specified
+            cp -rf "${user_omz_filepaths[@]}" /root
+            # Set permissions for root user
+            chown -R root:root "${omz_added_filesnames[@]/#//root/}"
```
samruddhikhandale, line 510: Curious, why is this needed?

src/common-utils/main.sh
```diff
@@ -376,6 +376,15 @@ if [ "${USERNAME}" != "root" ] && [ "${EXISTING_NON_ROOT_USER}" != "${USERNAME}"
...
 
+# *********************************
+# ** Ensure config directory **
+# *********************************
+user_config_dir="${user_home}/.config"
+if [ ! -d "${user_config_dir}" ]; then
+    mkdir -p "${user_config_dir}"
+    chown ${USERNAME}:${group_name} "${user_config_dir}"
+fi
+
```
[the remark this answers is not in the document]
naturedamends: Yeah good catch. I want a clean home directory, so if im going to program this, I want it before any rc files. Not sure how much of a pain this is going to be 

I came to the conclusion that you would not want any related flags in common-utils

Is the shell-setup feature going to be in features repo?
Then would it be prudent to add a flag to display any rc generation here


Thus you can enable the following to get equal functionality:
Common-utils rc false
Shell-setup shell zsh


I'm not familiar with any other fancy shells. Other than one I heard of that has a consistent output table format.

Would you consider adding a shell-setup to the features repo?

Sorry for bad grammar on tablet

src/common-utils/main.sh
naturedamends: Security. User scripts making it into root and being sourced.

Ie custom shell scripts
Or scripts appended to users zshrc

Nvm. I will close pr.

PR #659, opened 2023-08-18
[docker-in-docker] Reverts the retry logic

samruddhikhandale:
Ref: https://github.com/devcontainers/features/issues/634

Reverts https://github.com/devcontainers/features/pull/637

 With addition of the retry logic, we have seen issues where docker starts but stops in between lifecycle hooks (esp `onCreateCommand`). See https://github.com/devcontainers/features/issues/634#issuecomment-1683282792
 
Reverting the retry logic as it seems to cause other issues, will follow-up with another issue to fix this and `sed: couldn't flush stdout: Device or resource` errors

<s>Even I was able to reproduce the issue, (repro https://github.com/TRYON-Technology/Docker-in-Docker-Issue) it happens quite too often. My hypothesis is that the retry logic checks for docker status too early and triggers for a retry which brings the docker in an inconsistent state. (Docker needs at least few mili-seconds to start). Hence, this PR aims to add sleep in between docker start and retry logic. 

Note: See https://github.com/samruddhikhandale/Docker-in-Docker-Issue, with these PR changes I was able to create codespaces successfully 10 times in a row. </s>


src/docker-in-docker/install.sh
```diff
@@ -435,13 +435,14 @@ do
         eval "${dockerd_start}"
     fi
 
+    sleep 5s
```
joshspicer, line 438: This will slow down all container starts by 5s, right?  I know we're trying to fix a problem that is likely racey, but waiting this long might be too big a performance hit?

src/docker-in-docker/install.sh
```diff
@@ -435,13 +435,14 @@ do
...
+
     set +e
         docker info > /dev/null 2>&1 && docker_ok="true"
 
         if [ "${docker_ok}" != "true" ]; then
-            echo "(*) Failed to start docker, retrying in 5s..."
+            echo "(*) Failed to start docker, retrying..."
             retry_count=`expr $retry_count + 1`
-            sleep 5s
```
joshspicer, line 444: reading your PR description, perhaps just lowering this delay (or making is low to start and then increase exponentially) would help?  That still provides the opportunity for this loop to exit quickly, but then also retries again in a reasonable amount of time?

samruddhikhandale: > My hypothesis is that the retry logic checks for docker status too early and triggers for a retry which brings the docker in an inconsistent state.

Mmm... The issue I am trying to fix is to ensure that we have some buffer in between https://github.com/devcontainers/features/pull/659/files#diff-a72a4e49c44e54f09feb84b7767d355d68d34fbee561d8f96a99d12d2488b9c5R431 and https://github.com/devcontainers/features/pull/659/files#diff-a72a4e49c44e54f09feb84b7767d355d68d34fbee561d8f96a99d12d2488b9c5R441.

So how would lowering this help? 🤔 
The buffer in between docker start ---> retry was 5s and it is still the same.

samruddhikhandale: Reverted!

joshspicer (approved):

Issue #386, opened 2023-01-09
Add option to opt out of extensions


ghost:
Some features install extensions along with themselves and I'm not always interested in them. For example, the `golang.Go` extension that is installed in the Go feature or the `dbaeumer.vscode-eslint` extension that is installed in the Node feature are not extensions that I use. 

To prevent bloat it would be very nice if extensions could be opted out of in feature options.

Here is an example of how I think this could work:
```json
"features": {
  "ghcr.io/devcontainers/features/node:1": {
     "extensions": {
        "dbaeumer.vscode-eslint": false
     }
  }
}
```

samruddhikhandale:
Hi @andreilg 👋 

https://github.com/devcontainers/cli/pull/262 introduced an experimental `--skip-persisting-customizations-from-features` flag for the `devcontainer build` command. It omits **all** the Feature specified `customizations` ie. if your `devcontainer.json` referred to a set of Features, then their corresponding `customizations` (which includes `extensions`) will not be applied to your dev container.

Wondering if that could be sufficient for your scenario? I could help make this flag more formalized.

Let me know if your scenario requires more control/options to pick & choose the customizations/extensions, then we could chat further.

andreilg:
Thank you for your response.

The way I'm currently using devcontainers is through a `devcontainer.json` file with features running in GitHub Codespaces.

That cli flag seems helpful but I don't necessarily want to turn off all customizations or even all extensions but rather pick and choose within the `devcontainer.json` feature what customizations to include.

My suggested example above is how I envision this could work since it allows for granular control of customizations.

samruddhikhandale:
Thanks for the clarification.

> Here is an example of how I think this could work:
> ```json
> "features": {
>     "ghcr.io/devcontainers/features/node:1":
>         "extensions": [
>             "dbaeumer.vscode-eslint": false
>         ]
> }
> ```

Shouldn't `extensions` be wrapped in an object? 👇 


```json
"features": {
    "ghcr.io/devcontainers/features/node:1": {
          "extensions": [
              "dbaeumer.vscode-eslint": false
          ]
     }
}
```

Currently, every key-value inside this object is expected to be a [Feature option](https://containers.dev/implementors/features/#options-property). Doing something like this needs a spec change & would need to reserve 'extensions' keyword for this specific purpose. This could potentially be a breaking change for users who have been using `extensions` as a Feature option in their `devcontainer-feature.json`.

[...]

PR #665, opened 2023-08-23
[Rust] Remove `mutantdino.resourcemonitor` extension

nyuyuyu:
I don't see this extension as essential to Rust development.

As far as I know, there is no way to opt-out of the extension defined here(Ref: https://github.com/devcontainers/features/issues/386), and I would like to see it removed.

Users who need this extension can set it to `dev.containers.defaultExtensions` in their own configuration files.

### DIFF
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.0.12",
+    "version": "1.0.13",
     "name": "Rust",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/rust",
     "description": "Installs Rust, common Rust utilities, and their required dependencies",
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.0.12",
+    "version": "1.1.0",
     "name": "Rust",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/rust",
     "description": "Installs Rust, common Rust utilities, and their required dependencies",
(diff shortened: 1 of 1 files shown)


src/rust/devcontainer-feature.json
```diff
@@ -39,7 +39,6 @@
         "vscode": {
             "extensions": [
                 "vadimcn.vscode-lldb",
-                "mutantdino.resourcemonitor",
```
samruddhikhandale, line 42: From https://marketplace.visualstudio.com/items?itemName=mutantdino.resourcemonitor, doesn't seem like this extension is useful or even essential for rust scenarios. Removing makes sense to me!

Thanks @nyuyuyu for bringing this up! ❇️ 

// Would appreciate any other thoughts from @bamurtaugh @joshspicer (Let me know if I am missing any use case)

src/rust/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.0.12",
+    "version": "1.0.13",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.1.0",
```

Can we bump the minor version instead?

We usually bump major versions for breaking changes, however, this doesn't seem like a breaking change to me. Hence, bumping minor version should be fine.

src/rust/devcontainer-feature.json
bamurtaugh: I agree with removing it - the extension seems to be a nice to have, but I don't see it as being essential for Rust development. It's also not mentioned in the VS Code docs on Rust: https://code.visualstudio.com/docs/languages/rust.

nyuyuyu:
Thank you for agreeing to my request 😄 

> Can we bump the minor version instead?

Sure. bumped the minor version with [6da5dc7](https://github.com/devcontainers/features/pull/665/commits/6da5dc7eafcf1b67e78dad1887234974e969aced) commit.

Issue #660, opened 2023-08-18
Docker-in-docker - "sed: couldn't flush stdout: Device or resource" errors


samruddhikhandale:
Tracks issues with docker startup due to `sed: couldn't flush stdout: Device or resource errors`. See https://github.com/devcontainers/features/actions/runs/5907721355/job/16026113620?pr=659#step:4:1968

Need to add a retry mechanism around this logic --> https://github.com/devcontainers/features/blob/d48a9d534eb321385826058f3abc24bb7f6c6702/src/docker-in-docker/install.sh#L375-L398

samruddhikhandale:
`docker-in-docker v2.4.0` includes the following changes, see https://github.com/devcontainers/features/pull/669

- Adds retries for docker daemon startup
- We have seen errors like `sed: couldn't flush stdout: Device or resource errors` which fails to start, this adds retries to fix such `sed` errors
- Adds a workflow which runs 100 jobs for validating docker startup --> 50 jobs validates dockerd after the container is started, and 50 jobs validates within the `onCreateCommand`. The stress test is 🟢 

**Important Note:** `/usr/local/share/docker-init.sh` which starts/retries dockerd is added to the `entrypoint` command. This command runs in the background and is not a blocking script for the container startup. Since it's in the background, `onCreateCommand/postCreateCommand/postStartCommand` could all start executing before docker is fully running. If it takes docker too long, that could introduce flakiness in those lifecycle scripts. 

Opened https://github.com/devcontainers/spec/issues/299 which requests a new semantics to have "blocking" entrypoints that the CLI waits for. This way we can ensure that docker is already up and running for the mentioned ^ lifecycle scripts and is available in the container.

Closing in the favor of https://github.com/devcontainers/features/issues/671. Feel free to reopen if needed, or comment on https://github.com/devcontainers/features/issues/671 if you still run into docker not running issues. Thank you!

PR #669, opened 2023-08-24
[docker in docker] Adds retries for docker daemon startup and cgroup nesting

samruddhikhandale:
Ref: https://github.com/devcontainers/features/issues/634 and https://github.com/devcontainers/features/issues/660

Updates in this PR - 
- https://github.com/devcontainers/features/pull/669/commits/19c98929080f3e5283964cf881d99d2fed1a9a0d adds retries for docker daemon startup
- We have seen errors like `sed: couldn't flush stdout: Device or resource errors` which fails to start docker.https://github.com/devcontainers/features/pull/669/commits/25d2e8069c11f35acbf959fdf2c1e44239bc7969 adds retries to fix such `sed` errors
- https://github.com/devcontainers/features/pull/669/commits/4b2c9fc1331fa611fd5e74570c26c212aab08eb7 adds a workflow which runs 100 jobs for validating docker startup --> 50 jobs validates dockerd after the container is started, and 50 jobs validates within the `onCreateCommand`

**Important Note:** `/usr/local/share/docker-init.sh` which starts/retries dockerd is added to the `entrypoint` command. This command runs in the background and is not a blocking script for the container startup. Since it's in the background, `onCreateCommand/postCreateCommand/postStartCommand` could all start executing before docker is fully running. If it takes docker too long, that could introduce flakiness in those lifecycle scripts. 

Opened https://github.com/devcontainers/spec/issues/299 which requests a new semantics to have "blocking" entrypoints that the CLI waits for. This way we can ensure that docker is already up and running for the mentioned ^ lifecycle scripts and is available in the container.


samruddhikhandale:
Note: The [failing test](https://github.com/devcontainers/features/actions/runs/5967344914/job/16188874984?pr=669) will be fixed with https://github.com/devcontainers/features/pull/661

joshspicer (approved):
nice! This looks really reasonable and the 50 passing tests doesn't hurt 😉

samruddhikhandale:
> nice! This looks really reasonable and the 50 passing tests doesn't hurt 😉

@joshspicer **100 😛 See https://github.com/devcontainers/features/actions/runs/5967344918/job/16188845503?pr=669

PR #673, opened 2023-08-25
[node] Install the latest version of nvm by default

### LINKED ISSUE #666: Why not the node Feature installs the latest nvm by default?
Related to #190 and #194

Currently the default version of nvm is fixed at 10 months old `0.39.2` and does not appear to auto-detect the latest version.

https://github.com/devcontainers/features/blob/71ac2021e1350e673c0602b2ab6526d4353bc293/src/node/devcontainer-feature.json#L31-L35

Do you think we can automatically select the latest version here as is done in a smattering of other Features?

eitsupi:
Close #666

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.2.1",
+    "version": "1.3.0",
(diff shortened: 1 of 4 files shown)


eitsupi:
The test failure does not seem to do with this change (#675).

samruddhikhandale (approved):
Looks good, thank you! ✨

Issue #680, opened 2023-08-31
Move from Terraform to OpenTF as it might be illegal for many companies to continue using HashiCorp's version


mering:
New Terraform LICENSE: https://github.com/hashicorp/terraform/blob/main/LICENSE

OpenTF manifesto with some more explanation: https://opentf.org/

Issue tracking first release: https://github.com/opentffoundation/roadmap/issues/10

maazghani:
I’m not sure we want to change behavior of the terraform feature out from under folks, as it’s the responsibility of the user/company to manage and maintain which tf option they use. 

That said, we do want to provide an option for those who choose to move. I’m working on adding, in preparation of, the first opentf release [here](https://github.com/devcontainers/features/commit/a1513754bbb4eb58bcd5edda9cdadb81dc093e97) Could we rename this to “add opentf support” or “implement opentf feature”?

samruddhikhandale:
Hi 👋 

Thanks for starting this discussion. Reiterating my response in https://github.com/orgs/devcontainers/discussions/72

The dev containers spec or the Features doesn't says that a dev container Feature(s) has to only install OSS components. The MIT license is for the Feature which is just an installer, and we install from an official source. What a user installs is governed by its own license and we are not redistributing by persisting it ourself. Hence, there are no changes/deprecations to be made to the Terraform Feature as it's still a legit option, thus, we have https://github.com/devcontainers/features/pull/697 opened to note the licensing changes in the README file.

We shouldn't be deprecating the Terraform Feature, and for adding support for the `OpenTofu` as it's new its a bit unclear if the community will embrace it or not. Hence, we can keep https://github.com/devcontainers/features/issues/695 open and reiterate on our decision based on community interest/feedbacks.

Let us know if you any followup questions/concerns. Thank you! 

Closing this in favor of https://github.com/devcontainers/features/issues/695

PR #692, opened 2023-09-19
add opentf feature 

maazghani:
#680


maazghani:

Issue #588, opened 2023-06-21
Can't install jupyterlab in debian-slim


pamelafox:
Here's my setup:

```
FROM debian:bullseye-slim
```

devcontainer.json:

```
{
	"name": "Data 8 notebooks",
	"build": {
		"dockerfile": "dockerfile",
		"context": ".."
	},
	"features": {
		"ghcr.io/devcontainers/features/python:1": {
			"version": "3.11",
			"installJupyterlab": true
		}
	},
```

That results in this error:

```
#18 325.5 ./install.sh: line 307: /usr/local/python/3.11.4/bin/python3 -m pip install --user --upgrade --no-cache-dir jupyterlab: No such file or directory
#18 325.5 ERROR: Feature "Python" (ghcr.io/devcontainers/features/python) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/python for help troubleshooting this error.
```

Notably, I'm intentionally using a Debian that has no built-in Python package, as I'm trying to create an image with a single Python version. So maybe the PYTHON_SRC logic isn't compatible with that?

samruddhikhandale:
HI 👋 

@alexander-smolyakov can you help investigate? thanks! 

@pamelafox  In the meanwhile, can you use the `mcr.microsoft.com/devcontainers/base:bullseye` image instead? Dev container was successfully created with it 👇 

![image](https://github.com/devcontainers/features/assets/24955913/4aae8d79-ae6a-43f9-a7b9-b55e37122c53)

pamelafox:
Thanks @samruddhikhandale - that does build, but I believe it also has a base Python version   from the OS (try running `python 3.9`). I'm using slim to avoid that, and only have one version.

Also, I haven't managed to get that one to load in Jupyterlab (from "your codespaces" tab), but I'll bring that up with Codespaces team.

alexander-smolyakov:
Hey @pamelafox, we have merged the fix for the Python feature. Could you please check if the issue is resolved now?

Link to fix:
- #694

alexander-smolyakov:
I'm closing this issue since it was resolved. Feel free to request a reopening if the issue still occurs.

PR #694, opened 2023-09-20
[python] Change the `sudo_if` function to avoid issues with Bash

alexander-smolyakov:
**Feature name**:

- ghcr.io/devcontainers/features/python

**Description**:

This PR aims to resolve the `No such file or directory` error when commands are executed via the `sudo_if` function.

_Changelog_:

- Removed double quotes to avoid issues with tokenization;
- Added ability to install the `jupyterlab` package under root user;
- Fixed `jupyterlab` tests;
- Added new tests to cover the "install under root user" scenario;

**Attached related issue**:

- #588

**Checklist**:


src/python/install.sh
```diff
@@ -304,7 +304,7 @@ sudo_if() {
     if [ "$(id -u)" -eq 0 ] && [ "$USERNAME" != "root" ]; then
         su - "$USERNAME" -c "$COMMAND"
     else
-        "$COMMAND"
+        $COMMAND
     fi
```
samruddhikhandale, line 308: Can we fix the `jupyterlab-git` check failures? We'd need to update 👇  to `check "jupyterlab_git" grep jupyterlab_git <<< "$packages" `.

https://github.com/devcontainers/features/blob/1f83ebc28c0b7966726151d8712c785c5c5f4f99/test/python/install_jupyterlab.sh#L20

Same change to https://github.com/devcontainers/features/blob/1f83ebc28c0b7966726151d8712c785c5c5f4f99/test/python/install_additional_jupyterlab.sh#L20

src/python/install.sh
```diff
@@ -304,7 +304,7 @@ sudo_if() {
     if [ "$(id -u)" -eq 0 ] && [ "$USERNAME" != "root" ]; then
         su - "$USERNAME" -c "$COMMAND"
     else
-        "$COMMAND"
+        $COMMAND
```
samruddhikhandale, line 307: Can we bump the minor version? Thanks! https://github.com/devcontainers/features/blob/1f83ebc28c0b7966726151d8712c785c5c5f4f99/src/python/devcontainer-feature.json#L3

src/python/install.sh
alexander-smolyakov: Fixed in https://github.com/devcontainers/features/pull/694/commits/51bec72fd3e2c89d825e831c0c12fb3325d47e12

src/python/install.sh
alexander-smolyakov: Fixed in f4ff7f9920ea62151fc90d9342e7a128042da4fd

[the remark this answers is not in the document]
alexander-smolyakov: I added the ability to install the jupyterlab package under the root user and implemented tests to cover related changes.

src/python/install.sh
```diff
@@ -304,25 +304,31 @@ sudo_if() {
...
+        sudo_if "${PYTHON_SRC}" -m pip install --upgrade --no-cache-dir "$PACKAGE"
+    else
+        sudo_if "${PYTHON_SRC}" -m pip install --user --upgrade --no-cache-dir "$PACKAGE"
+    fi
 }
 
 add_user_jupyter_config() {
-    CONFIG_DIR="/home/$USERNAME/.jupyter"
-    CONFIG_FILE="$CONFIG_DIR/jupyter_server_config.py"
+    CONFIG_DIR=$1
```
samruddhikhandale, line 323:
```suggestion
    CONFIG_DIR="$1"
```

src/python/install.sh
```diff
@@ -461,13 +467,26 @@ if [ "${INSTALL_JUPYTERLAB}" = "true" ]; then
         exit 1
     fi
 
-    install_user_package jupyterlab
-    install_user_package jupyterlab-git
+    INSTALL_UNDER_ROOT=true;
```
samruddhikhandale, line 470:
```suggestion
    INSTALL_UNDER_ROOT=true
```

samruddhikhandale (approved):
Left some minor comments, besides that LGTM

samruddhikhandale (approved):
Left some minor comments, besides that LGTM

src/python/install.sh
alexander-smolyakov: Fixed in b39c96402c296ff9d158f61741af84d0b07c0003

src/python/install.sh
alexander-smolyakov: Fixed in b39c96402c296ff9d158f61741af84d0b07c0003

samruddhikhandale (approved):
Thank you! 🚢

PR #697, opened 2023-09-20
Terraform - Adds licensing information

samruddhikhandale:


src/terraform/NOTES.md
```diff
@@ -1,5 +1,9 @@
 
 
+## Licensing
+
+On August 10, 2023, HashiCorp announced a change of license for its products, including Terraform. After ~9 years of Terraform being open source under the MPL v2 license, it was to move under a non-open source BSL v1. 1 license, starting from the next (1.6) version. See https://github.com/hashicorp/terraform/blob/main/LICENSE
```
bamurtaugh, line 5: Is this supposed to have a space?

```suggestion
On August 10, 2023, HashiCorp announced a change of license for its products, including Terraform. After ~9 years of Terraform being open source under the MPL v2 license, it was to move under a non-open source BSL v1.1 license, starting from the next (1.6) version. See https://github.com/hashicorp/terraform/blob/main/LICENSE
```

PR #698, opened 2023-09-24
[Updates] Automated vendor dotnet-install script

samruddhikhandale:
[Updates] Automated vendor dotnet-install script


sliekens:
I think this is pretty cool 😎

samruddhikhandale (commented):
This is super cool, thank you @sliekens for the workflow! 🎉 

One 💡 , as we need version bump for the Feature to get republished, it would be cool if the workflow could auto-bump the patch version for the dotnet Feature.

joshspicer (approved):
nice!

PR #701, opened 2023-09-24
[java] fix version regex to support graalce jdk 21

### LINKED ISSUE #700: Java feature not supporting jdk 21 from GraalCE
When using Java (via SDKMAN!) feature and trying to use `21` version of `graalce` distribution, it fails to recognize the version

Also by looking up into the install.sh, I've noticed that there is a problem within the regex that validates versions being installed. it only supports versions with format x.y.z, being all numeric values. But for some scenarios, other version formats are not being recognized with the regex. example version format being `21-graalce` for graalVM community edition.

I'm working on a fork to fix the bug

prulloac:
Resolves #700 and also fixes failing test/java/install_from_non_default_distro.sh scenario

### DIFF
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -111,13 +111,13 @@ sdk_install() {
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
-        local regex="${prefix}\\K[0-9]+\\.[0-9]+\\.[0-9]+${suffix}"
+        local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
         if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ]; then
             requested_version="$(echo "${version_list}" | head -n 1)"
         else
             set +e
-            requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|$)")"
+            requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|-\s*|$)")"
             set -e
         fi
         if [ -z "${requested_version}" ] || ! echo "${version_list}" | grep "^${requested_version//./\\.}$" > /dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


prulloac:

src/java/install.sh
```diff
@@ -111,13 +111,13 @@ sdk_install() {
...
     else
-        local regex="${prefix}\\K[0-9]+\\.[0-9]+\\.[0-9]+${suffix}"
+        local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
         if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ]; then
             requested_version="$(echo "${version_list}" | head -n 1)"
         else
             set +e
-            requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|$)")"
+            requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|-\s*|$)")"
```
samruddhikhandale, line 120:
```suggestion
            requested_version="$(echo "${version_list}" | grep -E -m 1 "^${requested_version//./\\.}([\\.\\s]|-|$)")"
```

To me it's looks like the updated `requested_version` is additionally trying to match lines where `requested_version` is followed by a hyphen and zero or more whitespace characters.

I don't think we need to match whitespaces, also, with ^ code `install_non_conventional_version ` test scenario passed. 

Feel free to let me know if I have misunderstood the use case, thanks!

src/java/install.sh
```diff
@@ -111,13 +111,13 @@ sdk_install() {
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
-        local regex="${prefix}\\K[0-9]+\\.[0-9]+\\.[0-9]+${suffix}"
+        local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
```
samruddhikhandale, line 114: Can we update the minor version of the Feature, thanks! 

https://github.com/devcontainers/features/blob/d53f4f1c1d9cc34e537d32188d9ca5341467918a/src/java/devcontainer-feature.json#L3

samruddhikhandale (commented):
This looks great! Thanks for the quick fix, left some minor comments 🎉

src/java/install.sh
prulloac: Bumped patch version!

samruddhikhandale (approved):
Looks great, thank you!

PR #703, opened 2023-09-26
[common-utils] fix bug when user home is custom by base docker

### LINKED ISSUE #639: common-utils doesn't handle custom $HOME locations
The script expects `HOME=/home/$username`:

https://github.com/devcontainers/features/blob/685fa2e0f151b4545b444135a5ef6877de20e851/src/common-utils/main.sh#L385

But some docker images set the user home directory to somewhere else i.e.

https://github.com/siemens/kas/blob/a560dde924a9e6bdbdd74b901d71a450bec44646/Dockerfile#L59

prulloac:
Resolves #639 

main.sh first checks if user has home directory already set and different than /home/${USERNAME}. if it does, then uses the custom home directory, otherwise behaves as usual

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -407,16 +407,14 @@ fi
 
 if [ "${USERNAME}" = "root" ]; then
     user_home="/root"
+# Check if user already has a home directory other than /home/${USERNAME}
+elif [ "/home/${USERNAME}" != $( getent passwd $USERNAME | cut -d: -f6 ) ]; then
+    user_home=$( getent passwd $USERNAME | cut -d: -f6 )
 else
-    # Check if user already has a home directory other than /home/${USERNAME}
-    if [ "/home/${USERNAME}" != $( getent passwd $USERNAME | cut -d: -f6 ) ]; then
-        user_home=$( getent passwd $USERNAME | cut -d: -f6 )
-    else
-        user_home="/home/${USERNAME}"
-        if [ ! -d "${user_home}" ]; then
-            mkdir -p "${user_home}"
-            chown ${USERNAME}:${group_name} "${user_home}"
-        fi
+    user_home="/home/${USERNAME}"
+    if [ ! -d "${user_home}" ]; then
+        mkdir -p "${user_home}"
+        chown ${USERNAME}:${group_name} "${user_home}"
     fi
 fi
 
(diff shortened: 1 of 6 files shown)


src/common-utils/main.sh
```diff
@@ -408,10 +408,15 @@ fi
 if [ "${USERNAME}" = "root" ]; then
     user_home="/root"
 else
-    user_home="/home/${USERNAME}"
-    if [ ! -d "${user_home}" ]; then
-        mkdir -p "${user_home}"
-        chown ${USERNAME}:${group_name} "${user_home}"
+    # Check if user already has a home directory other than /home/${USERNAME}
```
samruddhikhandale, line 411: How about updating L:408 to LL421 with following code? 👇 (The logic is the same, just reducing the number of conditions to improve readability)

```
if [ "${USERNAME}" = "root" ]; then
    user_home="/root"
# Check if user already has a home directory other than /home/${USERNAME}
elif [ "/home/${USERNAME}" != $( getent passwd $USERNAME | cut -d: -f6 ) ]; then
    user_home=$( getent passwd $USERNAME | cut -d: -f6 )
else
    user_home="/home/${USERNAME}"
    if [ ! -d "${user_home}" ]; then
        mkdir -p "${user_home}"
        chown ${USERNAME}:${group_name} "${user_home}"
    fi
fi
```

src/common-utils/main.sh
```diff
@@ -408,10 +408,15 @@ fi
 if [ "${USERNAME}" = "root" ]; then
     user_home="/root"
```
samruddhikhandale, line 409: Can we bump up the minor version?

https://github.com/devcontainers/features/blob/ed0060ebc48e948f8ea11270a0c06fa840926a1b/src/common-utils/devcontainer-feature.json#L3

test/common-utils/scenarios.json
```diff
@@ -183,5 +183,21 @@
                 "configureZshAsDefaultShell": true
             }
         }
+    },
+    "devcontainer-custom-home": {
```
samruddhikhandale, line 187: thanks for adding them 🎉

samruddhikhandale (commented):
Thanks for contributing, we appreciate it ✨ 

Left some minor comments, besides that looks good to me.

src/common-utils/main.sh
prulloac: bumped to version 2.2.0

src/common-utils/main.sh
prulloac: thanks! I've added the suggestion to improve readability

samruddhikhandale (approved):
Looks great, thank you! 🎉

Issue #671, opened 2023-08-24
[docker-in-docker] - Use blocking entrypoints to ensure dockerd startup


samruddhikhandale:
We have seen flakiness in docker startups for `docker-in-docker` Feature where docker is not running in the container. See https://github.com/devcontainers/features/issues/634 and https://github.com/devcontainers/features/issues/660. 

https://github.com/devcontainers/features/pull/669 adds few retry mechanisms to attempt to fix this issue. 

`/usr/local/share/docker-init.sh` which starts/retries dockerd is added to the `entrypoint` property for the Feature. This command runs in the background and is not a blocking script for the container startup. Since it's in the background, `onCreateCommand/postCreateCommand/postStartCommand` could all start executing before docker is fully running. If it takes docker too long, that could introduce flakiness in those lifecycle scripts. 

Opened https://github.com/devcontainers/spec/issues/299 which requests a new semantics to have "blocking" entrypoints that the CLI waits for. This way we can ensure that docker is already up and running for the mentioned ^ lifecycle scripts and is 💯 available in the container.

This issue tracks updates to dnd Feature, when blocking entrypoint is available.

Blocked by -
- https://github.com/devcontainers/spec/issues/299

Christopholus:
Seconding this! We are unfortunately still running into this issue. We're using DDEV and transitioned from using docker-in-docker, to the universal:2 package as per the conversation over in #634 yet any DDEV command failed with the following error message:
`Could not connect to a docker provider. Please start or install a docker provider.
For install help go to: https://ddev.readthedocs.io/en/latest/users/install/`

Doing a partial and full rebuild is still netting us the same issue. The only thing we've found works is completely deleting the Codespace and recreating it from scratch. However after a single codespace inactivity timeout sends us back to the docker to no longer being recognized error.

Staying tuned here to see if there's any activity on this fix 🤞 

If it's helpful, here's a snapshot of our devcontainer.json
```{
{
	"image": "mcr.microsoft.com/devcontainers/universal:2",
	"features": {
		"ghcr.io/ddev/ddev/install-ddev:latest": {}
	},
	"forwardPorts": [3000],
	"portsAttributes": {
		"3000": {
			"label": "Vite DevServer"
		},
		"3306": {
			"label": "MySQL"
		},
		"6006": {
			"label": "Storybook"
		},
		"8027": {
			"label": "Mailhog"
		},
		"8036": {
			"label": "PHPMyAdmin"
		},
		"8443": {
[...]

PR #711, opened 2023-09-28
Docker-in-docker: Updates docker-init with "pkill"

samruddhikhandale:
Fixes : https://github.com/devcontainers/images/issues/729

When the universal image is built with latest `docker-in-docker` Feature, it fails to start docker daemon as `containerd` does not start (Reported in https://github.com/devcontainers/features/issues/671#issuecomment-1702809633) .The image is trying to retry the docker daemon startup, but as `containerd` (or/and `dockerd)` is not in a good state before, it fails to restart causing `docker ps` to fail.

>  find /run /var/run -iname 'docker*.pid' -delete || :
    find /run /var/run -iname 'container*.pid' -delete || :

We do have these commands ^ but it simply removes the PID files, does not stop/clean the dockerd and containerd daemons 

🧪  Tested these PR changes with `universal:dev@sha256:b7a147f15c61f02b8220d1719431f205f65e48267ffed93e0b9c642c14cf956a` image, `docker ps` is successful. 

image

### DIFF
--- a/src/docker-in-docker/devcontainer-feature.json
+++ b/src/docker-in-docker/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "docker-in-docker",
-    "version": "2.5.0",
+    "version": "2.6.0",
     "name": "Docker (Docker-in-Docker)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/docker-in-docker",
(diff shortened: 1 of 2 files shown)


joshspicer (approved):
looks very reasonable to me 👍

Issue #702, opened 2023-09-25
Node feature does not pick up proxy settings


jmedelmann:
The [install.sh](https://github.com/devcontainers/features/blob/main/src/node/install.sh) script for the node feature uses npm to install pnpm (Line 231: `npm install -g pnpm`).
Because npm does not pickup the proxy environmentals `http_proxy`, `https_proxy` and `no_proxy` (which are automatically propagated by docker when present in the `~/.docker/config.json` configuration as `httpProxy`, `httpsProxy` and `noProxy`) by itself, the script will always fail behind a proxy.
This is a bug in the install script because this is a special situation not automatically handled by Visual Studio Code or Docker but which must be taken care of by the calling script.

Could you please fix this as follows (bash-compatible)?
```bash
[ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
[ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
[ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
npm install -g pnpm
```

momohnasirumuri:
Run

bamurtaugh:
@samruddhikhandale I'd love to get your thoughts on this one!

samruddhikhandale:
The request makes sense to me. Wondering, @jmedelmann would you be interested in contributing a PR with this change? We'd really appreciate it, thanks!

We could use https://github.com/devcontainers/features/pull/451 as reference, which adds a new "httpProxy" Feature option for the Terraform Feature. It uses the proxy which is either set by the Feature option or if one exists in the env variables.

samruddhikhandale:
Closing as completed with https://github.com/devcontainers/features/pull/712

PR #712, opened 2023-10-03
Configure npm proxy when proxy environmentals exist

jmedelmann:
PR for issue #702 (Node feature does not pick up proxy settings).
Docker automatically sets up proxy environmentals in the container when configured (see https://docs.docker.com/network/proxy/) so usually everything works without additional configuration. An exception is npm, which does not pick up these environmentals and needs to be configured manually. This PR adds the necessary configuration steps.

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.3.0",
+    "version": "1.3.1",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -228,6 +228,9 @@ if type pnpm > /dev/null 2>&1; then
     echo "pnpm already installed."
 else
     if type npm > /dev/null 2>&1; then
+        [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
+        [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
+        [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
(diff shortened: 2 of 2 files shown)


jmedelmann:

momohnasirumuri (commented):
Run

samruddhikhandale (changes requested):
Thanks for contributing this change. 

Currently, this PR is automatically setting proxy if the env variables exist. Instead, can we wrap this logic inside a condition on a new Feature option (say `useProxyConfig:boolean` with default `false`). This way, the users are willingly enabling it as per their needs, and we won't have to worry about folks raising security concerns.

As per https://docs.docker.com/network/proxy/#configure-the-docker-client, adding proxies to `~/.docker/config.json` seems like a manual approach where users are opting into it. So wrapping it inside a Feature option makes sense to me!

src/node/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.3.0",
+    "version": "1.3.1",
```
samruddhikhandale, line 3: We should bump the minor version when we we add a new Feature option.

jmedelmann:
> Currently, this PR is automatically setting proxy if the env variables exist. Instead, can we wrap this logic inside a condition on a new Feature option (say `useProxyConfig:boolean` with default `false`). This way, the users are willingly enabling it as per their needs, and we won't have to worry about folks raising security concerns.
> 
> As per https://docs.docker.com/network/proxy/#configure-the-docker-client, adding proxies to `~/.docker/config.json` seems like a manual approach where users are opting into it. So wrapping it inside a Feature option makes sense to me!

These env variables are a kind of standard and almost all programs will use them automatically. If we introduce an additional option for just npm we will create an inconsistent state with no benefit. Note that other programs already use these environmentals. If we want to make them optional it would require to make them optional for the other programs as well.

samruddhikhandale (approved):
Had a quick chat with the team, turns out the devcontainers/cli supports the automatic support as well. Approving! ✅

PR #451, opened 2023-02-13
Terraform: Add proxy setting (#435)

ksaito1125:
Added proxy settings to features/terraform.

PR #719, opened 2023-10-09
Connect to Ubuntu keyserver via port 80

### LINKED ISSUE #717: Python – Installation from source fails because GnuPG cannot import keys without user ID
**Update & tl;dr**
I figured our company firewall blocks connections to the default `hkp` port. The only `hkps` keyserver (to which connections are **not** blocked) specified in `install.sh` strips user IDs of some keys. GnuPG then refuses to import those keys. A solution could be to access the Ubuntu keyserver via `hkp` through port 80, which is not blocked. See my last comment: https://github.com/devcontainers/features/issues/717#issuecomment-1752455796

**Original post**
Installation of the `python` feature from source fails:

- At least one of the [specified key servers](https://github.com/devcontainers/features/blob/4fca96b5e8a4bfc93679098cb19d73c65ce571eb/src/python/install.sh#L31-L33) doesn't report the user ID for five of [the requested keys](https://github.com/devcontainers/features/blob/4fca96b5e8a4bfc93679098cb19d73c65ce571eb/src/python/install.sh#L30),
-  hence GnuPG refuses to import those keys:

```shell
 => => # (*) Downloading GPG key...                                            
 => => # gpg: key 3A5CA953F73C700D: new key but contains no user ID - skipped  
 => => # gpg: key 04C367C218ADD4FF: new key but contains no user ID - skipped  
 => => # gpg: key 0EDDC5F26A45C816: new key but contains no user ID - skipped  
 => => # gpg: key 6AF053F07D9DC8D2: new key but contains no user ID - skipped  
 => => # gpg: key C9BE28DEE6DF025C: new key but contains no user ID - skipped  
```

`devcontainer.json`:
```json
{
  "name": "dev",
  "image": "mcr.microsoft.com/devcontainers/base:debian-11",
  "features": {
    "ghcr.io/devcontainers/features/python:1": {
      "version": "3.11",
    },
  },
}
```

I can manually reproduce the problem on that container:
```shell
vscode ➜ ~ $ gpg --keyserver "hkps://keys.openpgp.org" --recv-keys 0x3A5CA953F73C700D
gpg: key 3A5CA953F73C700D: new key but contains no user ID - skipped
gpg: Total number processed: 1
gpg:           w/o user IDs: 1
```

GnuPG version:
```shell
vscode ➜ ~ $ gpg --version
gpg (GnuPG) 2.2.27
libgcrypt 1.8.8
Copyright (C) 2021 Free Software Foundation, Inc.
License GNU GPL-3.0-or-later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /home/vscode/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
```

Any help would be greatly appreciated.

hoechenberger:
This helps work around corporate firewalls blocking the default HKP port 11371.

Closes #717

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "name": "Python",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/python",
   "description": "Installs the provided version of Python, as well as PIPX, and other common Python utilities.  JupyterLab is conditionally installed with the python feature. Note: May require source code compilation.",
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -28,7 +28,7 @@ ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
 DEFAULT_UTILS=("pylint" "flake8" "autopep8" "black" "yapf" "mypy" "pydocstyle" "pycodestyle" "bandit" "pipenv" "virtualenv" "pytest")
 PYTHON_SOURCE_GPG_KEYS="64E628F8D684696D B26995E310250568 2D347EA6AA65421D FB9921286F5E1540 3A5CA953F73C700D 04C367C218ADD4FF 0EDDC5F26A45C816 6AF053F07D9DC8D2 C9BE28DEE6DF025C 126EB563A74B06BF D9866941EA5BBD71 ED9D77D5"
-GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com
+GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com:80
 keyserver hkps://keys.openpgp.org
 keyserver hkp://keyserver.pgp.com"
 


hoechenberger:
@microsoft-github-policy-service agree company="DSM-Firmenich"

samruddhikhandale (commented):
Hi 👋 

Thanks for contributing this change and providing a detailed explanation on https://github.com/devcontainers/features/issues/717

Have you looked at https://github.com/devcontainers/features/issues/323? We intentionally updated the keyserver in https://github.com/devcontainers/features/issues/323 to address "gpg key" issues with several different users. Hence, as this could be a breaking change for users, I don't think we should submit this PR as is.

Instead, I wonder if a similar change to https://github.com/devcontainers/features/pull/451 would work in your case?

hoechenberger:
Hello @samruddhikhandale and thank you for the friendly and detailed message!

While I am behind a corporate firewall, we do **not** use a proxy server (none that is explicitly specified by the user anyway). Hence, specifying a proxy setting is unfortunately it an option for me.

Since you [mentioned](https://github.com/devcontainers/features/issues/323#issuecomment-1363185344) that apparently for some users the Ubuntu key server can **only** be accessed via port 80 (that would include me), while for others it can **only** be accessed via the default port (which is blocked in my network and in many others -- which is the reason why Ubuntu chose to support port 80 in addition in the first place!), I would like to revive your proposal of simply adding **both** options to all features? If the connection to one fails, the next one will be tried.

What do you think?

samruddhikhandale:
> , I would like to revive your proposal of simply adding both options to all features? If the connection to one fails, the next one will be tried.

Yes, that makes complete sense to me and thanks for being willing to contribute the change. Let me know if you need any help with the changes.

PR #730, opened 2023-10-21
[Apple ARM - NVIDIA CUDA] Skip install

noe-sautel:
Skip installation for ARM chip because Nvdia Cuda Toolkit is no longer supported.

Reference: [https://developer.nvidia.com/nvidia-cuda-toolkit-11_7_0-developer-tools-mac-hosts](https://developer.nvidia.com/nvidia-cuda-toolkit-11_7_0-developer-tools-mac-hosts)
PR related to this discussion : [https://github.com/orgs/devcontainers/discussions/78#discussioncomment-7343974](https://github.com/orgs/devcontainers/discussions/78#discussioncomment-7343974)

### DIFF
--- a/src/nvidia-cuda/install.sh
+++ b/src/nvidia-cuda/install.sh
@@ -17,6 +17,11 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+if [[ $(uname -m) = "arm64" ]]; then
+    echo -e "NVIDIA® CUDA Toolkit no longer supports development or running applications on machines with the 'arm64' architecture. Please consider an alternative or check for updates or support for your architecture."
+    exit 1
+fi
+
 apt_get_update()
 {
     if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
--- a/src/nvidia-cuda/devcontainer-feature.json
+++ b/src/nvidia-cuda/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "nvidia-cuda",
-  "version": "1.1.0",
+  "version": "1.1.1",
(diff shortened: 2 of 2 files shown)


noe-sautel:

src/nvidia-cuda/install.sh
```diff
@@ -17,6 +17,11 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+if [[ $(uname -m) = "aarch64" ]]; then
+    echo -e "NVIDIA® CUDA Toolkit no longer supports development or running applications on machines with the 'arm64' architecture. Please consider an alternative or check for updates or support for your architecture."
+    exit 1
```
samruddhikhandale, line 22: Doesn't https://developer.nvidia.com/nvidia-cuda-toolkit-11_7_0-developer-tools-mac-hosts refer to NVIDIA toolkit no longer supporting MacOS? 

Hence, instead of failing the entire Feature, why don't we update 👇 instead?

https://github.com/devcontainers/features/blob/9f152b5725273fef7d863d81e29c7e03d5263f14/src/nvidia-cuda/install.sh#L95-L98

samruddhikhandale, line 22: > Note that, I have other developers on my repository who are using containers with a classic chip architecture, and it's essential for me to retain this feature for them.
> My primary question is: How can I add a condition to specify the architecture and install or not the feature?

From the discussion, note that if the Feature returns exit code `1`, then your dev container creation will fail. Unless, you switch the Feature option `installToolkit` to `false`.

samruddhikhandale (commented):
Thanks for the prompt PR creation, left some clarifying questions.

PR #736, opened 2023-10-27
[common-utils] Bug fix: zsh missing rc snippet

samruddhikhandale:
Bug fixes: 
- We were incorrectly adding [rc_snippet.sh](https://github.com/devcontainers/features/blob/main/src/common-utils/scripts/rc_snippet.sh) for zsh. 
- `zsh` uses `.zprofile` instead of `.profile`. Hence, we source `.profile` within `~/.zprofile`. See https://superuser.com/questions/187639/zsh-not-hitting-profile/187673#187673

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -467,7 +467,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
         else
             global_rc_path="/etc/zsh/zshrc"
         fi
-        cat "${FEATURE_DIR}/scripts/rc_snippet.sh" >> /etc/zshrc
+        cat "${FEATURE_DIR}/scripts/rc_snippet.sh" >> ${global_rc_path}
         ZSH_ALREADY_INSTALLED="true"
     fi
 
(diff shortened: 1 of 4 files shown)


src/common-utils/main.sh
```diff
@@ -456,13 +457,17 @@ fi
 
 # Optionally configure zsh and Oh My Zsh!
 if [ "${INSTALL_ZSH}" = "true" ]; then
+    if [ ! -f "${user_home}/.zprofile" ] || ! grep -Fxq 'source $HOME/.profile' "${user_home}/.zprofile" ; then
```
joshspicer, line 460: is this $HOME variable supposed to be replaced or no?  It's in single quotes so I don't think it would

joshspicer (approved):

src/common-utils/main.sh
samruddhikhandale: > is this $HOME variable supposed to be replaced or no?

Nope, It should be saved without any expansion.

PR #737, opened 2023-10-30
Remove rebornix.Ruby extension

0sc:
The rebornix.Ruby extension is [now deprecated](https://github.com/rubyide/vscode-ruby/commit/79031948d032e97cf2a5116e27c167ade5125e15). 

This PR removes it from the list of preinstalled extension for the Ruby feature devcontainer.

### DIFF
--- a/src/ruby/devcontainer-feature.json
+++ b/src/ruby/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "ruby",
-    "version": "1.1.2",
+    "version": "1.2.0",
     "name": "Ruby (via rvm)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/ruby",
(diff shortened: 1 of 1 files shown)


src/ruby/devcontainer-feature.json
```diff
@@ -21,7 +21,6 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "rebornix.Ruby",
```
samruddhikhandale, line 24: This makes sense to me 👍 

Looping in @bamurtaugh for product insights.

bamurtaugh: Makes sense to me too, thank you!

samruddhikhandale (approved):
Thanks for the change, appreciate it!

timscharfenort8 (commented):
Good work

Issue #757, opened 2023-11-16
Ruby Feature: Which ruby version manager would you prefer?


samruddhikhandale:
Hello! 👋 

We are looking forward to get some feedback for refactoring the [Ruby Feature](https://github.com/devcontainers/features/tree/main/src/ruby).

Which Ruby version manager would you like the Feature to install by default? You can use the respective emojis for voting your preference! 

1. None - 😕 
2. `rbenv` - 🎉 
3. `rvm` - 🚀 
4. Any other (feel free to drop a comment)

For the ruby Feature, we have heard issues with conflicting ruby managers (`rbenv` and `rvm`) which can be problematic for complex projects. We have a community member who is willing to contribute a PR with the necessary changes ✨   See https://github.com/devcontainers/features/issues/603#issuecomment-1806853742 ❤ 

### (Draft) Proposal

- Create a new Feature option which will define the ruby version manager to be used for installing ruby
    - Let’s say we call it `rubyVersionManager`
    - It will be `string` with `enum: ["none", "rbenv", "rvm"])`
    - Default value: "TBD: Poll result"
- Based on the value provided by `rubyVersionManager`, the Feature will use that version manager to install ruby
- In case of `none`, we will first attempt to install it with `apt`. 
    - If it fails, then install from source.

Looking forward to everybody's feedback, thank you!

ThomasOwens:
@nevans

I was trying to not scope-creep this too much, but I think I see what you're getting at.

From what I can tell, the default `none` case does rely on whatever comes from the container. That isn't the best solution. It looks like Debian comes with Ruby and the `common-utils` feature runs an upgrade of all OS packages. If I'm understanding the scripts correctly, `none` always gets you the latest Ruby release (in the Debian repos, anyway). Not necessarily what people want. Plus, if Debian stopped distributing Ruby for some reason, we'd be in a tough spot.

I believe the solution is this:

1. Create a new Feature option `rubyVersionManager` to define the Ruby version manager. It will be a `string` with `enum: ["none", "rbenv", "rvm"]`. The default value will be the poll results here.
2. Create a new Feature option `rubyVersion` to define the version of Ruby to install. It will be a `string` with some proposals for supported Ruby versions. This should be freeform, though, and parsed out to download any version of Ruby.
3. Based on the values of `rubyVersionManager` and `rubyVersion`, install the specified Ruby version. Install from source for `none` or using the version manager otherwise. When installing from source, set the path variables properly.

[...]

Issue #603, opened 2023-06-29
Install only one of `rvm` or `rbenv` for managing ruby versions


nevans:
Is there any benefit to installing both rvm and rbenv?  

### Context

`rvm` does so much: it installs plugins that change the behavior of rubygems and the `gem` command, it overrides and updates a bunch of environment variables, it updates `umask`, it edits the `ruby` binary, and it performs various other mutations on your environment whenever you load your shell or change directories.

Unfortunately, if you load `rbenv` or place its shims earlier in the PATH than `rvm`, then some truly weird situations arise.  E.g: `which gem` reports the rbenv shim, but running `gem` uses the rvm installation.  And rvm can "infect" the rbenv installation with its own gem plugins, and other similar clashes.  This creates an unstable system with arcane error messages that can be incredibly difficult and frustrating to debug.

Even worse, the `rvm` documentation for how to uninstall `rvm` simply doesn't work inside devcontainers that use the ruby feature.  After the rvm files have been removed and various `/etc` files have been edited, the rbenv installation is still broken because the `devcontainer-feature.json` sets the `GEM_HOME` and `GEM_PATH` environment variables to rvm directories.

In situations (such as GitHub Codespaces) where the default image has both `rvm` and `rbenv` installed, this can run into trouble with user's dotfiles, which are often configured to check for `rbenv` and prefer it over `rvm`.

Unfortunately, setting `GEM_HOME` and `GEM_PATH` to empty strings is not the same as unsetting them.  And it isn't possible to simply unset Dockerfile ENV vars.  (Is it possible to unset `devcontainer.json` env vars?)

### Workarounds

We can manually override those variables to something else in our own devcontainer.json files, but that still effectively breaks `rbenv`:  `rbenv` uses shims to load the correct versions of `ruby` or `gem`, and those installed versions provide their own sensible _dynamic_ defaults.  This is especially useful when switching between branches with different ruby versions, when testing multiple ruby versions with a shell script that sets `RBENV_VERSION`, when using multiple repositories, or when multiple directories in a single repository have different .ruby-version files.  Setting `GEM_HOME` and `GEM_PATH` to a static path can break all of these scenarios.  

[...]

PR #764, opened 2023-11-24
Improving the Ruby Feature.

ThomasOwens:
This is a work-in-progress draft pull request demonstrating some improvements to the Ruby feature to address https://github.com/devcontainers/features/issues/603 and https://github.com/devcontainers/features/issues/757.

In the current state, the Ruby Feature now has parity with the Node Feature. The issue where multiple Ruby version managers have been installed has been solved. Currently, RVM is used. Although I prefer rbenv (and the poll seems to perhaps favor rbenv as well), I noticed the Node Feature uses NVM and doesn't provide an option for nodenv or any other Node version managers. Since Node is frequently used with Ruby for web development (Rails, Sinatra), it seemed to make since to make the first step resolving the key problem and making it mimic the Node Feature.


src/ruby/devcontainer-feature.json
```diff
@@ -1,21 +1,27 @@
 {
     "id": "ruby",
-    "version": "1.2.0",
-    "name": "Ruby (via rvm)",
+    "version": "1.2.0", // TODO - Is this a major or a minor update? Major, probably.
```
ThomasOwens, line 3: Need feedback on versioning. I suspect a major change here, given the changes to the options.

src/ruby/devcontainer-feature.json
```diff
@@ -1,21 +1,27 @@
...
+    "name": "Ruby",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/ruby",
     "description": "Installs Ruby, rvm, rbenv, common Ruby utilities, and needed dependencies.",
     "options": {
         "version": {
             "type": "string",
             "proposals": [
                 "latest",
+                "os-provided",
```
ThomasOwens, line 12: Still need to implement the use of the system Ruby, if it exists.

src/ruby/install.sh
```diff
@@ -7,26 +7,21 @@
...
-# Note: ruby-debug-ide will install the right version of debase if missing and
-# installing debase directly fails on Ruby 3.1.0 as of 1/7/2022, so omitting.
-# installing ruby-debug-ide on debian fails, so omitting.
-DEFAULT_GEMS="rake"
+GPG_KEY_SERVERS="keyserver hkps://keys.openpgp.org
+keyserver hkp://keyserver.pgp.com
+keyserver hkp://pgp.mit.edu
+keyserver hkp://ipv4.pool.sks-keyservers.net
+keyserver hkp://keyserver.ubuntu.com
+keyserver hkp://keyserver.ubuntu.com:80"
```
ThomasOwens, line 21: I added the keyservers recommended in the RVM documentation in addition to the ones previously included.

test/ruby/scenarios.json
```diff
@@ -1,17 +1,36 @@
 {
-    "install_additional_ruby": {
-        "image": "ubuntu:focal",
+    "non_root_user": {
+        "image": "mcr.microsoft.com/devcontainers/base:bullseye",
```
ThomasOwens, line 3: I noticed that the Ruby image used a Ruby Docker image as its base. That seemed strange to me. Not sure if it impacts the test scenarios, though. The Node image also uses a Node Docker image as its base, but the Node Feature tests test on Microsoft base devcontainer images. Perhaps a broader concern?

test/ruby/scenarios.json
ThomasOwens, line 1: Could use some ideas on additional test coverage.

src/ruby/devcontainer-feature.json
```diff
@@ -10,12 +10,17 @@
             "proposals": [
```
ThomasOwens, line 10: Implementing an os-provided Ruby is difficult.

The mcr.microsoft.com/devcontainers/base:bullseye image does not include Ruby. The Ruby devcontainer image (`images/src/ruby/.devcontainer/Dockerfile` in the devcontainer/images repository) is based upon a Ruby image from the official Ruby Docker images and not an image in the Microsoft container registry, which does include an installed Ruby. This is why my (not yet committed) test and change to handle this os-provided option isn't working.

I'm even hesitant to include a "none" option. But I did aim for parity with the Node Feature, and it offers a none option which just installs NVM. So I carried that option over here. I did notice, though, that Node also has the issue where their devcontainer image is based on an official Node docker image and not an image in the MCR.

nevans:
Copied from https://github.com/devcontainers/features/issues/757#issuecomment-1860734245, although I did make a few additional (mostly minor) edits to this comment:

@ThomasOwens Thank you so much for taking this on!  To summarize my current thoughts here:

As an incremental improvement, this is fine.  But if we are going to bump the version number for backward incompatibility, I personally vote for delaying until we can make `rvm` fully optional.  This is only a minor improvement over the current version: `rbenv` is unusable now, and (after this PR, in its current form) `rbenv` will still be unusable without _more_ backward incompatible changes.

I _personally_ consider the "minimum feature set" to be:
* Don't set `GEM_HOME` or `GEM_PATH` _at all_.  They **will** screw up alternate version managers or other customization added by the user (either before or after this feature).  _(Adding `rvm` or `rbenv` specific directories to `PATH` should be just fine.)_
* Support `os-provided`, at least for `apt`-based distributions.  Perhaps the code for this can be copied from other features.
* Use the version of ruby that's already on the `PATH`, if the version number matches the requested version.  _This allows the ruby feature to cooperate with the official docker hub `ruby` image or the user's own custom base image._  Perhaps the code for this can be copied from other features.
* When the version of the available ruby does _not_ match the requested version, install the requested version _somehow_ that won't break (existing or future) standard installs of `rvm` or `rbenv`.  _How_ it's installed is not part of my "minimum feature set", so long as my env vars constraint is met.  🙂

My recommendation on how to satisfy my "minimum feature set":
* Do _not_ install either `rvm` or `rbenv` (unless requested by the user by a config option).
* _Do_ install [`ruby-build`](https://github.com/rbenv/ruby-build) to `/usr/local`, unless it's already installed on the `PATH` or (perhaps) disabled by a config option.  _As far as I can tell, this is the **only** version management tool that is actively supported by the core teams of CRuby, JRuby, and TruffleRuby._  It works just fine without `rbenv`.  And, even if it is unused, it isn't large and installing it shouldn't hurt anything.
* When necessary, install from source using `ruby-build` to either `/opt/ruby` or `/usr/local`.  Ensure the default `PATH` can find all of ruby's bins.

I'm not sure if it should be part of the "minimum feature set", but I think we should probably do _one_ of the following, to aid with backward-incompatible upgrade pains:
* Ensure that the user _can_ use `rvm`, even with `GEM_HOME` and `GEM_PATH` unset.  Check the feature against at least a two base images: the default `devcontainer` base image, and the default docker hub `ruby` image.  I _suspect_ `rvm` might work without those env vars on the _default_ devcontainer images, but it might break if the user bases the feature on the docker hub `ruby` image (which _also_ sets these env vars!!!).
* Document that users may need to set these environment variables themselves to make everything play nicely with `rvm`.

None of the following ar.e in my _personal_ minimum feature set... but they _would_ all be _nice_ to have:
* _If_ you make the install location user-configurable, document that variable as _only_ applying to the ruby installed by `ruby-build` (i.e: not the os-provided version, nor to any version installed via rvm or rbenv).
* If the OS provided version matches the requested version but isn't installed, install it (at least for `apt` distros).  The version detection code in other official features can probably be copied to accomplish this.
* A config option to install a version manager: `none` or `rvm` or `rbenv` (or `chruby`).
* Install via `rvm` when `rvm` is installed (by the feature, or already on PATH).
* Install via `rbenv install` when `rbenv` is installed (by the feature, or already on PATH).
* If we need (or want) any further customization to how our default installation of rubygems behaves, use their recommended 'operating_system.rb' and not environment variables.  Environment variables are great for _overrides_, but (IMO) they are unsuitable for defaults.

What do you think?  _(My apologies: I'll probably be too busy with other end-of-year stuff both at work and in my personal time, and I probably can't really offer any more than my criticism right now.)_

PR #769, opened 2023-11-30
Update Oryx feature to install if a newer version of .NET is present

seesharprun:
Oryx fails to install if a newer version of .NET is present.

I suspect this is due to a missing `--framework` argument.

I also took this opportunity to spell out the full name of other `dotnet build` and `dotnet publish` arguments to make it easier to debug for the person that comes next.

### DIFF
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -133,12 +133,12 @@ if dotnet --version > /dev/null ; then
 fi
 
 # Oryx needs to be built with .NET 7
-if [[ "${DOTNET_BINARY}" = "" ]] || [[ "$(dotnet --version)" != *"7"* ]] ; then
-    echo "'dotnet 7' was not detected. Attempting to install .NET 7 to build oryx."
+if [[ "${DOTNET_BINARY}" = "" ]] || [[ "$(dotnet --list-sdks)" != *"7.0."* ]] ; then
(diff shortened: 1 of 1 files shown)


seesharprun:
This PR is an attempt to resolve devcontainers/images#855

seesharprun:
There is still some more work that needs to be done. Even on my local machine, I can't build Oryx from source if I have both .NET 7 and .NET 8 installed.

seesharprun:
It looks like the feature runs `dotnet --version` which will always return the latest version of .NET.

Instead it should check `dotnet --list-sdks` to see if .NET 7 is included in the list.

seesharprun:
A little more investigating has uncovered that some of the assemblies (Ex: Microsoft.NET.Sdk.Razor/source-generators/Microsoft.NET.Sdk.Razor.SourceGenerators.dll) will always use the latest .NET version if multiple versions are enabled.

I looked around on GitHub and found a couple of example issues where the .NET team says this is intentional behavior. They don't want someone to "freeze" on an old dependency and lock it in for years after moving on to newer versions of .NET.

samruddhikhandale:
Feel free to close this in favor of https://github.com/devcontainers/features/pull/825 which should unblock  https://github.com/devcontainers/images/pull/855

PR #775, opened 2023-12-06
Fix `.zprofile` is owned by root

joshspicer:
We got a report that the `remoteUser`'s .zprofile was owned by root, which was causing subsequent tools to fail to write to that file (eg: updating PATH).  Along the way noticed some interesting changes introduces in #736 that I think might be the cause, or at least appears to be an anti-pattern to me.

This change reverts [a previous change](https://github.com/devcontainers/features/pull/736/files#diff-3e71c0a0669a0410f7dd0d8f2b83f3b6bf6b525d3eabd354f19f32822da669fcR460-R463).  I don't think we should be sourcing a `.profile` at all in `.zprofile` (mixing bash and zsh scripts).  There's perhaps some extra configuration that we need to port over to `.zprofile`, which should be reviewed and added in a follow up change.

context/ [slack](https://github.slack.com/archives/C9XJ6156E/p1701898167089739)

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.3.0",
+    "version": "2.3.1",
     "name": "Common Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",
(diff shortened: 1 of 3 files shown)


joshspicer:
Yea, agreed after some more testing. 

Updated to add an empty .zprofile if there isn't one already, and assert that the file is owned by remoteUser.  Make sense to you @markphip ?

markphip:
Yes, I think that is reasonable. sourcing .profile likely needs to be revisited because it seems likely there will be scenarios that does not work but I'd imagine @samruddhikhandale  tested that manually and it at least is working in the simple scenarios where the file can be handled by zsh.

The file being owned by root was the main problem introduced here

eljog (approved):

PR #782, opened 2023-12-11
Add RHEL support to git feature

### LINKED ISSUE #781: Add RHEL support to git feature
Add RHEL support to git feature. Use centos7, almalinux 8 and almalinux 9 to test. Include ability to build feature on alma minimal base images.

PR to address this issue is forthcoming.

jdputschadi:
This PR adds RHEL support to git feature.

CentOS 7, RHEL/Alma/Rocky Linux 8 & 9 should all work. Testing was done with centos7, almalinux 8 and almalinux 9.

Includes ability to build feature on alma minimal base images.

resolves #781

### DIFF
--- a/src/git/NOTES.md
+++ b/src/git/NOTES.md
@@ -2,6 +2,6 @@
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
+This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, Alma, and RockyLinux distributions with the `apt`, `yum`, or `dnf` package manager installed.
 
 `bash` is required to execute the `install.sh` script.
--- a/src/git/README.md
+++ b/src/git/README.md
@@ -22,7 +22,7 @@ Install an up-to-date version of Git, built from source as needed. Useful for wh
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
+This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, Alma, and RockyLinux distributions with the `apt`, `yum`, or `dnf` package manager installed.
 
 `bash` is required to execute the `install.sh` script.
 
--- a/src/git/install.sh
+++ b/src/git/install.sh
@@ -16,15 +16,47 @@ keyserver hkp://keyserver.ubuntu.com:80
 keyserver hkps://keys.openpgp.org
 keyserver hkp://keyserver.pgp.com"
 
-set -e
-
-# Clean up
-rm -rf /var/lib/apt/lists/*
-
 if [ "$(id -u)" -ne 0 ]; then
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
 fi
+
+# Bring in ID, ID_LIKE, VERSION_ID, VERSION_CODENAME
+. /etc/os-release
+# Get an adjusted ID independent of distro variants
+if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = "debian" ]; then
+    ADJUSTED_ID="debian"
+elif [[ "${ID}" = "rhel" || "${ID}" = "fedora" || "${ID}" = "mariner" || "${ID_LIKE}" = *"rhel"* || "${ID_LIKE}" = *"fedora"* || "${ID_LIKE}" = *"mariner"* ]]; then
+    ADJUSTED_ID="rhel"
+    VERSION_CODENAME="${ID}{$VERSION_ID}"
+else
+    echo "Linux distro ${ID} not supported."
+    exit 1
+fi
+
+if [ ${ADJUSTED_ID} = "rhel" ]; then
+    if type dnf > /dev/null 2>&1; then
+        INSTALL_CMD=dnf
+    elif type microdnf > /dev/null 2>&1; then
+        INSTALL_CMD=microdnf
+    else
+        INSTALL_CMD=yum
+    fi
+fi
+
+# Clean up
+clean_up() {
+    case $ADJUSTED_ID in
+        debian)
+            rm -rf /var/lib/apt/lists/*
+            ;;
+        rhel)
+            rm -rf /var/cache/dnf/*
+            rm -rf /var/cache/yum/*
+            ;;
+    esac
+}
+clean_up
 
 # Import the specified key in a variable name passed in as 
 receive_gpg_keys() {
@@ -61,40 +93,66 @@ receive_gpg_keys() {
     fi
 }
 
-apt_get_update()
-{
-    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
-        echo "Running apt-get update..."
-        apt-get update -y
+pkg_mgr_update() {
+    if type apt >/dev/null 2>&1; then
+        if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
+            echo "Running apt-get update..."
+            apt-get update -y
+        fi
+    elif [ ${INSTALL_CMD} = "dnf" ] || [ ${INSTALL_CMD} = "yum" ]; then
+        if [ "$(find /var/cache/${INSTALL_CMD}/* | wc -l)" = "0" ]; then
+            echo "Running ${INSTALL_CMD} check-update ..."
+            ${INSTALL_CMD} check-update
+        fi
     fi
 }
 
+
 # Checks if packages are installed and installs them if not
 check_packages() {
-    if ! dpkg -s "$@" > /dev/null 2>&1; then
-        apt_get_update
-        apt-get -y install --no-install-recommends "$@"
+    if type apt >/dev/null 2>&1; then
+        if ! dpkg -s "$@" > /dev/null 2>&1; then
+            pkg_mgr_update
+            apt-get -y install --no-install-recommends "$@"
+        fi
+    elif [ ${INSTALL_CMD} = "dnf" ] || [ ${INSTALL_CMD} = "yum" ]; then
+        _num_pkgs=$(echo "$@" | tr ' ' \\012 | wc -l)
+        _num_installed=$(${INSTALL_CMD} -C list installed "$@" | sed '1,/^Installed/d' | wc -l)
+        if [ ${_num_pkgs} != ${_num_installed} ]; then
+            pkg_mgr_update
+            ${INSTALL_CMD} -y install "$@"
+        fi
+    elif [ ${INSTALL_CMD} = "microdnf" ]; then
+        ${INSTALL_CMD} -y install \
+            --refresh \
+            --best \
+            --nodocs \
+            --noplugins \
+            --setopt=install_weak_deps=0 \
+            "$@"
+    else
+        echo "Linux distro ${ID} not supported."
+        exit 1
     fi
 }
 
 export DEBIAN_FRONTEND=noninteractive
 
-# Source /etc/os-release to get OS info
-. /etc/os-release
+# Debian / Ubuntu packages
 
 # If the os provided version is "good enough", just install that.
 if [ ${GIT_VERSION} = "os-provided" ] || [ ${GIT_VERSION} = "system" ]; then
     if type git > /dev/null 2>&1; then
         echo "Detected existing system install: $(git version)"
         # Clean up
-        rm -rf /var/lib/apt/lists/*
+        clean_up
         exit 0
     fi
 
-    echo "Installing git from OS apt repository"
+    echo "Installing git from OS apt/dnf/microdnf/yum repository"
     check_packages git
     # Clean up
-    rm -rf /var/lib/apt/lists/*
+    clean_up
     exit 0
 fi
 
@@ -112,7 +170,35 @@ if ([ "${GIT_VERSION}" = "latest" ] || [ "${GIT_VERSION}" = "lts" ] || [ "${GIT_
 fi
 
 # Install required packages to build if missing
-check_packages build-essential curl ca-certificates tar gettext libssl-dev zlib1g-dev libcurl?-openssl-dev libexpat1-dev
+if [ "${ADJUSTED_ID}" = "debian" ]; then
+
+    check_packages build-essential curl ca-certificates tar gettext libssl-dev zlib1g-dev libcurl?-openssl-dev libexpat1-dev
+
+    check_packages libpcre2-dev
+
+    if [ "${VERSION_CODENAME}" = "focal" ] || [ "${VERSION_CODENAME}" = "bullseye" ]; then
+        check_packages libpcre2-posix2
+    elif [ "${VERSION_CODENAME}" = "bionic" ] || [ "${VERSION_CODENAME}" = "buster" ]; then
+        check_packages libpcre2-posix0
+    else
+        check_packages libpcre2-posix3
+    fi
+
+elif [ "${ADJUSTED_ID}" = "rhel" ]; then
+
+    if [ $VERSION_CODENAME = "centos7" ]; then
+        check_packages centos-release-scl 
+        check_packages devtoolset-11
+        source /opt/rh/devtoolset-11/enable
+    else
+        check_packages gcc
+    fi
+
+    check_packages libcurl-devel expat-devel gettext-devel openssl-devel perl-devel zlib-devel cmake pcre2-devel tar gzip
+    if [ $VERSION_ID -lt "9" ]; then
+        check_packages curl
+    fi
+fi
 
 # Partial version matching
 if [ "$(echo "${GIT_VERSION}" | grep -o '\.' | wc -l)" != "2" ]; then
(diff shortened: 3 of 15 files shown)


src/git/install.sh
```diff
@@ -61,40 +93,66 @@ receive_gpg_keys() {
...
     if type git > /dev/null 2>&1; then
         echo "Detected existing system install: $(git version)"
         # Clean up
-        rm -rf /var/lib/apt/lists/*
+        clean_up
         exit 0
     fi
 
-    echo "Installing git from OS apt repository"
+    echo "Installing git from OS apt/dnf/microdnf/yum repository"
```
eljog, line 152: Should we do `echo "Installing git from ${INSTALL_CMD} repository"`

src/git/install.sh
```diff
@@ -61,40 +93,66 @@ receive_gpg_keys() {
...
     fi
 }
 
+
 # Checks if packages are installed and installs them if not
 check_packages() {
-    if ! dpkg -s "$@" > /dev/null 2>&1; then
-        apt_get_update
-        apt-get -y install --no-install-recommends "$@"
+    if type apt >/dev/null 2>&1; then
```
eljog, line 113: Should we use `INSTALL_CMD` in all cases for consistency?

eljog (commented):
Thank you for the PR. We really appreciate the time you have taken to contribute to the community.
- I have added a few comments
- Please [accept the CLA](https://github.com/devcontainers/features/pull/782#issuecomment-1851054314)

src/git/install.sh
jdputschadi: Added commit to use `INSTALL_CMD` in all cases.

src/git/install.sh
jdputschadi: Added commit to echo either "apt" or "yum/dnf" based on `INSTALL_CMD`. Yum and DNF repositories are, as best as I can tell, the same thing. Yum and DNF are just package managers using the same underlying repository setup.

jdputschadi:
CLA has been submitted for review by our Legal team. I'm working to get it approved and signed as soon as possible.

src/git/install.sh
```diff
@@ -61,40 +93,70 @@ receive_gpg_keys() {
     fi
 }
 
-apt_get_update()
-{
-    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
-        echo "Running apt-get update..."
-        apt-get update -y
+pkg_mgr_update() {
+    if type apt >/dev/null 2>&1; then
```
eljog, line 97: Should this be, same as your check inside `check_packages()`:
```suggestion
    if [ ${INSTALL_CMD} = "apt-get" ]; then
```

jdputschadi: Done.

jdputschadi:
@microsoft-github-policy-service agree company="Analog Devices"

src/git/NOTES.md
```diff
@@ -2,6 +2,6 @@
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
+This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, Alma, and RockyLinux distributions with the `apt`, `yum`, or `dnf` package manager installed.
```
samruddhikhandale, line 5: 🎉 🎉

src/git/README.md
```diff
@@ -22,7 +22,7 @@ Install an up-to-date version of Git, built from source as needed. Useful for wh
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
```
samruddhikhandale, line 25: Once this PR is merged, https://github.com/devcontainers/action will run automatically ; which opens a PR that updates the README file based on `devcontainer-feature.json` and `NOTES.md`.

Hence, can we revert the changes to this file to avoid merge conflicts? Thanks!

src/git/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "git",
-    "version": "1.1.6",
+    "version": "1.1.7",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.2.0",
```

Can we bump the minor version as this is a nice and significant change?

src/git/install.sh
```diff
@@ -16,16 +16,48 @@ keyserver hkp://keyserver.ubuntu.com:80
...
+
+if type apt-get > /dev/null 2>&1; then
+    INSTALL_CMD=apt-get
+elif type dnf > /dev/null 2>&1; then
+    INSTALL_CMD=dnf
+elif type microdnf > /dev/null 2>&1; then
+    INSTALL_CMD=microdnf
+else
+    INSTALL_CMD=yum
+fi
```
samruddhikhandale, line 45:
```suggestion
elif type yum > /dev/null 2>&1; then
    INSTALL_CMD=yum
else
    echo "Error...."
fi
```

Just for sanity... 😄

src/git/install.sh
```diff
@@ -61,40 +93,70 @@ receive_gpg_keys() {
...
+        if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
+            echo "Running apt-get update..."
+            ${INSTALL_CMD} update -y
+        fi
+    elif [ ${INSTALL_CMD} = "dnf" ] || [ ${INSTALL_CMD} = "yum" ]; then
+        if [ "$(find /var/cache/${INSTALL_CMD}/* | wc -l)" = "0" ]; then
+            echo "Running ${INSTALL_CMD} check-update ..."
+            ${INSTALL_CMD} check-update
+        fi
     fi
```
samruddhikhandale, line 107: Should we also update packages using `microdnf` ?

src/git/install.sh
```diff
@@ -16,16 +16,48 @@ keyserver hkp://keyserver.ubuntu.com:80
...
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
 fi
 
+# Bring in ID, ID_LIKE, VERSION_ID, VERSION_CODENAME
+. /etc/os-release
+# Get an adjusted ID independent of distro variants
+if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = "debian" ]; then
+    ADJUSTED_ID="debian"
+elif [[ "${ID}" = "rhel" || "${ID}" = "fedora" || "${ID}" = "mariner" || "${ID_LIKE}" = *"rhel"* || "${ID_LIKE}" = *"fedora"* || "${ID_LIKE}" = *"mariner"* ]]; then
```
samruddhikhandale, line 29: With respect to https://github.com/devcontainers/features/pull/782/files#diff-0aeb158b92b97b97e3e82b987e9a296326133791101116ce40f2e33bea021711R5, should this condition also check for `almalinux` ? 🤔

src/git/NOTES.md
samruddhikhandale: Should this mention about `microdnf` package manager as well?

src/git/install.sh
```diff
@@ -16,16 +16,48 @@ keyserver hkp://keyserver.ubuntu.com:80
...
+else
+    echo "Linux distro ${ID} not supported."
+    exit 1
+fi
+
+if type apt-get > /dev/null 2>&1; then
+    INSTALL_CMD=apt-get
+elif type dnf > /dev/null 2>&1; then
+    INSTALL_CMD=dnf
+elif type microdnf > /dev/null 2>&1; then
```
samruddhikhandale, line 41: With my limited knowledge of Fedora distros, I believe Fedora-based minimal containers can have both `dnf` and `microdnf` package managers present. 

In this case, do we want to give priority to `microdnf` as the user might be looking for slim down containers (ie. should the check for `microdnf` precede the `dnf` check? Curious on what you think 💭

test/common-utils/alma-8.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
samruddhikhandale, line 1: Hmm, why don't I see test action runs for this PR? 😅  Taking a look!

test/git/scenarios.json
```diff
@@ -43,5 +43,50 @@
...
+    "install_git_from_src_alma-8-minimal": {
+        "image": "almalinux:8-minimal",
+        "features": {
+            "git": {
+                "version": "latest",
+                "ppa": "false"
+            }
+        }
+    },
+    "install_git_from_src_alma-9-minimal": {
```
samruddhikhandale, line 83: Appreciate all these test scenarios, thank you! 

Can we also add similar scenarios for other distros this PR supports? (eg. fedora, mariner, rocky etc). Feel free to refer to https://github.com/devcontainers/features/blob/main/test/common-utils/scenarios.json

samruddhikhandale (changes requested):
Thank you so much for taking the time in contributing this PR, we really appreciate it 🙌 
This is going to be super helpful and useful for the community 🎉 

Left some questions/comments, thanks again!

test/common-utils/alma-8.sh
samruddhikhandale: @jdputschadi Do you mind if I close and open this PR? I hope the action should run as expected post that.

jdputschadi: I don't mind at all.

src/git/NOTES.md
jdputschadi: Done. commit with `microdnf` included coming soon.

src/git/devcontainer-feature.json
jdputschadi: Commit with change coming soon.

src/git/install.sh
jdputschadi: `microdnf` does not have a `check-update` command, so it cannot be included in this behvaior.

src/git/install.sh
jdputschadi: Quite happy to reverse the check/priority. I'll do so and run the tests, if they pass expect to see the switch in a commit/updated PR.

test/git/scenarios.json
jdputschadi: Will do.

src/git/README.md
jdputschadi: Will do.

src/git/install.sh
jdputschadi: The `"${ID_LIKE}" = *"rhel"*` clauses match `almalinux` and `rockylinux` flavors of RedHat.

src/git/README.md
jdputschadi: Commit `cba4c71` resolves this request.

src/git/install.sh
jdputschadi: Commit cba4c71 resolves this request.

src/git/install.sh
jdputschadi: Commit cba4c71 resolves this request.

test/git/scenarios.json
jdputschadi: Commit cba4c71 add requested tests and resolves this request.

test/git/scenarios.json
```diff
@@ -43,5 +43,86 @@
                 "ppa": "false"
             }
         }
+    },
+    "install_git_from_src_centos-7": {
+        "image": "centos:centos7",
+        "features": {
+            "git": {
+                "version": "latest",
+                "ppa": "false"
```
samruddhikhandale, line 52: Can we also add similar test scenarios with `ppa: true`? Thank you! 

This way we can also validate the code changes against 👇 

(TLDR: Make sure `git` [can be installed fine with the package managers](https://github.com/devcontainers/features/blob/3e96db3a6a28667bdc012ebecc3c798ddad6af99/src/git/install.sh#L172) With `ppa:false` `git` gets installed from source.)


https://github.com/devcontainers/features/blob/3e96db3a6a28667bdc012ebecc3c798ddad6af99/src/git/install.sh#L166-L176

samruddhikhandale (changes requested):
Thanks for resolving my comments so quickly ✨ 

Added few more minor comments, thanks!

test/common-utils/alma-8.sh
samruddhikhandale: Looks like the test ran this time 🎉

test/git/scenarios.json
jdputschadi: From https://launchpad.net/ubuntu/+ppas, Personal Package Archives (PPA) allow you to upload Ubuntu source packages to be built and published as an apt repository by Launchpad. You can find out more about PPAs and how to use them in our [help page](https://help.launchpad.net/Packaging/PPA).

Unfortunately PPAs are not for RedHat based systems so I cannot add these tests.

I'm not sure what the equivalent approach would be for RedHat systems: to my knowledge are no standard community package archives broadly accepted by corporations that use RedHat derived systems, and even those that exist tend not to keep broad ranges of versions of a package. 

I could add tests where `ppa=false` and `version=system`, which will use the system's package manager to pull the latest version available via the standard vendor (RedHat, Alma, Microsoft, Fedora, Rocky) repositories.

Would that be acceptable?

test/common-utils/scenarios.json
```diff
@@ -41,14 +41,42 @@
             "common-utils": {}
         }
     },
+    "alma-minimal-8": {
+        "image": "almalinux:8-minimal",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
```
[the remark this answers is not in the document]
jdputschadi: This is a failing test in the `common-utils` feature, not caused by the git feature. I can certainly take a look, but I think the fix for that test would be in a separate PR.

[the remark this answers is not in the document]
jdputschadi: The other reason for the failure is I've got a test on "minimal" images of AlmaLinux and RedHat Linux where yum is not installed, but microdnf is installed. It can be nice to use those as images for the basis of images/containers because you end up with a much smaller image in the end.

I see three choices here:
1. I remove the "minimal" image tests from the PR -- personally, I'd like to keep them because I want to know I can start with a minimal build.
2. I don't test with the `common-utils` feature included and rather update the "git" feature to install, if needed, more packages to get git built when needed.
3. I help fix the `common-utils` feature and add tests to it for the "minimal" images like I have for git.

I'm quite willing to file a ticket reflecting this problem, get it fixed and submit a PR against the ticket... basically taking approach #3.

test/git/scenarios.json
jdputschadi: Alternately, I could either set `ppa=true` in the scenarios or remove the `ppa=false` from the scenarios, and it still won't affect the build or test because PPAs only apply to Ubuntu.

src/git/install.sh
```diff
@@ -16,16 +16,51 @@ keyserver hkp://keyserver.ubuntu.com:80
...
-
 if [ "$(id -u)" -ne 0 ]; then
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
 fi
 
+# Bring in ID, ID_LIKE, VERSION_ID, VERSION_CODENAME
+. /etc/os-release
+# Get an adjusted ID independent of distro variants
+if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = "debian" ]; then
```
samruddhikhandale, line 27:
```suggestion
if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = "ubuntu" ]; then
```

samruddhikhandale: Looks like `common-utils` Feature need a similar change, but we can isolate and make changes in a separate PR. See https://github.com/devcontainers/features/blob/3e96db3a6a28667bdc012ebecc3c798ddad6af99/src/common-utils/main.sh#L335

test/git/scenarios.json
samruddhikhandale: > Unfortunately PPAs are not for RedHat based systems so I cannot add these tests.

Thanks for pointing that out, really helpful. 

Can we update the `ppa` option description to explicitly specify that it is only available for `Ubuntu`?  This way we can avoid such similar (future) confusions. (eg. `Install from PPA if available (only supported for Ubuntu distributions`)

https://github.com/devcontainers/features/blob/3e96db3a6a28667bdc012ebecc3c798ddad6af99/src/git/devcontainer-feature.json#L20

> I could add tests where ppa=false and version=system, which will use the system's package manager to pull the latest version available via the standard vendor (RedHat, Alma, Microsoft, Fedora, Rocky) repositories.
> Would that be acceptable?

Yep, that's a good enough test case and definitely makes sense. Thank you!

samruddhikhandale (changes requested):
One more comment, thanks!

test/common-utils/scenarios.json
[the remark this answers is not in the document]
samruddhikhandale: Appreciate your valuable thoughts and contributions  🙌 

> I'm quite willing to file a ticket reflecting this problem, get it fixed and submit a PR against the ticket... basically taking approach https://github.com/devcontainers/features/pull/3.

Like this approach, can we also remove the failing test scenarios this PR adds? 

>  personally, I'd like to keep them because I want to know I can start with a minimal build.

I understand your ask here, however, a failing build is not great and it stops the action runs, hence we will also get a ❌ for test runs. This can be misleading and we won't be able to check the complete health of `common-utils` Feature. Hence, I'd prefer if we remove the tests, and add them back with the fix PR. 

Let me know if you think otherwise, thanks!

test/git/scenarios.json
samruddhikhandale: > Alternately, I could either set ppa=true in the scenarios or remove the ppa=false from the scenarios, and it still won't affect the build or test because PPAs only apply to Ubuntu: see lines 165 & 166 of install.sh.

Yep, so ideally we need two test scenarios for each distro with `version=latest` and `version=system` (which is default ; so can be omitted too)

test/common-utils/scenarios.json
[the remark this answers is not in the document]
jdputschadi: I've submitted a PR, https://github.com/devcontainers/features/pull/811, to fix the common-utils issue, allowing the `common-utils` feature to work with minimal os image. That will fix the failing tests in this PR. Once https://github.com/devcontainers/features/pull/811 is merged, there should be no failing test scenarios in this PR.

I can remove the test scenarios from this PR and file a new PR later to add them back after my `common-utils` PR is merged.

src/git/install.sh
jdputschadi: This check is correct and does not need a PR. 

For Ubuntu images `ID=ubunto` and `ID_LIKE=debian`. 
For Debian images `ID_LIKE` is not set.

The purpose here is to recognize that Ubuntu should be treated like debian in the script, same for `common-utils`.

Jeff.

samruddhikhandale: Hah! Apologies, I definitely misread the code line.

test/git/scenarios.json
jdputschadi: RedHat, Alma, Microsoft, Fedora, Rocky for building from source (`ppa=false` and `version=latest`) as well as installing via system packages (`ppa=true` and `version=system`) are in place and should be passing as of commit 0bd3add1.

jdputschadi:
Samrhuddhi:

Thank you for all the feedback and input. I think I've addressed all your inputs and requested changes and improvements.

Please let me know if there is more I need to do.

Thanks,

Jeff.

samruddhikhandale (approved):
Beautiful, thank you so much ✨ 🙏

jdputschadi:
Samrhuddhi:

Thank you again for the thoughtful and helpful sugguestions.

Jeff.

samruddhikhandale:
@jdputschadi Let me know if you'd like to be added to the community slack channel. Instructions in https://github.com/orgs/devcontainers/discussions/3

PR #787, opened 2024-01-01
feature nix: add option useAttrPath to use packages attribute path

nohzafk:
Add an option `useAttrPath` to install packages with `nix-env -iA`

## Name conversion
If package name doesn't start with `nixpkgs.`, `nixpkgs.` will be added as prefix to pacakge name automatically.

## Compatibility
`useAttrPath` use `false` as default value, which won't break existing usage and remains compatibility.

## Precision and Reliability
Using `nix-env -iA` ensures the installation of the exact package specified by its attribute path. This precision eliminates any ambiguity that might arise from `nix-env --install`, which can install unintended packages if multiple matches are found for a given name or description.

## Consistency with NixOS Package Search
 The attribute paths used with `nix-env -iA` directly correspond to those listed on the NixOS package search site(https://search.nixos.org/packages). This consistency simplifies the process of finding and installing packages as per official documentation and resources.


nohzafk:

eljog:
Thank you for the contribution. This sounds like a great addition.
I have approved for the workflow runs.

nohzafk:
@eljog hello~ just ping to see how was the result, still waiting for one approving review to be merge 😃

src/nix/devcontainer-feature.json
```diff
@@ -21,6 +21,11 @@
             "default": "",
```
samruddhikhandale, line 21: As we are adding a new Feature option, can we bump the `minor` version of the Feature?

https://github.com/devcontainers/features/blob/da7abcca2b5986a3c2cb3ddf62e1067d2fa2a19f/src/nix/devcontainer-feature.json#L3

src/nix/devcontainer-feature.json
```diff
@@ -21,6 +21,11 @@
...
             "description": "Optional comma separated list of Nix packages to install in profile."
         },
+        "useAttrPath": {
```
samruddhikhandale, line 24:
```suggestion
        "useAttributePath": {
```

src/nix/install.sh
```diff
@@ -8,6 +8,7 @@ cd "${FEATURE_DIR}"
 VERSION="${VERSION:-"latest"}"
 MULTIUSER="${MULTIUSER:-"true"}"
 PACKAGES="${PACKAGES//,/ }"
+USEATTRPATH="${USEATTRPATH:-"false"}"
```
samruddhikhandale, line 11: This would need to be updated to sync with my previous comment

samruddhikhandale (changes requested):
Thank you for adding this super cool Feature option, appreciate it ✨ 

Left some minor comments, otherwise looks great!

src/nix/devcontainer-feature.json
nohzafk: bump to 1.2.0

src/nix/devcontainer-feature.json
nohzafk: updated

nohzafk:
@samruddhikhandale all fixed, thanks.

eljog (dismissed):
⚡

nohzafk:
sorry, forget to update the opiton name to useAttributePath in test. updated.

samruddhikhandale (approved):
🚢 Thanks!

Issue #789, opened 2024-01-03
docker-outside-of-docker socat proxy can't keep up with docker compose


groner:
Docker compose creates concurrent requests quicker than the socat proxy can accept them with it's single main accepting process/thread. The proxy socket's backlog eventually becomes full and docker compose gets errors and gives up.

```bash
: kai@merman poc; devcontainer exec --workspace-folder=. docker compose up -d
[+] Running 9/0
 ✔ Network poc_default                      Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-10  Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-5   Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-7   Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-1   Created                                                                                                                           0.0s
 ⠋ Container poc-parallel-sleep-service-6   Creating                                                                                                                          0.0s
 ✔ Container poc-parallel-sleep-service-2   Created                                                                                                                           0.0s
 ⠋ Container poc-parallel-sleep-service-3   Creating                                                                                                                          0.0s
 ✔ Container poc-parallel-sleep-service-4   Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-9   Created                                                                                                                           0.0s
 ✔ Container poc-parallel-sleep-service-8   Created                                                                                                                           0.0s
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

I get failures running `docker compose up`, `docker compose ps`, and `docker compose down`. Running `docker ps` always works.

This heisenbug likely occurs with different numbers of containers depending on the platform docker is running on. On my M1 MBP with 8+2 cores things break with around 9 or 10 containers in the compose stack.

I've created a gist with instructions to reproduce the problem.

https://gist.github.com/groner/a8e4bb8915895ad3c11a7d95f744ffce#file-readme-md

eljog:
Thank you for sharing the issue details and for the repro-steps.

@joshspicer do you have any thoughts on this? Does this look like a limitation with socat?

eljog:
Just noticed that you have a PR open for this issue: https://github.com/devcontainers/features/pull/790
We will take a look as soon as possible - Thank you

eljog:
PR #790 has been checked in. Hope that helps everyone.
Thank you @groner for contributing the fix ✨

PR #790, opened 2024-01-03
[Docker-outside-of-docker] Increase backlog depth for socat proxy (#789)

groner:
The underlying problem described in #789 is that the main `accept(2)` loop in `socat` is single threaded and is easily overwhelmed by docker compose, which can create connection attempts in multiple cores.

Fortunately the listening socket already has a backlog which we can increase from [socat's default of 5](http://www.dest-unreach.org/socat/doc/socat.html#GROUP_LISTEN).

In this PR I have chosen a depth of 128 as probably high enough for most devcontainer needs. If it isn't high enough this could be made tunable, but reevaluating the proxy implementation might be in order also.

### DIFF
--- a/src/docker-outside-of-docker/install.sh
+++ b/src/docker-outside-of-docker/install.sh
@@ -329,7 +329,7 @@ if [ "${ENABLE_NONROOT_DOCKER}" = "true" ] && [ "${SOURCE_SOCKET}" != "${TARGET_
             log "Enabling socket proxy."
             log "Proxying ${SOURCE_SOCKET} to ${TARGET_SOCKET} for vscode"
             sudoIf rm -rf ${TARGET_SOCKET}
-            (sudoIf socat UNIX-LISTEN:${TARGET_SOCKET},fork,mode=660,user=${USERNAME} UNIX-CONNECT:${SOURCE_SOCKET} 2>&1 | sudoIf tee -a \${SOCAT_LOG} > /dev/null & echo "\$!" | sudoIf tee \${SOCAT_PID} > /dev/null)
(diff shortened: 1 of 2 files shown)


eljog:
Thank you for the contribution.
I can approve to run the pipeline checks once the CLA has been accepted

groner:

groner:
@eljog Thanks for looking at this.

I've added some details to the description, but refer to #789 for context. I hope that's alright.

eljog:
I have approved to run the workflow checks

src/docker-outside-of-docker/install.sh
```diff
@@ -329,7 +329,7 @@ if [ "${ENABLE_NONROOT_DOCKER}" = "true" ] && [ "${SOURCE_SOCKET}" != "${TARGET_
             log "Enabling socket proxy."
             log "Proxying ${SOURCE_SOCKET} to ${TARGET_SOCKET} for vscode"
             sudoIf rm -rf ${TARGET_SOCKET}
-            (sudoIf socat UNIX-LISTEN:${TARGET_SOCKET},fork,mode=660,user=${USERNAME} UNIX-CONNECT:${SOURCE_SOCKET} 2>&1 | sudoIf tee -a \${SOCAT_LOG} > /dev/null & echo "\$!" | sudoIf tee \${SOCAT_PID} > /dev/null)
+            (sudoIf socat UNIX-LISTEN:${TARGET_SOCKET},fork,mode=660,user=${USERNAME},backlog=128 UNIX-CONNECT:${SOURCE_SOCKET} 2>&1 | sudoIf tee -a \${SOCAT_LOG} > /dev/null & echo "\$!" | sudoIf tee \${SOCAT_PID} > /dev/null)
```
eljog, line 332: LGTM
@joshspicer any concerns with raising the backlog to 128 (from default 5)?

eljog (approved):

src/docker-outside-of-docker/install.sh
joshspicer: No concern here, looks reasonable and passes the current checks!

PR #793, opened 2024-01-08
Alpine - Install libssl1.1 only if available

eljog:
Alpine repo for 3.19 does not include the outdated libssl1.1.
This is a blocker for upgrading alpine image to 3.19 https://github.com/devcontainers/images/issues/786, as the image uses on common-utils feature.

Error details:
https://github.com/devcontainers/images/actions/runs/7282516027/job/20100948870?pr=897#step:3:906

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -265,7 +265,6 @@ install_alpine_packages() {
             libstdc++ \
             krb5-libs \
             libintl \
-            libssl1.1 \
             lttng-ust \
             tzdata \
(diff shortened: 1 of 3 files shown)


joshspicer (approved):
Nice! I wonder if we should also mirror installing libssl 3 (like we do on the debian side) in case alpine drops libssl 3

eljog:
> I wonder if we should also mirror installing libssl 3

Since we don't install that explicitly today, I think we can leave it as it is, unless a need arises in the future.

PR #795, opened 2024-01-09
add `shared` config to build Python shared library

adehad:
I was trying to use this devcontainer with PyInstaller, and it seems like the build fails due to the Python installation not built with the shared libraries.

## Summary of Changes
1. Provides a new option to control the build flag for the shared libraries

### Concerns / Thoughts
1. Do we want to preserve the original behaviour (ie have the option as false) 
2. Do we want to do something special regarding the MacOS specific commands

## Details

Here are the relevant logs

```
    run()
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/__main__.py", line 198, in run
    run_build(pyi_config, spec_file, **vars(args))
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/__main__.py", line 69, in run_build
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/building/build_main.py", line 1071, in main
    build(specfile, distpath, workpath, clean_build)
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/building/build_main.py", line 1011, in build
    exec(code, spec_namespace)
  File "app.spec", line 13, in <module>
    a = Analysis(
        ^^^^^^^^^
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/building/build_main.py", line 470, in __init__
    self.__postinit__()
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/building/datastruct.py", line 184, in __postinit__
    self.assemble()
  File "/workspaces/pyside6-boilerplate/.venv/lib/python3.12/site-packages/PyInstaller/building/build_main.py", line 581, in assemble
    raise PythonLibraryNotFoundError()
PyInstaller.exceptions.PythonLibraryNotFoundError: Python library not found: libpython3.12.so, libpython3.12.so.1.0
    This means your Python installation does not come with proper shared library files.
    This usually happens due to missing development package, or unsuitable build parameters of the Python installation.

    * On Debian/Ubuntu, you need to install Python development packages:
      * apt-get install python3-dev
      * apt-get install python-dev
    * If you are building Python by yourself, rebuild with `--enable-shared` (or, `--enable-framework` on macOS).
```

### DIFF
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -10,6 +10,7 @@
 PYTHON_VERSION="${VERSION:-"latest"}" # 'system' or 'os-provided' checks the base image first, else installs 'latest'
 INSTALL_PYTHON_TOOLS="${INSTALLTOOLS:-"true"}"
 OPTIMIZE_BUILD_FROM_SOURCE="${OPTIMIZE:-"false"}"
+SHARED_LIBS="${SHARED:-"false"}"
 PYTHON_INSTALL_PATH="${INSTALLPATH:-"/usr/local/python"}"
 OVERRIDE_DEFAULT_VERSION="${OVERRIDEDEFAULTVERSION:-"true"}"
 
@@ -264,7 +265,11 @@ install_from_source() {
     if [ "${OPTIMIZE_BUILD_FROM_SOURCE}" = "true" ]; then
         config_args="--enable-optimizations"
     fi
-    ./configure --prefix="${INSTALL_PATH}" --with-ensurepip=install ${config_args}
+    local shared_libs=""
+    if [ "${SHARED_LIBS}" = "true" ]; then
+        shared_libs="--enable-shared"
+    fi
+    ./configure --prefix="${INSTALL_PATH}" --with-ensurepip=install ${config_args} ${shared_libs}
     make -j 8
     make install
     cd /tmp
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -32,6 +32,11 @@
       "default": false,
       "description": "Optimize Python for performance when compiled (slow)"
     },
+    "shared": {
+      "type": "boolean",
+      "default": false,
+      "description": "Build Python with the shared library files."
+    },
     "installPath": {
       "type": "string",
       "default": "/usr/local/python",


adehad:

src/python/install.sh
```diff
@@ -264,7 +265,11 @@ install_from_source() {
     if [ "${OPTIMIZE_BUILD_FROM_SOURCE}" = "true" ]; then
         config_args="--enable-optimizations"
     fi
-    ./configure --prefix="${INSTALL_PATH}" --with-ensurepip=install ${config_args}
+    local shared_libs=""
+    if [ "${SHARED_LIBS}" = "true" ]; then
+        shared_libs="--enable-shared"
+    fi
+    ./configure --prefix="${INSTALL_PATH}" --with-ensurepip=install ${config_args} ${shared_libs}
```
adehad, line 272:
```suggestion
    ./configure --prefix="${INSTALL_PATH}" --with-ensurepip=install ${config_args} --enable-shared
```

Perhaps this is a feature we want by default and no configuration option is needed.

joshspicer: Do you happen to know the performance consequences of adding `--enable-shared`?  I suspect that this would be to have as a default (perhaps inverting the option to let us disable) as long as it doesn't significantly increase the (already long) compile time

joshspicer:
Would you mind adding a [scenario test](https://github.com/devcontainers/features/blob/main/test/python/scenarios.json#L84) to exercise your new flag ([docs](https://github.com/devcontainers/cli/blob/main/docs/features/test.md#scenarios))

src/python/install.sh
adehad: Unfortunately I do not, the 'default' Python Docker compilation includes other flags that might improve performance however?

https://github.com/docker-library/python/blob/c007698b9f91bedba5e6300d9394c10acd194dcb/3.11/bookworm/Dockerfile#L44-L53

adehad:
> Would you mind adding a [scenario test](https://github.com/devcontainers/features/blob/main/test/python/scenarios.json#L84) to exercise your new flag ([docs](https://github.com/devcontainers/cli/blob/main/docs/features/test.md#scenarios))?

Of course, my bad for missing this.

any preference for using an `ldd` command (nix specific) 

```
root@c098cfd1b2e0:/workspaces/python# ldd $(which python)
        linux-vdso.so.1 (0x00007fff0a3fb000)
        libpython3.12.so.1.0 => /usr/local/bin/../lib/libpython3.12.so.1.0 (0x00007f43580d5000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4357ef8000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f4357ed6000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f4357ed0000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f4357ecb000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f4357d87000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f4358706000)
``` 


When not built with shared looks like:
```
['/usr/local/python/3.10.13/lib', 'libpython3.10.a']
```

or a Python only
```
root@c098cfd1b2e0:/workspaces/python# python -c "import sysconfig; print(sysconfig.get_config_vars('LIBDIR', 'INSTSONAME'))" 
['/usr/local/lib', 'libpython3.12.so.1.0']
```

I'll grep either with option with ` | grep 'libpython3.*so'` in the check

PR #802, opened 2024-01-22
[Dotnet] - Fix tests ; removes stale check

samruddhikhandale:
Ref: https://github.com/devcontainers/features/actions/runs/7615173541/job/20739065620

Removes checks which are dynamic and prone to fail when new version becomes GA.

### DIFF
--- a/test/dotnet/install_aspnetcore_runtime_only.sh
+++ b/test/dotnet/install_aspnetcore_runtime_only.sh
@@ -18,10 +18,6 @@ expected=$(fetch_latest_version "aspnetcore")
 check "Latest ASP.NET Core Runtime version installed" \
 is_aspnetcore_runtime_version_installed "$expected"
 
-# Expect this check to fail in November 2023 when .NET 8.0 becomes GA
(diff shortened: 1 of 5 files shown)


test/dotnet/install_aspnetcore_runtime_only.sh
```diff
@@ -18,10 +18,6 @@ expected=$(fetch_latest_version "aspnetcore")
 check "Latest ASP.NET Core Runtime version installed" \
 is_aspnetcore_runtime_version_installed "$expected"
 
-# Expect this check to fail in November 2023 when .NET 8.0 becomes GA
-check "It is a flavor of .NET 7.0" \
```
samruddhikhandale, line 22: Not super confident if removing checks is the right way to go, we can also update it to .Net 8.0 (which would again fail in Nov 2024)

Appreciate any thoughts!

joshspicer: I think it's ok to remove

joshspicer (approved):

PR #804, opened 2024-01-24
Add support for including PowerShell profile

nebula-it:


src/powershell/install.sh
```diff
@@ -162,6 +163,12 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
     done
 fi
 
+# If URL for powershell profile is provided, download it to '/opt/microsoft/powershell/7/profile.ps1'
+if [ -n "$POWERSHELL_PROFILE_URL" ]; then
```
samruddhikhandale, line 167: I like the idea of being able to configure the Powershell profile with the Powershell Feature. Can we convert this to a [Feature option](https://containers.dev/implementors/features/#options-property)?

I'd expect it to be a `string` and called something like `powershellProfileURL` which is an optional param set to empty string.

src/powershell/install.sh
```diff
@@ -162,6 +163,12 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
...
+    echo "Downloading PowerShell Profile from: $POWERSHELL_PROFILE_URL"
+    curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"
```
samruddhikhandale, line 169: Any thoughts if we should configure it for all users vs current user? Looks like the current changes are configuring for all users.

I wonder if we need more community feedbacks for this one!

src/powershell/install.sh
```diff
@@ -162,6 +163,12 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
...
+fi
```
samruddhikhandale, line 170: Also, can you help me understand your motivation behind this change? I have never used Powershell profiles before, and I don't have my own script (hence, this question). 

Do folks like to publish their setting to some location? or are there official scripts available which you like to use?

src/powershell/install.sh
samruddhikhandale, line 169: > /opt/microsoft/powershell/7/profile.ps1

Looks like ^ would work for Powershell version 7, however, how about other versions? Like 5.1, doesn't seem like `/opt/microsoft/powershell/7/profile.ps1` would work. See https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-5.1

samruddhikhandale (changes requested):
Thank you for taking the time in contributing the changes.

I have left some comments and clarifying questions, thanks!

src/powershell/install.sh
nebula-it: Done :)

src/powershell/install.sh
nebula-it: The dev containers are mostly intended for single users and for shared users so I thought it should be fine to add it at system level. 
Another reason I chose above path is that it remains static v/s a new user might be created inside container at a later stage of dev container setup.
If someone really needs to run pwsh without profile they can launch it using `pwsh -noprofile`. But I'm happy to hear community feedback on this.

src/powershell/install.sh
nebula-it: My PowerShell profile holds a lot of customization (e.g prompt to show git status, custom aliases and functions) and useful utilities like [Completion Predictor](https://github.com/PowerShell/CompletionPredictor) and tab auto completer for lots of tools.

Folks usually publish their profiles to Github. So there are no official scripts etc, just personal profiles usually from Github but could be anywhere on a webserver as long as its publicly accessible.

src/powershell/install.sh
nebula-it: PowerShell 5.1 is not available for Linux and PowerShell 6 is deprecated. Only 7.* is [currently supported](https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.4), so this should cover most use cases for now. We can expand it later to support other version as well, if needed.

samruddhikhandale: That makes sense, thanks for you explanation.

src/powershell/devcontainer-feature.json
```diff
@@ -19,6 +19,11 @@
             "type": "string",
             "default": "",
             "description": "Optional comma separated list of PowerShell modules to install."
+        },
+        "powershellProfileURL ": {
+            "type": "string",
+            "default": "",
+            "description": "Optional URL to download PowerShell profile."
```
samruddhikhandale, line 26:
```suggestion
            "description": "Optional (publicly accessible) URL to download PowerShell profile."
```

src/powershell/install.sh
samruddhikhandale, line 167: Can we add test scenarios to validate these PR changes? Feel free to provide me a script, and I can host it publicly which will be accessible to the test scenario. This is good for maintainability and that way you don't have to maintain/support it long term.

samruddhikhandale (changes requested):
Thank you for answering my questions, appreciate it. Left few more comments!

src/powershell/install.sh
nebula-it: Yep for sure, I do not have a public profile atm. For something to test you can use https://github.com/ChrisTitusTech/powershell-profile , the value you will need to use for this parameter is `https://github.com/ChrisTitusTech/powershell-profile/raw/main/setup.ps1`
There are lot more profiles [here](https://github.com/topics/powershell-profile) if you want additional tests.

samruddhikhandale: Thanks, I created a fork at https://github.com/codspace/powershell-profile/tree/main

Can you write tests by using that? Would be great to validate these changes and have some tests added which massages this new Feature option.

samruddhikhandale (changes requested):
Looking for some test coverage, then we are all set.

src/powershell/install.sh
nebula-it: Hey @samruddhikhandale,
Apologies for delay in getting back to you, I was bit busy.
Picking this up again, I see three files [here](https://github.com/devcontainers/features/tree/main/test/powershell), but not really clear on how those are being used. Can you please share any docs on how to write test for this or you can guide me through this?

nebula-it: NM, I found the test docs. Will get the test implemented in next day or two.

nebula-it: Alright, I've added test now. I create a simple profile [here](https://github.com/nebula-it/features/blob/patch-2/test/powershell/testProfile.ps1) for testing, as the one I linked is too extensive and not really a good option for testing.

With the current test, if the variable `$env:ProfileLoaded` exists and returns value `True`, that means our profile has been successfully loaded.

src/powershell/README.md
```diff
@@ -17,6 +17,7 @@ Installs PowerShell along with needed dependencies. Useful for base Dockerfiles
 |-----|-----|-----|-----|
 | version | Select or enter a version of PowerShell. | string | latest |
 | modules | Optional comma separated list of PowerShell modules to install. | string | - |
+| powershellProfileURL | Optional URL to download Powershell Profile. | string | - |
```
samruddhikhandale, line 20: Once this PR is merged, https://github.com/devcontainers/action will run automatically ; which opens a PR that updates the README file based on `devcontainer-feature.json` and `NOTES.md`.

Hence, can we revert the changes to this file to avoid merge conflicts? Thanks!

test/powershell/scenarios.json
```diff
@@ -3,7 +3,8 @@
     "image": "mcr.microsoft.com/devcontainers/base:jammy",
     "features": {
       "powershell": {
-        "modules": "az.resources, az.storage"
+        "modules": "az.resources, az.storage",
+        "powershellProfileURL": "https://raw.githubusercontent.com/nebula-it/features/patch-2/test/powershell/testProfile.ps1"
```
samruddhikhandale, line 7: Would you be able to support and maintain it for longer term? I am worried that the test might break in future if that's not the case.

Can we use https://github.com/codspace/powershell-profile/blob/main/Test-Profile.ps1 instead? Updated the contents to match your test file

test/powershell/testProfile.ps1
```diff
@@ -0,0 +1 @@
+$env:ProfileLoaded = $true
```
samruddhikhandale, line 1: Why do we need to add this locally?

samruddhikhandale (changes requested):

test/powershell/scenarios.json
nebula-it: Yeap I can keep it, but using an external repo does not seem like a good solution to me.
So my thought process was that once this was merged, we can point to https://github.com/devcontainers/features/blob/main/test/powershell/testProfile.ps1 so its in same repo

nebula-it: Otherwise I can point it to your repo above, let me know which makes more sense to you

nebula-it: Oh I see that, above repo you posted is not a personal one. I have updated code to use that.

samruddhikhandale (approved):
Looks great ⚡ 

Let's wait for test runs to turn 🟢 before we merge it!

Issue #810, opened 2024-01-26
common-utils fails to build when base image is almalinux-[89]-minimal or rocky-[89]minimal


jdputschadi:
While submitting a PR  adding support for the RHEL OS family, we get a failing test: https://github.com/devcontainers/features/actions/runs/7671052872/job/20908644762?pr=782#step:4:7694:
```
[dev_containers_target_stage 5/5] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=common-utils_0,target=/tmp/build-features-src/common-utils_0     cp -ar /tmp/build-features-src/common-utils_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/common-utils_0  && cd /tmp/dev-container-features/common-utils_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/common-utils_0:
0.212     INSTALLOHMYZSH="true"
0.212     INSTALLOHMYZSHCONFIG="true"
0.212     UPGRADEPACKAGES="true"
0.212     USERNAME="automatic"
0.212     USERUID="automatic"
0.212     USERGID="automatic"
0.212     NONFREEPACKAGES="false"
0.212 ===========================================================================
0.241 ./main.sh: line 217: yum: command not found
0.241 ERROR: Feature "Common Utilities" (Unknown) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/common-utils for help troubleshooting this error.
------
```

This is been tracked down to the use of `microdnf` instead of `yum` or `dnf` in minimal images.

I will be submitting a PR to fix this problem.

samruddhikhandale:
Thank you! 🙌

PR #811, opened 2024-01-26
allow common-utils to be installed on "minimal" RHEL-based distributions

jdputschadi:
Fixes devcontainers/features#810

Allows `microdnf` to be used in addition to `dnf` or `yum` in the RedHat world. 

Tests for minimal base os images that are affected by this fix (almalinux-8-minimal, almalinux-9-minimal, rocky-8-minimal, and rocky-9-minimal).

### DIFF
--- a/test/common-utils/scenarios.json
+++ b/test/common-utils/scenarios.json
@@ -41,6 +41,34 @@
             "common-utils": {}
         }
     },
+    "alma-8": {
+        "image": "almalinux:8",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
+        }
+    },
+    "alma-9": {
+        "image": "almalinux:9",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
+        }
+    },
+    "alma-8-minimal": {
+        "image": "almalinux:8-minimal",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
+        }
+    },
+    "alma-9-minimal": {
+        "image": "almalinux:9-minimal",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
+        }
+    },
     "rocky-8": {
         "image": "rockylinux:8",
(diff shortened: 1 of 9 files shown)


test/common-utils/scenarios.json
```diff
@@ -41,6 +41,34 @@
             "common-utils": {}
         }
     },
+    "alma-8": {
```
samruddhikhandale, line 44: The tests are complaining about missing `alma-8.sh` file, which is added by https://github.com/devcontainers/features/pull/782

https://github.com/devcontainers/features/actions/runs/7674069177/job/20917976765?pr=811#step:4:7608

Can we either add the .sh file or remove the test scenario from this PR or to fix the test runs?

test/common-utils/scenarios.json
```diff
@@ -41,6 +41,34 @@
...
+        "image": "almalinux:8",
+        "remoteUser": "devcontainer",
+        "features": {
+            "common-utils": {}
+        }
+    },
+    "alma-9": {
```
samruddhikhandale, line 51: Same comment here!

samruddhikhandale (changes requested):
Thanks for the quick fix, left a minor comment!

test/common-utils/scenarios.json
jdputschadi: Yeah I noticed. I just pushed the missing .sh files.

test/common-utils/scenarios.json
jdputschadi: Yeah I noticed. I just pushed the missing .sh files.

test/common-utils/scenarios.json
jdputschadi: Commit 1a425bd fixes the missing .sh test scripts.

test/common-utils/scenarios.json
jdputschadi: Commit 1a425bd fixes the missing .sh test scripts.

samruddhikhandale (approved):
⚡ ⚡

Issue #817, opened 2024-01-29
[docker-in-docker] Docker buildx not installed in Codespaces


sinedied:
In my devcontainer, I'm enabling docker-in-docker using this config:
```
    "ghcr.io/devcontainers/features/docker-in-docker:2": {
      "moby": "false",
    },
```

However, when running in GitHub Codespaces, Docker Buildx doesn't seem to be installed as I get this error when trying to run `azd deploy` on an Azure sample repo using containers:
```
azd deploy

Deploying services (azd deploy)

  (x) Failed: Deploying service backend

ERROR: failed building service 'backend': failing invoking action 'build', building container: backend at ../..: building image: exit code: 1, stdout: , stderr: ERROR: BuildKit is enabled but the buildx component is missing or broken.
       Install the buildx component to build images with BuildKit:
       https://docs.docker.com/go/buildx/
```

Here's the installed docker version:
```
docker --version
Docker version 25.0.1, build 29cf629
```

samruddhikhandale:
Hi 👋 

Thanks for reporting this issue, I am able to reproduce this as well.

From the logs, and after adding some debugging statements, I can see that the `docker-in-docker` Feature is installing and saving `buildx` to https://github.com/devcontainers/features/blob/main/src/docker-in-docker/install.sh#L322

However, within the devcontainer/codespace, I don't see the required path available (ie. `/home/vscode/.docker/cli-plugins/docker-buildx`). It appears like the `~/.docker` folder is getting replaced ; which eventually makes the `buildx` unavailable.

![Image](https://github.com/devcontainers/features/assets/24955913/2317516c-94e9-4ee1-9393-d61fc4a96a53)

I'd need to investigate the purpose of sym-linking `/workspaces/.codespaces/.persistedshare/.docker` to `~/.docker` and either update Codespaces or update the `docker-in-docker` Feature.

PR #827, opened 2024-01-31
[docker-in-docker] : Bug fix - Buildx installation when moby:false

samruddhikhandale:
Closes https://github.com/devcontainers/features/issues/817

Syncs buildx installation directory when `moby:true` and `moby:false` which also avoids the problem mentioned in https://github.com/devcontainers/features/issues/817#issuecomment-1915916866


joshspicer (approved):
🔥

PR #833, opened 2024-02-06
Docker-in-docker - [Bug fix] : Fix externally-managed-environment python error for "bookworm"

samruddhikhandale:
Closes: https://github.com/devcontainers/features/issues/832

Changes needed to the "docker-in-docker" Feature post Debian bookworm / [PEP 668](https://peps.python.org/pep-0668/#abstract)

If OS is bookworm, install from git releases instead.

### DIFF
--- a/src/docker-in-docker/devcontainer-feature.json
+++ b/src/docker-in-docker/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "docker-in-docker",
-    "version": "2.8.0",
+    "version": "2.8.1",
     "name": "Docker (Docker-in-Docker)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/docker-in-docker",
     "description": "Create child containers *inside* a container, independent from the host's docker instance. Installs Docker extension in the container along with needed CLIs.",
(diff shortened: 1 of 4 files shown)


chrmarti (approved):

rubensa:
@samruddhikhandale I think this should also be ported to [docker-outside-of-docker](https://github.com/devcontainers/features/tree/main/src/docker-outside-of-docker), isn't it?

_NOTE_:  I'm currently getting following error when installing **docker-outside-of-docker** on a Docker container based on **mcr.microsoft.com/devcontainers/base:bookworm** for **arm64** architecture.

```bash
#12 90.97 error: externally-managed-environment
#12 90.97 
#12 90.97 × This environment is externally managed
#12 90.97 ╰─> To install Python packages system-wide, try apt install
#12 90.97     python3-xyz, where xyz is the package you are trying to
#12 90.97     install.
#12 90.97     
#12 90.97     If you wish to install a non-Debian-packaged Python package,
#12 90.97     create a virtual environment using python3 -m venv path/to/venv.
#12 90.97     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
#12 90.97     sure you have python3-full installed.
#12 90.97     
#12 90.97     If you wish to install a non-Debian packaged Python application,
#12 90.97     it may be easiest to use pipx install xyz, which will manage a
#12 90.97     virtual environment for you. Make sure you have pipx installed.
#12 90.97     
#12 90.97     See /usr/share/doc/python3.11/README.venv for more information.
#12 90.97 
#12 90.97 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
#12 90.97 hint: See PEP 668 for the detailed specification.
```

PR #867, opened 2024-02-16
sync `install.sh` to `devcontainer-feature.json` default values

### LINKED ISSUE #865: DooD line 187: err: command not found
Suddenly, running the `docker-outside-of-docker/install.sh` script gives me the following:

```
docker-outside-of-docker/install.sh: line 187: err: command not found
```

*I'm using the latest file (as of now) https://github.com/devcontainers/features/blob/eea8ec3a3702057ce3c0c3e54bbb205ed580867b/src/docker-outside-of-docker/install.sh

### Info:

- image: mcr.microsoft.com/devcontainers/base:ubuntu-22.04
- `docker info`:
```
Client:
 Version:    24.0.9-1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.12.1-1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.24.5-1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
```

### May be related with:

- https://github.com/devcontainers/features/pull/843
- https://github.com/devcontainers/features/issues/842

lucasfcnunes:
# Description

- Syncs SOME `install.sh` env vars to their respective `devcontainer-feature.json` default values 
- #865 is caused by a missing a version on the `MOBYBUILDXVERSION` env var.

# Execution

- Regex find `^[A-Z0-9_]+="\$\{[^:]+?\}` on `install.sh` files
- Bump patch versions on respective `devcontainer-feature.json` files

# Related

- Fixes #865


lucasfcnunes:

src/docker-in-docker/install.sh
```diff
@@ -10,7 +10,7 @@
 
 DOCKER_VERSION="${VERSION:-"latest"}" # The Docker/Moby Engine + CLI should match in version
 USE_MOBY="${MOBY:-"true"}"
-MOBY_BUILDX_VERSION="${MOBYBUILDXVERSION}"
+MOBY_BUILDX_VERSION="${MOBYBUILDXVERSION:-"latest"}"
```
samruddhikhandale, line 13: > *I'm using the latest file (as of now) https://github.com/devcontainers/features/blob/eea8ec3a3702057ce3c0c3e54bbb205ed580867b/src/docker-outside-of-docker/install.sh

From https://github.com/devcontainers/features/issues/865

Curious, @lucasfcnunes why are you using the script directly instead of using Dev Container Features? We don't provide guarantee or support that using scripts would always work.

Can you try updating your dev container to add Feature? See https://github.com/devcontainers/features/blob/main/src/docker-outside-of-docker/README.md#example-usage

samruddhikhandale, line 13: We intentionally avoided to set the default value for `MOBY_BUILDX_VERSION`, the @devcontainer/cli computes it from the [devcontainer-feature.json](https://github.com/devcontainers/features/blob/main/src/docker-outside-of-docker/devcontainer-feature.json#L25)

This adds a redundancy and whenever we want to update `default`, this leads to updating in both the placed. Failing to update in one place leads to confusions (we were bit by this few times, learned from experience)

lucasfcnunes: It's faster for the people using the devcontainer. We pre-build the image before distributing it.

lucasfcnunes: All other env vars already have a default value and haven't been changed for a long time. Using `latest` (like in `DOCKER_VERSION`) there wouldn't be any issue as the `.json` config overrides the default value: no need to change in the `.sh`.

samruddhikhandale: You can still prebuild the image with a `devcontainer.json` (topped with Features). Have you tried that? 

All the images hosted from https://github.com/devcontainers/images uses the same pattern, see https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/devcontainer.json

lucasfcnunes: > You can still prebuild the image with a devcontainer.json (topped with Features). Have you tried that?

I have. In the past.

Summarizing I do 4+ things that don't go so well with traditional devcontainer builds:

```Dockerfile
# 1. bleeding edge custom syntax (COPY --link, RUN --mount, ...)
# syntax=docker/dockerfile:1

# 2. heavy caching
RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache

RUN \
  --mount=type=cache,target=/var/cache/apt/,sharing=locked \
  --mount=type=cache,target=/var/lib/apt/,sharing=locked \
  {DooD, kubectl-helm-minikube}.sh # after `perl -p -i -e` out cache cleanse ('s|rm -rf /var/lib/apt/lists/\*|# <remove apt-cache>|g')

# 3. lots of `COPY --link` in multistage build (~9 FROM ... AS ...).
COPY --link ... ...

```

```
# 4. build all of it with just 01 `docker buildx bake` run (no `docker build` nor `docker compose build`)
```

---

I'm already setting the `MOBYBUILDXVERSION` env var in my build args to `latest` - so all good by my side.
This commit would close my initial issue and follow the tradition of all vars having default values.

samruddhikhandale: > follow the tradition of all vars having default values.

Sooner or later we will be cleaning up this stale tradition as it's not required for the dev container Features to work. Also, we don't recommend or support the use of script files.

A heads up, we can allow this PR contribution for now, but this could break at any point of time.

src/docker-in-docker/install.sh
```diff
@@ -10,7 +10,7 @@
 
 DOCKER_VERSION="${VERSION:-"latest"}" # The Docker/Moby Engine + CLI should match in version
 USE_MOBY="${MOBY:-"true"}"
```
samruddhikhandale, line 12: Can you bump patch versions for both of the updated Features?

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
lucasfcnunes: Done! https://github.com/devcontainers/features/pull/867/commits/55f07186eb973aaafdd13d13dbb59fb4ba990dd5

src/docker-in-docker/install.sh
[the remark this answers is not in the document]
lucasfcnunes: I am always inclined to `latest` everything but TO BE SAFE, I'll set to the working `0.12.0`.
I'll also point out that `DOCKERDASHCOMPOSEVERSION` is default in `v1` and I think I should sync it too.

I'll make it so this PR makes sense.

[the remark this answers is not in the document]
samruddhikhandale: Taking care of `DOCKERDASHCOMPOSEVERSION` in https://github.com/devcontainers/features/pull/873 for docker-in-docker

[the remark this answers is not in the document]
lucasfcnunes: Syncing done.

samruddhikhandale (approved):

PR #869, opened 2024-02-18
[ docker-in-docker ] - buildx can fallback to previous version if latest artifact not found

gauravsaini04:
**Feature name**:
 
 * Docker-in-Docker
 
 **Description**:
 
 This PR introduces the following feature:
 
 * `docker/buildx` plugin can fallback to previous version in case of latest tag artifact release not found;
  
 _Changelog_:
 
 * Updated install.sh file
   * Introduced a new flag `fallback` for testing this new feature
   * Code written for fallback to prev. version in case of latest artifact release not found
 * Added a scenario in `scenarios.json` and a corresponding test scenario `docker_build_fallback_buildx`
 
 _Command for testing_:
 
 `devcontainer features test . -f docker-in-docker --skip-autogenerated --filter "docker_build_fallback_buildx" --preserve-test-containers`

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -343,7 +343,8 @@ get_latest_version() {
 get_previous_version() {
     latest_version="$1"
     repo_url=$2
-    curl -s "$repo_url" | jq -r --arg latest "$latest_version" '.[].tag_name | select(. != $latest)' | sort -rV | head -n 1
+    # curl -s "$repo_url" | jq -r --arg latest "$latest_version" '.[].tag_name | select(. != $latest)' | sort -rV | head -n 1   # this helps filter tag_names based on semver specifications, this one was incorrectly installing 0.12.0-rc2 which is the third in the list i.e seeing from top to bottom
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name' # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
 }
 
 # Function to change the patch number in a semver version
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -343,10 +343,10 @@ fetch_previous_version() {
     if [ $wget_exit_code -ne 0 ]; then # means wget command to fetch latest version failed
         if [ $wget_exit_code -eq 8 ]; then  # failure due to 404: Not Found.
             echo -e "\n(!) Failed to fetch the latest artifacts for docker buildx v${buildx_version}..."
-            echo -e "\nAttempting to install v${previous_version}"
             repo_url="https://api.github.com/repos/docker/buildx/releases" # GitHub repository URL
             previous_version=$(get_previous_version "${repo_url}")
             buildx_file_name="buildx-${previous_version}.linux-${architecture}"
+            echo -e "\nAttempting to install ${previous_version}"
             wget https://github.com/docker/buildx/releases/download/${previous_version}/${buildx_file_name}
         else
             echo "(!) Failed to download docker buildx with exit code: $wget_exit_code"
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -19,7 +19,6 @@ INSTALL_DOCKER_BUILDX="${INSTALLDOCKERBUILDX:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy"
-FALLBACK="${FALLBACK:-"false"}" # fallback to previous version, for testing purpose only
 
 # Default: Exit on any failure.
 set -e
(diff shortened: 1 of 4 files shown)


src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+    wget https://github.com/docker/buildx/releases/download/v${buildx_version}/${buildx_file_name} || {
+        wget_exit_code=$?
+        if [ $wget_exit_code -ne 0 ]; then # means wget command to fetch latest version failed
+            if [ $wget_exit_code -eq 8 ]; then  # failure due to 404: Not Found.
+                echo "wget command failed with HTTP error 404 Not Found"
+                repo_url="https://api.github.com/repos/docker/buildx/releases" # GitHub repository URL
+                latest_version=$(get_latest_version "${repo_url}")  # can take latest_version from fn get_latest_version
+                # latest_version="v${buildx_version}" # can also take latest_version from fn find_version_from_git_tags
+                previous_version=$(get_previous_version "${latest_version}" "${repo_url}")
+                buildx_file_name="buildx-${previous_version}.linux-${architecture}"
```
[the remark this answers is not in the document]
sliekens: Honestly I would delete all the existing `find_version_from_git_tags` stuff, because it's unreliable for finding releases, and only keep the `get_latest_version()` which you introduced.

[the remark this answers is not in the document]
eitsupi: > Honestly I would delete all the existing `find_version_from_git_tags` stuff, because it's unreliable for finding releases, and only keep the `get_latest_version()` which you introduced.

The GitHub API has strict number of attempts limits and may fail.
I think that's why we are using Git tags here.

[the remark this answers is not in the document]
sliekens: Okay, I see, what if instead, first try to use the releases API, and if a rate limit is reached, fall back to git tags?

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
 
 usermod -aG docker ${USERNAME}
 
+# Function to fetch the latest version of the plugin
+get_latest_version() {
+    repo_url=$1
+    curl -s "$repo_url/latest" | jq -r '.tag_name'
```
samruddhikhandale, line 339: image

Doesn't work for me!

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
     echo "(*) Installing buildx ${buildx_version}..."
     buildx_file_name="buildx-v${buildx_version}.linux-${architecture}"
-    cd /tmp && wget "https://github.com/docker/buildx/releases/download/v${buildx_version}/${buildx_file_name}"
+    if [ $FALLBACK = "true" ]; then # flag to initiate testing of fallback feature
+        set -x # for getting detailed logs in case of a failure - flag to mark start of code
+        new_patch_number="42" # for testing a tag not found scenario for docker/buildx plugin
+        buildx_version_fallback_test=$(change_patch_number "$buildx_version" "$new_patch_number") # for testing a tag not found scenario for docker/buildx plugin
+        buildx_file_name="buildx-v${buildx_version_fallback_test}.linux-${architecture}"
+        buildx_version=$buildx_version_fallback_test
+    fi
```
samruddhikhandale, line 374: Same comment here. Looks like this is only meant for testing the new logic? Can we remove this and updates the tests instead?

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+    if [ $FALLBACK = "true" ]; then # flag to initiate testing of fallback feature
+        set -x # for getting detailed logs in case of a failure - flag to mark start of code
+        new_patch_number="42" # for testing a tag not found scenario for docker/buildx plugin
+        buildx_version_fallback_test=$(change_patch_number "$buildx_version" "$new_patch_number") # for testing a tag not found scenario for docker/buildx plugin
+        buildx_file_name="buildx-v${buildx_version_fallback_test}.linux-${architecture}"
+        buildx_version=$buildx_version_fallback_test
+    fi 
+    cd /tmp
+    wget https://github.com/docker/buildx/releases/download/v${buildx_version}/${buildx_file_name} || {
+        wget_exit_code=$?
```
samruddhikhandale, line 377: Can we add this alternate logic to a new function? That way the code will look cleaner

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+        if [ $wget_exit_code -ne 0 ]; then # means wget command to fetch latest version failed
```
samruddhikhandale, line 378: Feel free to remove this condition, it's all okay to not echo "Latest Version Fetch - Succeeded.\n"

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+            if [ $wget_exit_code -eq 8 ]; then  # failure due to 404: Not Found.
+                echo "wget command failed with HTTP error 404 Not Found"
```
samruddhikhandale, line 380:
```suggestion
                echo "(!) Failed to fetch the latest artifacts for docker buildx v${buildx_version}..."
```

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
 
+# Function to fetch the latest version of the plugin
+get_latest_version() {
+    repo_url=$1
+    curl -s "$repo_url/latest" | jq -r '.tag_name'
+}
+
+# Function to fetch the version released prior to the latest version
+get_previous_version() {
+    latest_version="$1"
```
samruddhikhandale, line 344: Doesn't seem like this is being used, remove?

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+    fi 
+    cd /tmp
+    wget https://github.com/docker/buildx/releases/download/v${buildx_version}/${buildx_file_name} || {
+        wget_exit_code=$?
+        if [ $wget_exit_code -ne 0 ]; then # means wget command to fetch latest version failed
+            if [ $wget_exit_code -eq 8 ]; then  # failure due to 404: Not Found.
+                echo "wget command failed with HTTP error 404 Not Found"
+                repo_url="https://api.github.com/repos/docker/buildx/releases" # GitHub repository URL
+                latest_version=$(get_latest_version "${repo_url}")  # can take latest_version from fn get_latest_version
+                # latest_version="v${buildx_version}" # can also take latest_version from fn find_version_from_git_tags
```
samruddhikhandale, line 383: Same here, remove?

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+                previous_version=$(get_previous_version "${latest_version}" "${repo_url}")
```
samruddhikhandale, line 384:
```suggestion
                previous_version=$(get_previous_version "${repo_url}")
```

src/docker-in-docker/install.sh
samruddhikhandale, line 385: Can we add an echo with message "Attempting to install v${previous_version}" ?

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+                buildx_file_name="buildx-${previous_version}.linux-${architecture}"
+                wget https://github.com/docker/buildx/releases/download/${previous_version}/${buildx_file_name}
+            else
+                echo "Wget failed with exit code: $wget_exit_code"
```
samruddhikhandale, line 388:
```suggestion
                echo "(!) Failed to download docker buildx with exit code: $wget_exit_code"
                exit 1
```

src/docker-in-docker/install.sh
```diff
@@ -350,6 +402,7 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
     chown -R "${USERNAME}:docker" "${docker_home}"
     chmod -R g+r+w "${docker_home}"
     find "${docker_home}" -type d -print0 | xargs -n 1 -0 chmod g+s
+    set +x # for getting detailed logs in case of a failure - flag to mark end of code
```
samruddhikhandale, line 405: Remove? (as we are removing other block of code)

src/docker-in-docker/install.sh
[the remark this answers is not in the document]
samruddhikhandale: > The GitHub API has strict number of attempts limits and may fail.
> I think that's why we are using Git tags here.

Yes, we should only attempt to use GitHub APIs if it is necessary This way we will avoid hitting the limits for users. Folks might require it for other purposes, so we shouldn't consume them as we have another alternative.

Hence, the approach in this PR makes sense to me. Find latest version with `find_version_from_git_tags`, if that version is currently in the release process (so it would fail to download artifacts), then attempt to use the API.

samruddhikhandale (changes requested):
Overall the PR makes sense, thanks for looking at it. Left few comments!

src/docker-in-docker/install.sh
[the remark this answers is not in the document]
gauravsaini04: thanks 👍 have corrected.

src/docker-in-docker/install.sh
```diff
@@ -19,6 +19,7 @@ INSTALL_DOCKER_BUILDX="${INSTALLDOCKERBUILDX:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy"
+FALLBACK="${FALLBACK:-"false"}" # fallback to previous version, for testing purpose only
```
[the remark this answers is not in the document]
gauravsaini04: have removed the new variable from install.sh, rather moved the testing code to the test case file

src/docker-in-docker/install.sh
```diff
@@ -19,6 +19,7 @@ INSTALL_DOCKER_BUILDX="${INSTALLDOCKERBUILDX:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy"
```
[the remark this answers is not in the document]
gauravsaini04: done 👍

src/docker-in-docker/install.sh
```diff
@@ -332,13 +333,64 @@ fi
...
+get_latest_version() {
+    repo_url=$1
+    curl -s "$repo_url/latest" | jq -r '.tag_name'
+}
+
+# Function to fetch the version released prior to the latest version
+get_previous_version() {
+    latest_version="$1"
+    repo_url=$2
+    # curl -s "$repo_url" | jq -r --arg latest "$latest_version" '.[].tag_name | select(. != $latest)' | sort -rV | head -n 1   # this helps filter tag_names based on semver specifications, this one was incorrectly installing 0.12.0-rc2 which is the third in the list i.e seeing from top to bottom
```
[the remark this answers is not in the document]
gauravsaini04: removed 👍

src/docker-in-docker/install.sh
gauravsaini04: have removed this code thanks 👍

src/docker-in-docker/install.sh
gauravsaini04: have added to a fn 👍

src/docker-in-docker/install.sh
gauravsaini04: corrected the echo string

src/docker-in-docker/install.sh
gauravsaini04: removed 👍

src/docker-in-docker/install.sh
gauravsaini04: removed 👍

src/docker-in-docker/install.sh
gauravsaini04: removed

src/docker-in-docker/install.sh
gauravsaini04: added this echo

src/docker-in-docker/install.sh
gauravsaini04: added this msg 👍

src/docker-in-docker/install.sh
gauravsaini04: removed 👍

src/docker-in-docker/install.sh
```diff
@@ -333,65 +332,38 @@ fi
...
-    # Extract major, minor, and current patch numbers
-    local major=$(echo "$version" | cut -d. -f1)
-    local minor=$(echo "$version" | cut -d. -f2)
-    local current_patch=$(echo "$version" | cut -d. -f3)
-    # Construct the new version with the updated patch number
-    local new_version="$major.$minor.$new_patch"
-    echo "$new_version"
+fetch_previous_version() {
+    wget_exit_code=$?
+    if [ $wget_exit_code -ne 0 ]; then # means wget command to fetch latest version failed
```
samruddhikhandale, line 343: Looks like we can remove this condition?

src/docker-in-docker/install.sh
```diff
@@ -333,65 +332,38 @@ fi
...
+        if [ $wget_exit_code -eq 8 ]; then  # failure due to 404: Not Found.
+            echo -e "\n(!) Failed to fetch the latest artifacts for docker buildx v${buildx_version}..."
```
samruddhikhandale, line 345: Question: Will `buildx_version` be available to this function or it needs to be passed in?

src/docker-in-docker/install.sh
```diff
@@ -333,65 +332,38 @@ fi
...
-    local version="$1"  # Input version
-    local new_patch="$2"  # New patch number
-    # Extract major, minor, and current patch numbers
-    local major=$(echo "$version" | cut -d. -f1)
-    local minor=$(echo "$version" | cut -d. -f2)
-    local current_patch=$(echo "$version" | cut -d. -f3)
-    # Construct the new version with the updated patch number
-    local new_version="$major.$minor.$new_patch"
-    echo "$new_version"
+fetch_previous_version() {
```
samruddhikhandale, line 341:
```suggestion
install_previous_version_artifacts() {
```

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: yes, 👍thanks, have removed !

src/docker-in-docker/install.sh
gauravsaini04: changed the name of fn 👍

samruddhikhandale (approved):
Looks great ⚡

Issue #745, opened 2023-11-07
docker-in-docker: Switch the default to v2


samruddhikhandale:
The `docker-in-docker` Feature uses `v1` as default value. As docker v1 is deprecated, we should set the default to `v2`.

TODOs - 

- Update `dockerDashComposeVersion`  default to `latest`  (ie `v2`)
- Deprecate support for option `v1`  for `dockerDashComposeVersion`
- Keep the installation code for docker compose switch
- Bump the major version as we are deprecating support for `v1` (Double check)

We had a community contribution to update the default, see https://github.com/devcontainers/features/pull/621 which was closed by the author. @jacobwoffenden Let us know if you are still interested in reopening the PR, if not can use it as reference (unless there are any concerns). Thank you!

PR #873, opened 2024-02-22
Docker-in-docker: Updates default value for "dockerDashComposeVersion" to "latest

samruddhikhandale:
Closes: https://github.com/devcontainers/features/issues/745

Docker compose v1 was deprecated last year, hence, we shouldn't set default to `v1`. This PR updates the default to `latest`, however, adds back compat (via docker compose switch) for folks depending on `v1`. As Docker still allows installing `v1`, we allow it too. However, we have added a warning to let users know.

This also adds `installDockerComposeSwitch:true` as a new Feature option which allows user to avoid installing the switch.


src/docker-in-docker/install.sh
```diff
@@ -245,75 +246,77 @@ fi
...
-        update-alternatives --install /usr/local/bin/docker-compose docker-compose "${target_v1_compose_path}" 1
-    fi
-    if [ "${DOCKER_DASH_COMPOSE_VERSION}" = "v1" ]; then
-        update-alternatives --set docker-compose "${target_v1_compose_path}"
+        mv "${current_compose_path}" "${target_compose_path}"
+        update-alternatives --install ${docker_compose_path} docker-compose /usr/local/bin/compose-switch 99
+        update-alternatives --install ${docker_compose_path} docker-compose "${target_compose_path}" 1
     else
-        update-alternatives --set docker-compose /usr/local/bin/compose-switch
+        err "Skipping installation of compose-switch as docker compose is unavailable..."
```
chrmarti, line 319: `docker-compose` (v1) is unavailable, but if `docker compose` (v2) is there, wouldn't it make sense to still install compose-switch to get a `docker-compose` executable that converts the command line args and forwards to `docker compose` for compatibility?

chrmarti (changes requested):
Looks good, left a question about compatibility.

chrmarti (approved):
LGTM!

Issue #334, opened 2022-12-14
kubectl-helm-minicube: Adds retry logic to find git tags 


xec-abailey:
Hello!

Have been using the kubectl devcontainer feature for about a year (or close to it) with no issues to speak of until this week. Here is the error showing up in the build log:

```
 #19 7.806 HELM_GPG_KEYS_URI=https://raw.githubusercontent.com/helm/helm/main/KEYS
2022-12-14 16:56:03.273Z: #19 7.809 GPG_KEY_SERVERS=keyserver hkp://keyserver.ubuntu.com:80
2022-12-14 16:56:03.273Z: #19 7.809 keyserver hkps://keys.openpgp.org
2022-12-14 16:56:03.274Z: #19 7.809 keyserver hkp://keyserver.pgp.com
2022-12-14 16:56:03.274Z: #19 8.012 Verification failed!
2022-12-14 16:56:03.274Z: #19 8.013 gpg: no valid OpenPGP data found.
2022-12-14 16:56:03.274Z: #19 8.013 gpg: the signature could not be verified.
```

And here is the devcontainer.json we are using: 

```json
{
	"name": "Machine Learning",
	"build": {
		"dockerfile": "../Dockerfile",
		"context": "..",
		"target": "development"
	},
	"hostRequirements": {
		"cpus": 4,
		"memory": "8gb",
		"storage": "32gb"
	},
	"runArgs": [
		"--env-file",
		".env"
	],
	"mounts": [
		"source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
		"source=projectname-bashhistory,target=/commandhistory,type=volume"
	],
	"initializeCommand": "docker login -u azureml-dev -p /******  ********",
	"postAttachCommand": "chmod +x scripts/build_and_push_dagster.sh",
	"extensions": [
		"ms-python.python",
		"ms-azuretools.vscode-docker",
		"ms-vscode.azure-account",
		"ms-kubernetes-tools.vscode-kubernetes-tools",
		"ms-kubernetes-tools.vscode-aks-tools",
		"eamodio.gitlens",
		"xyz.local-history",
		"ms-toolsai.jupyter",
		"gruntfuggly.todo-tree",
		"randomfractalsinc.vscode-data-preview",
		"wmaurer.change-case",
		"moshfeu.compare-folders",
		"github.copilot",
		"mutantdino.resourcemonitor",
		"tyriar.sort-lines",
		"redhat.vscode-yaml",
[...]

PR #875, opened 2024-02-23
[kubectl-helm-minikube]: Handle failing build with fallback to prev. version for helm

gauravsaini04:
Closes https://github.com/devcontainers/features/issues/334

 **Feature Name**:
 
 * kubectl-helm-minikube
 
 **Description**:
 
 * If package artifact (source binary) of latest version of released tag for `helm` is not able to download (i.e. it is not found) then download the previous version to the latest
 
 _Changelog_:
 
 * Updated install.sh file in `src/kubectl-helm-minikube/install.sh`
 * Added test file to test the feature that was added
 * Increased the patch version in `src/kubectl-helm-minikube/devcontainer-feature.json`
 
 **Checklist**:
 

src/kubectl-helm-minikube/install.sh
```diff
@@ -135,6 +141,16 @@ if [ ${KUBECTL_VERSION} != "none" ]; then
     fi
     curl -sSL -o /usr/local/bin/kubectl "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${architecture}/kubectl"
     chmod 0755 /usr/local/bin/kubectl
+    if [ -e "/usr/local/bin/kubectl" ]; then
+        if grep -q "The specified key does not exist." "/usr/local/bin/kubectl"; then
```
samruddhikhandale, line 145: Why do we need to do this? I'd expect `curl` to return a non-zero exit code which denotes it couldn't find artifacts?

samruddhikhandale (changes requested):

src/kubectl-helm-minikube/install.sh
gauravsaini04: it shows the same exit code (12) in both the cases i.e. with correct version with which it could find the artifact and without correct version with which it could not. Though the file /usr/local/bin/kubectl showed the change when I did not use the correct version to fetch artifacts:
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: kubernetes-release/release/v1.29.xyz/bin/linux/amd64/kubectl</Details></Error>

src/kubectl-helm-minikube/install.sh
```diff
@@ -135,6 +141,16 @@ if [ ${KUBECTL_VERSION} != "none" ]; then
     fi
     curl -sSL -o /usr/local/bin/kubectl "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${architecture}/kubectl"
     chmod 0755 /usr/local/bin/kubectl
+    if [ -e "/usr/local/bin/kubectl" ]; then
+        if grep -q "The specified key does not exist." "/usr/local/bin/kubectl"; then 
+            echo -e "\n(!) Failed to fetch the latest artifacts for docker buildx ${KUBECTL_VERSION}..."
```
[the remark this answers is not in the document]
gauravsaini04: done 👍

src/kubectl-helm-minikube/install.sh
```diff
@@ -122,6 +122,13 @@ case $architecture in
     *) echo "(!) Architecture $architecture unsupported"; exit 1 ;;
```
samruddhikhandale, line 122: Can you bump the patch version of the Feature? 

ℹ️ Once the PR is merged, the action automatically publishes new version of this Feature (**provided the version is updated). We follow semver rules for bumping versions.

src/kubectl-helm-minikube/install.sh
```diff
@@ -134,6 +141,15 @@ if [ ${KUBECTL_VERSION} != "none" ]; then
         KUBECTL_VERSION="v${KUBECTL_VERSION}"
     fi
     curl -sSL -o /usr/local/bin/kubectl "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${architecture}/kubectl"
+    if [ -e "/usr/local/bin/kubectl" ]; then
+        if grep -q "The specified key does not exist." "/usr/local/bin/kubectl"; then 
+            repo_url=https://api.github.com/repos/kubernetes/kubernetes/releases
+            requested_version=$(get_previous_version "${repo_url}")
+            echo -e "\nAttempting to install ${requested_version}"
+            curl -sSL -o /usr/local/bin/kubectl "https://dl.k8s.io/release/${requested_version}/bin/linux/${architecture}/kubectl"
```
samruddhikhandale, line 149: Looks like we are using `curl -L -s https://dl.k8s.io/release/stable.txt` to find the latest version (this will always work). Hence, we don't need to add retry logic for `kubectl`. Can you clean it up?

src/kubectl-helm-minikube/install.sh
```diff
@@ -167,6 +183,18 @@ if [ ${HELM_VERSION} != "none" ]; then
     tmp_helm_filename="/tmp/helm/${helm_filename}"
     curl -sSL "https://get.helm.sh/${helm_filename}" -o "${tmp_helm_filename}"
     curl -sSL "https://github.com/helm/helm/releases/download/${HELM_VERSION}/${helm_filename}.asc" -o "${tmp_helm_filename}.asc"
```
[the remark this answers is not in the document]
samruddhikhandale: Here we are trying to find if `helm` artifacts were installed based on `/tmp/helm/"${helm_filename}.asc`. This file is used for verifying integrity of the artifact. Hence, this approach doesn't look correct, can we determine if the artifacts were downloaded based on `curl -sSL "https://get.helm.sh/${helm_filename}" -o "${tmp_helm_filename}"` ?

src/kubectl-helm-minikube/install.sh
```diff
@@ -214,6 +242,16 @@ if [ "${MINIKUBE_VERSION}" != "none" ]; then
     fi
     # latest is also valid in the download URLs 
     curl -sSL -o /usr/local/bin/minikube "https://storage.googleapis.com/minikube/releases/${MINIKUBE_VERSION}/minikube-linux-${architecture}"    
+    if [ -e "/usr/local/bin/minikube" ]; then
+        if grep -q "The specified key does not exist." "/usr/local/bin/minikube"; then
```
samruddhikhandale, line 246: Same comment as before. Can we remove this? 

We add retry logic only if the `latest` version is determined by the git tags. It's not the case here, see https://github.com/devcontainers/features/blob/0f5716c8efdfb491b4747e20f1a1aa72df39ff32/src/kubectl-helm-minikube/install.sh#L236

src/kubectl-helm-minikube/install.sh
samruddhikhandale: No longer relevant

src/kubectl-helm-minikube/install.sh
```diff
@@ -167,6 +183,18 @@ if [ ${HELM_VERSION} != "none" ]; then
     tmp_helm_filename="/tmp/helm/${helm_filename}"
     curl -sSL "https://get.helm.sh/${helm_filename}" -o "${tmp_helm_filename}"
     curl -sSL "https://github.com/helm/helm/releases/download/${HELM_VERSION}/${helm_filename}.asc" -o "${tmp_helm_filename}.asc"
+    output=$(cat /tmp/helm/"${helm_filename}.asc")
+    if [[ "${output}" == *"Not Found"* ]]; then
+        echo -e "\n(!) Failed to fetch the latest artifacts for docker buildx ${HELM_VERSION}..."
+        repo_url=https://api.github.com/repos/helm/helm/releases
+        requested_version=$(get_previous_version "${repo_url}")
+        echo -e "\nAttempting to install ${requested_version}"
```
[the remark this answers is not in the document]
samruddhikhandale: No longer relevant

samruddhikhandale (changes requested):

src/kubectl-helm-minikube/install.sh
gauravsaini04: done

src/kubectl-helm-minikube/install.sh
gauravsaini04: removed !

src/kubectl-helm-minikube/install.sh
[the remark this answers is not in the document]
gauravsaini04: doing the similar only. Checking if file was not created i.e. it is throwing error `Not Found` when using a dummy version to check. So, if need any particular check, let me know.

src/kubectl-helm-minikube/install.sh
```diff
@@ -167,6 +174,17 @@ if [ ${HELM_VERSION} != "none" ]; then
...
     curl -sSL "https://github.com/helm/helm/releases/download/${HELM_VERSION}/${helm_filename}.asc" -o "${tmp_helm_filename}.asc"
+    if grep -q "BlobNotFound" "${tmp_helm_filename}"; then
+        echo -e "\n(!) Failed to fetch the latest artifacts for helm ${HELM_VERSION}..."
+        repo_url=https://api.github.com/repos/helm/helm/releases
+        requested_version=$(get_previous_version "${repo_url}")
+        echo -e "\nAttempting to install ${requested_version}"
+        helm_filename="helm-${requested_version}-linux-${architecture}.tar.gz"
+        tmp_helm_filename="/tmp/helm/${helm_filename}"
+        curl -sSL "https://get.helm.sh/${helm_filename}" -o "${tmp_helm_filename}"
+        curl -sSL "https://github.com/helm/helm/releases/download/${requested_version}/${helm_filename}.asc" -o "${tmp_helm_filename}.asc"
```
samruddhikhandale, line 185: This command looks repetitive, how about moving it out of the condition block? and syncing this line with :176 (`curl -sSL "https://github.com/helm/helm/releases/download/${HELM_VERSION}/${helm_filename}.asc" -o "${tmp_helm_filename}.asc"` 

Both commands are similar, we only need to update variables

samruddhikhandale, line 185: These lines are repetitive as well, how about moving this block of code to a function and reusing it in here as well as the normal flow?

samruddhikhandale (changes requested):

src/kubectl-helm-minikube/install.sh
gauravsaini04: have updated ! thanks 👍

gauravsaini04: have updated to a function

samruddhikhandale (approved):
Looks great, thanks!

PR #876, opened 2024-02-23
Fix profile not downloading

nebula-it:
Since the powershell folder is owned by root, the profile download fails without becoming sudo first.
```
ls -l /opt/microsoft/powershell/
total 0
drwxr-xr-x. 1 root root 22 Feb 23 08:37 7
```


nebula-it:
@samruddhikhandale Quick fix here

src/powershell/install.sh
```diff
@@ -166,7 +166,7 @@ fi
 # If URL for powershell profile is provided, download it to '/opt/microsoft/powershell/7/profile.ps1'
 if [ -n "$POWERSHELL_PROFILE_URL" ]; then
     echo "Downloading PowerShell Profile from: $POWERSHELL_PROFILE_URL"
-    curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"
+    sudo curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"
```
eitsupi, line 169: Generally, sudo does not inherit environment variables, so this command will fail behind a proxy.
Use su instead of sudo as you would elsewhere.

eitsupi (changes requested):

src/powershell/install.sh
samruddhikhandale: I agree ^. Also, along with using `su`, should we be considerate of the user as well? So the command will be 👇 

```
 su ${USERNAME} -c "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"
 ```

`USERNAME` can be computed by referring to any other Feature. See https://github.com/devcontainers/features/blob/f4f492ce2b9c11afc441870424aef560de4f092c/src/node/install.sh#L232-L247

Comment (nebula-it) [reply src/powershell/install.sh]: @eitsupi can we use `sudo -E` instead? That would preserve the env as well.

I tried `su -c 'curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "https://raw.githubusercontent.com/codspace/powershell-profile/main/Test-Profile.ps1"' root` but it asks for password.

@samruddhikhandale I tried that code in a test container and it returns empty value for `USERNAME`

Comment (eitsupi) [reply src/powershell/install.sh]: > can we use `sudo -E` instead?

If I remember correctly, it cannot be used.

Comment (nebula-it) [reply src/powershell/install.sh]: I tested and it works fine in container. If not, do you have a way around for `su` asking for password?

Comment (nebula-it) [reply src/powershell/install.sh]: @eitsupi please let me know your thoughts on how you want to proceed here

Comment (eitsupi) [reply src/powershell/install.sh]: @nebula-it I think you need to check other Features in this repository, set `$USERNAME` and run like `su ${USERNAME} -c ...`.
I am not the maintainer of this repository so I cannot approve this PR.

Comment (nebula-it) [reply src/powershell/install.sh]: @samruddhikhandale Do you mind adding and testing the code you want here if `sudo -E` is not desirable 🙏

Comment (samruddhikhandale) [reply src/powershell/install.sh]: > I think you need to check other Features in this repository, set $USERNAME and run like su ${USERNAME} -c ....

I have run the test workflow, however, those are my thoughts as well ^, see https://github.com/devcontainers/features/pull/876#discussion_r1501031595

Comment (samruddhikhandale) [reply src/powershell/install.sh]: What `devcontainer.json` or scenario did you use for testing? That could be the reason for it to be `empty`. Also, if it's empty then below code helps to compute the correct one.

https://github.com/devcontainers/features/blob/f4f492ce2b9c11afc441870424aef560de4f092c/src/node/install.sh#L232-L247

Comment (nebula-it) [reply src/powershell/install.sh]: Here are the contents of devcontainer.json
```
{
	"name": "Node.js & TypeScript",
	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-20-bullseye",
  "features": {
    "powershell": {
      "version": "latest",
      "powershellProfileURL": "https://raw.githubusercontent.com/codspace/powershell-profile/main/Test-Profile.ps1"
    }
  }
}
```
once the container is up, I saved this to a script and ran it, but output is empty
```
#!/bin/bash
# Determine the appropriate non-root user 
 if [ "${USERNAME}" = "auto" ] || [ "${USERNAME}" = "automatic" ]; then 
     USERNAME="" 
     POSSIBLE_USERS=("vscode" "node" "codespace" "$(awk -v val=1000 -F ":" '$3==val{print $1}' /etc/passwd)") 
     for CURRENT_USER in "${POSSIBLE_USERS[@]}"; do 
         if id -u ${CURRENT_USER} > /dev/null 2>&1; then 
             USERNAME=${CURRENT_USER} 
             break 
         fi 
     done 
     if [ "${USERNAME}" = "" ]; then 
         USERNAME=root 
     fi 
 elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then 
     USERNAME=root 
fi

echo $USERNAME
```
The output is empty.

Additionally, no matter what username I pass to `su`, it always prompts for password. Few examples I tried:
```
su -c 'curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "https://raw.githubusercontent.com/codspace/powershell-profile/main/Test-Profile.ps1"'
su node -c 'curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "https://raw.githubusercontent.com/codspace/powershell-profile/main/Test-Profile.ps1"'
su root -c 'curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "https://raw.githubusercontent.com/codspace/powershell-profile/main/Test-Profile.ps1"'
```
(the above devcontainer had `node` as user)

Comment (nebula-it) [reply src/powershell/install.sh]: @samruddhikhandale Can you please check latest update. I've used `sudo` to do `su` so it should not ask for password but still behave same as `su` would

Comment (samruddhikhandale) [reply src/powershell/install.sh]: > once the container is up, I saved this to a script and ran it, but output is empty

That's the issue, the compute-USER script should be run at the same time the Feature is running (ie during dev container setup)
Hence, we suggested to add the logic to the Feature script.

I tested with the local Feature, and the user is computed correctly

image

Comment (nebula-it) [reply src/powershell/install.sh]: Sure, I can do that then. However `su` will still ask for password, wont it?

Comment (samruddhikhandale) [reply src/powershell/install.sh]: It doesn't ask when you run this code within the Feature (I believe that's because the `install.sh` is run by root)

Comment (nebula-it) [reply src/powershell/install.sh]: Alrighty, sounds good. I've updated the code for that.

**Comment by nebula-it** [comment]
@samruddhikhandale tests are good, mind approving this please

File: src/powershell/install.sh [hunk 2]
```diff
@@ -163,10 +163,26 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
     done
 fi
```
samruddhikhandale, line 164: Can we bump patch version of this Feature?

src/powershell/install.sh
```diff
@@ -163,10 +163,26 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
...
 
+ # Determine the appropriate non-root user
```
samruddhikhandale, line 166: Can we add 👇 somewhere at the top?

```
USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
```

src/powershell/install.sh
```diff
@@ -163,10 +163,26 @@ if [ ${#POWERSHELL_MODULES[@]} -gt 0 ]; then
...
+         USERNAME=root 
+     fi 
+ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then 
+     USERNAME=root 
+ fi 
 # If URL for powershell profile is provided, download it to '/opt/microsoft/powershell/7/profile.ps1'
 if [ -n "$POWERSHELL_PROFILE_URL" ]; then
     echo "Downloading PowerShell Profile from: $POWERSHELL_PROFILE_URL"
-    curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"
+    su ${USERNAME} -c 'curl -sSL -o "/opt/microsoft/powershell/7/profile.ps1" "$POWERSHELL_PROFILE_URL"'
```
samruddhikhandale, line 185:
```suggestion
    su ${USERNAME} -c "curl -sSL -o '/opt/microsoft/powershell/7/profile.ps1' '$POWERSHELL_PROFILE_URL'"
```

:nit

samruddhikhandale (changes requested):

samruddhikhandale (approved):
Looks good, thank you! 
Let's merge in when the tests are 🟢

nebula-it:
Yayy, everything looks 🟢

PR #889, opened 2024-03-01
Fix PowerShell profile not loading

nebula-it:
Hello @samruddhikhandale, Its me again.

I tested the https://github.com/devcontainers/features/pull/876 and it still does not work. I found out how to test a feature locally and confirm what we are trying to achieve will not work with `su` as `su` will always ask for password when switching to `root` user. The code snippet we added from the node is to figure out non-root user and not to become root user. So, using `sudo` is our only option. I've added `-E` so that should take care of concerns about environment not passed to sudo.

Additionally, I've updated the code to get Profile path from PowerShell itself instead of hardcoding it.

Let me know if you have any questions

### DIFF
--- a/src/powershell/devcontainer-feature.json
+++ b/src/powershell/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "powershell",
-    "version": "1.3.4",
+    "version": "1.3.3",
     "name": "PowerShell",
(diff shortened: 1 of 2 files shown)


nebula-it:
@samruddhikhandale 👋

samruddhikhandale (approved):

PR #891, opened 2024-03-04
Add Docker on Docker shim to docker-outside-of-docker

felipecrs:
As an avid user of Docker on Docker (or as you call, Docker outside of Docker), I figured that not being able to handle volume mounts within the container is unnaceptable. That's why I built the [Docker on Docker shim](https://github.com/felipecrs/docker-on-docker-shim), which is a transparent shim that translates paths within the container to their correct counterparts on the host.

Apart from just devcontainer's needs, I built the tool for some business use cases as well. For example, to have it packed in a CI/CD pipeline to allow running workload within containers, or to have a tool that ships as a container but needs to spawn more containers within it, without the overhead of running a full Docker daemon and the `--privileged` flag.

I have been testing internally it for almost an year by now in production, and I believe it has matured enough to be shipped by this feature by default. It is also [very well tested](https://github.com/felipecrs/docker-on-docker-shim?tab=readme-ov-file#highlights).

When `useDockerOnDockerShimAsDocker` is enabled, calling `docker` from within the container will actually call the shim. Otherwise, the shim is available as the `dond` command.

It is a big improvement for the feature, in my opinion.

Please let me know your thoughts.

### TODOs

- [ ] Add tests for the feature

### DIFF
--- a/src/oryx/devcontainer-feature.json
+++ b/src/oryx/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.3.0",
+    "version": "1.3.1",
     "name": "Oryx",
     "description": "Installs the oryx CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/oryx",
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -86,7 +86,6 @@ install_dotnet_with_script()
 
     DOTNET_BINARY="dotnet"
     export PATH="${PATH}:/usr/share/dotnet"
-    DOTNET_BINARY_INSTALLATION="/usr/share/dotnet/sdk/${version}"
 }
 
 install_dotnet_using_apt() {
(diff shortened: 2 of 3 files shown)


samruddhikhandale:
Hi 👋 

We appreciate your enthusiasm and contribution for adding `docker-on-docker-shim` to the first party Feature. However, as https://github.com/felipecrs/docker-on-docker-shim is hosted on a personal repository, I'm afraid we won't be able to accept this contribution. We rely on first party/official releases of tools/artifacts which are trusted by the community and are happy to add it to their dev container without second thoughts.

@felipecrs Have you thought of publishing it as new Feature which you could share with the community?  The [devcontainers/feature-starter](https://github.com/devcontainers/feature-starter) repo has more guidance on self authoring them.

Once you've self authored and if you'd like to share it with the community, we'd recommend [adding it to the index](https://github.com/devcontainers/feature-starter#adding-features-to-the-index). It will make them visible in our [supporting tools](https://containers.dev/supporting) (like https://containers.dev/templates, Dev Containers Extension, GitHub Codespaces etc).

Feel free to ask any questions/queries along the way, feedbacks are welcome too!

Looping in @bamurtaugh (whose a PM on our team) to share her thoughts as well, thanks!

felipecrs:
Hi @samruddhikhandale, super understand. I will try creating a devcontainer feature for it too. Thanks for the advice.

Issue #883, opened 2024-02-26
Docker-in-docker: Add retry mechanism for fetching tags


samruddhikhandale:
Similar to https://github.com/devcontainers/features/pull/869, can we add fallback mechanism (ie when artifacts of `latest` version are unavailable, find the previous one) to "docker compose" and "docker switch" installations?

https://github.com/gauravsaini04/features/blob/main/src/docker-in-docker

PR #897, opened 2024-03-06
[docker] - Install previous version if current tags are missing artifacts

samruddhikhandale:
Fixes: https://github.com/devcontainers/features/issues/883

Currently, `docker-in-docker` Feature builds are failing as the latest tag available for docker-compose is [v2.24.7](https://github.com/docker/compose/releases/tag/v2.24.7), however, the corresponding releases are still not available. This PR, attempts to find a previous version in such scenarios


joshspicer (approved):
Looks good to me 👍

PR #901, opened 2024-03-07
[docker-outside-of-docker] compose-switch can fallback to previous version

gauravsaini04:
**Feature name**:
 
 * Docker-outside-of-Docker
 
 **Description**:
 
 * Fallback mechanism for docker-compose-switch
 
 _Changelog_:
 
 * Updated `install.sh` for `docker-outside-of-docker` feature.
 * Updated test file `docker_build_compose_fallback.sh` to verify that fallback mechanism works as expected through an added scenario in `scenarios.json`. 
 
 **Checklist**:
 
### DIFF
--- a/src/docker-outside-of-docker/install.sh
+++ b/src/docker-outside-of-docker/install.sh
@@ -99,6 +99,22 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
 }
 
+# Function to fetch the previous version of the plugin
+get_previous_version() {
+    repo_url=$1
+    # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name' 
+}
+
+
+install_compose_switch_fallback() {
+    echo -e "\n(!) Failed to fetch the latest artifacts for compose-switch v${compose_switch_version}..."
+    previous_version=$(get_previous_version "https://api.github.com/repos/docker/compose-switch/releases")
+    echo -e "\nAttempting to install ${previous_version}"
+    compose_switch_version=${previous_version#v}
+    curl -fsSL "https://github.com/docker/compose-switch/releases/download/v${compose_switch_version}/docker-compose-linux-${architecture}" -o /usr/local/bin/docker-compose
+}
+
 # Ensure apt is in non-interactive to avoid prompts
 export DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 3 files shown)


src/docker-outside-of-docker/install.sh
```diff
@@ -99,6 +99,22 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
 }
 
+# Function to fetch the previous version of the plugin
+get_previous_version() {
+    repo_url=$1
+    # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name'
```
samruddhikhandale, line 106:
```suggestion
    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name' 
```

Shouldn't this be updated to ^? 

image

^ However, the latest available release is v1.0.5, see https://github.com/docker/compose-switch/releases/tag/v1.0.5

samruddhikhandale: Can you update this in other Features as well where you previously made these changes?

src/docker-outside-of-docker/install.sh
```diff
@@ -99,6 +99,22 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
```
samruddhikhandale, line 99: Can we bump patch version of the Feature?

test/docker-outside-of-docker/docker_build_compose_fallback.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo -e "\n(!) Failed to fetch the latest artifacts for compose-switch ${test_compose_switch_version}..."
+    previous_version=$(get_previous_version)
+    echo -e "\nAttempting to install ${previous_version}"
+    compose_switch_version=${previous_version}
+    sudo curl -fsSL "https://github.com/docker/compose-switch/releases/download/${compose_switch_version}/docker-compose-linux-${architecture}" -o /usr/local/bin/docker-compose
+}
+
+install_compose-switch_as_docker-compose() {
+    echo "(*) Installing compose-switch as docker-compose..."
+    test_compose_switch_version=$(change_version_to_fail)
```
samruddhikhandale, line 79: Why do you need any computation here? You can give any version which won't exist eg. v1.5.x or anything which is not semver

samruddhikhandale (changes requested):

src/docker-outside-of-docker/install.sh
gauravsaini04: reverted to this logic

src/docker-outside-of-docker/install.sh
gauravsaini04: bumped the patch version

test/docker-outside-of-docker/docker_build_compose_fallback.sh
gauravsaini04: made changes

samruddhikhandale (approved):
Looks good, thank you! ⚡

PR #902, opened 2024-03-10
Reverting get_previous_version logic to get the first tag_name's value from json

gauravsaini04:
**Features names**:
 
 * Kubectl-helm-minikube
 
 **Description**:
 
 This PR corrects the existing code with the following functionality:
 
 * `kubectl-helm-minikube` feature was picking up the second to latest tag from json array of releases url, now have changed that to pick the first value for `tag_name` key in the same.
 
 _Changelog_:
 
 * Updated `install.sh` file
 * Updated tests
 
 **Checklist**:
 
### DIFF
--- a/src/kubectl-helm-minikube/install.sh
+++ b/src/kubectl-helm-minikube/install.sh
@@ -159,7 +159,7 @@ fi
 get_previous_version() {
     repo_url=$1
     # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
-    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name' 
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name' 
 }
 
 get_helm() {
--- a/src/kubectl-helm-minikube/install.sh
+++ b/src/kubectl-helm-minikube/install.sh
@@ -158,7 +158,7 @@ fi
 # Function to fetch the version released prior to the latest version
 get_previous_version() {
     repo_url=$1
-    # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
+    # this would del the assets key and then get the first encountered tag_name's value from the filtered array of objects
     curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name' 
 }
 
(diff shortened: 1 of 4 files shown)


src/kubectl-helm-minikube/install.sh
```diff
@@ -159,7 +159,7 @@ fi
 get_previous_version() {
     repo_url=$1
     # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
-    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name' 
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name'
```
samruddhikhandale, line 162: Can we bump patch version of this Feature?

As a rule of thumb, whenever you contribute changes to any Feature, you always need to bump the Feature versions. Follow semantic versioning rules to determine the version bump.

Failing the bump version will lead into not releasing a new version of the Feature with these changes.

src/kubectl-helm-minikube/install.sh
```diff
@@ -159,7 +159,7 @@ fi
 get_previous_version() {
     repo_url=$1
     # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
```
samruddhikhandale, line 161: Can we update the comment?

test/docker-in-docker/docker_build_fallback_buildx.sh
```diff
@@ -26,7 +26,7 @@ get_latest_version() {
 
 # Function to fetch the previous version of the plugin
 get_previous_version() {
-    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[1].tag_name' # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
+    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name' # this would del the assets key and then get the second encountered tag_name's value from the filtered array of objects
```
samruddhikhandale, line 29: Same here. Can we update the comment?

src/kubectl-helm-minikube/install.sh
gauravsaini04: made changes to the file

test/docker-in-docker/docker_build_fallback_buildx.sh
gauravsaini04: updated the comment

src/kubectl-helm-minikube/install.sh
gauravsaini04: updated the comment 👍

samruddhikhandale (approved):
Nice 🚀

PR #903, opened 2024-03-10
[python] add default formatter

hellodword:


src/python/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.4.1",
+  "version": "1.4.2",
```
samruddhikhandale, line 3: Can we bump the minor version as we are adding in the new extension?

src/python/devcontainer-feature.json
```diff
@@ -73,15 +73,19 @@
     "vscode": {
       "extensions": [
         "ms-python.python",
-        "ms-python.vscode-pylance"
+        "ms-python.vscode-pylance",
+        "ms-python.autopep8"
```
samruddhikhandale, line 77: This sounds good to me, we always supported it when it was part of `ms-python.python`. So these changes make sense to me. 

// cc @bamurtaugh unless there are any concerns.

samruddhikhandale (approved):
Looks good to me!

hoechenberger:
Why would you want to use this instead of the Ruff extension?

samruddhikhandale (approved):
Looks like we missed merging this, approving again ⚡

pickfire:
This broke ruff installation that I had to remove this feature altogether, can this be added behind a feature gate disabled by default? Given that there are quite a few well known alternative, I don't think autopep8 should be enabled by default.

hoechenberger:
@pickfire I added `“-ms-python.autopep8“` to the extensions sections of my dev container config (mind the minus sign, which will remove the extension). And I added a formatted entry to use ruff.

I have to agree that this PR made life harder. Not a fan.

samruddhikhandale:
Okay, revert makes sense in this case. Thanks for bringing this us, feel free to send out a revert PR.

PR #904, opened 2024-03-12
[Node] - nvm - fallback to previous version - code fix

gauravsaini04:
**Feature name**:
 
 * Node
 
 **Description**:
 
 * When the latest tag for `nvm` doesn't have a source binary released yet, then it will continue installing the previous working version for which a source binary will be available
 
 _Changelog_:
 
 * Updated install.sh
 * Updated tests to validate that when a fake `nvm` version is tried to be downloaded with curl command thus representing trying to access a source binary which is not yet deployed by `nvm`, then will pick up the binary for the tag (previous version) which will eventually work successfully.
 * Updated the `devcontainer-feature.json` with an updated patch version for this feature.
 
 **Checklist**:
 
### DIFF
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -293,8 +293,11 @@ set -e
 umask 0002
 # Do not update profile - we'll do this manually
 export PROFILE=/dev/null
-curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash
-
+curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash ||  {
+    PREV_NVM_VERSION=$(curl -s https://api.github.com/repos/nvm-sh/nvm/releases/latest | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
+    curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/\${PREV_NVM_VERSION}/install.sh" | bash
+    NVM_VERSION="\${PREV_NVM_VERSION}"
+}
 source "${NVM_DIR}/nvm.sh"
 if [ "${NODE_VERSION}" != "" ]; then
     nvm alias default "${NODE_VERSION}"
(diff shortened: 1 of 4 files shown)


src/node/install.sh
```diff
@@ -293,8 +293,11 @@ set -e
 umask 0002
 # Do not update profile - we'll do this manually
 export PROFILE=/dev/null
-curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash
-
+curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash ||  {
+    PREV_NVM_VERSION=$(curl -s https://api.github.com/repos/nvm-sh/nvm/releases/latest | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
+    curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/\${PREV_NVM_VERSION}/install.sh" | bash
+    NVM_VERSION="\${PREV_NVM_VERSION}"
+}
```
samruddhikhandale, line 300: Can we move this to a function?

src/node/install.sh
```diff
@@ -293,8 +293,11 @@ set -e
 umask 0002
 # Do not update profile - we'll do this manually
 export PROFILE=/dev/null
-curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash
-
+curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.sh" | bash ||  {
+    PREV_NVM_VERSION=$(curl -s https://api.github.com/repos/nvm-sh/nvm/releases/latest | grep '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')
```
samruddhikhandale, line 297: Instead of the complexity of `sed` , can you use `jq` instead?

samruddhikhandale (approved):
Thanks!

PR #906, opened 2024-03-12
[Powershell] - Fallback to prev. version - fix

gauravsaini04:
**Feature name**:
 
 * Powershell
 
 **Description**:
 
 * When the latest tag for powershell doesn't have a source binary released yet, then it will continue installing the previous working version for which a source binary will be available
 
 _Changelog_:
 
 * Updated install.sh
 * Updated tests to validate that when a fake powershell version is tried to be downloaded with curl command thus representing trying to access a source binary which is not deployed yet, then will pick up the binary for the tag which will eventually work successfully.
 * Updated `devcontainer-feature.json` to update the patch version for this feature. 
 
 **Checklist**:
 
### DIFF
--- a/src/powershell/devcontainer-feature.json
+++ b/src/powershell/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "powershell",
-    "version": "1.3.3",
+    "version": "1.3.4",
     "name": "PowerShell",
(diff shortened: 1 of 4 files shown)


samruddhikhandale (approved):
Looks good, thanks!

PR #907, opened 2024-03-12
[python] - cpython - prev. version fallback - fix

gauravsaini04:
**Feature name**:
 
 * Python
 
 **Description**:
 
 * When the latest tag for `cpython` doesn't have a source binary released yet, then it will continue installing the previous working version for which a source binary will be available
 
 _Changelog_:
 
 * Updated install.sh
 * Updated tests to validate that when a fake `cpython` version is tried to be downloaded with curl command thus representing trying to access a source binary which is not deployed yet, then will pick up the binary for the tag which will eventually work successfully.
 * Updated patch version in `devcontainer-feature.json`
 
 **Checklist**:
 
### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.4.1",
+  "version": "1.4.2",
   "name": "Python",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/python",
   "description": "Installs the provided version of Python, as well as PIPX, and other common Python utilities.  JupyterLab is conditionally installed with the python feature. Note: May require source code compilation.",
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -248,6 +248,48 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
 }
 
+# Use semver logic to decrement a version number then look for the closest match
+find_prev_version_from_git_tags() {
+    local variable_name=$1
+    local current_version=${!variable_name}
+    local repository=$2
+    # Normally a "v" is used before the version number, but support alternate cases
+    local prefix=${3:-"tags/v"}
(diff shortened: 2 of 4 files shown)


samruddhikhandale (approved):
Looks great, one minor comment

samruddhikhandale (dismissed):
Looks great, one minor comment

src/python/install.sh
```diff
@@ -396,23 +454,23 @@ install_from_source() {
...
-    local tgz_filename="Python-${VERSION}.tgz"
-    local tgz_url="https://www.python.org/ftp/python/${VERSION}/${tgz_filename}"
-    echo "Downloading ${tgz_filename}..."
-    curl -sSL -o "/tmp/python-src/${tgz_filename}" "${tgz_url}"
-
+    install_cpython "${VERSION}"
+    if [ -f "/tmp/python-src/${cpython_tgz_filename}" ]; then 
+        if grep -q "404 Not Found" "/tmp/python-src/${cpython_tgz_filename}"; then
+            # Use grep to search for "404 Not Found" in the file
+            echo "\"404 Not Found\" found in /tmp/python-src/${cpython_tgz_filename}. Not able to create source binary"
```
samruddhikhandale, line 461: Looks like this was used for debugging, remove?

src/python/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "python",
-  "version": "1.4.1",
+  "version": "1.4.2",
```
samruddhikhandale, line 3: Looks like tests are failing with `configure: error: Invalid --with-openssl value`, can you help fix them?

src/python/install.sh
gauravsaini04: removed 👍

src/python/devcontainer-feature.json
gauravsaini04: done. not failing now 👍

samruddhikhandale (approved):
Thank you 👏

PR #908, opened 2024-03-12
[Php] - fallback to previous version - code fix

gauravsaini04:
**Feature name**:
 
 * PHP
 
 **Description**:
 
 * When the latest tag for `php` doesn't have a source binary released yet, then it will continue installing the previous working version for which a source binary will be available
 
 _Changelog_:
 
 * Updated install.sh
 * Updated tests to validate that when a fake `php` version is tried to be downloaded with curl command thus representing trying to access a source binary which is not deployed yet, then will pick up the binary for the tag which will eventually work successfully.
 * Updated patch version in `devcontainer-feature.json`
 
 **Checklist**:
 

src/php/install.sh
```diff
@@ -156,6 +197,33 @@ install_php() {
...
+    CONF_DIR="${PHP_INI_DIR}/conf.d"
+    mkdir -p "${CONF_DIR}";
+
+    PHP_EXT_DIR="${PHP_INSTALL_DIR}/extensions"
+    mkdir -p "${PHP_EXT_DIR}"
+
+    PHP_SRC_DIR="/usr/src/php"
+    mkdir -p $PHP_SRC_DIR
+    cd $PHP_SRC_DIR
+    wget -O php.tar.xz "$PHP_URL" || install_previous_version
```
samruddhikhandale, line 226: We use `install_php` to install `version` as well as `additionalVersions`. We don't check that the fallback logic is applied only if the `version` is `latest`. This will lead into discrepancies as we talked before. Can we update the logic?

src/php/install.sh
```diff
@@ -156,6 +197,33 @@ install_php() {
...
+    PHP_INSTALL_DIR="${PHP_DIR}/${PHP_VERSION}"
+    if [ -d "${PHP_INSTALL_DIR}" ]; then
+        echo "(!) PHP version ${PHP_VERSION} already exists."
+        exit 1
+    fi
+
+    if ! cat /etc/group | grep -e "^php:" > /dev/null 2>&1; then
+        groupadd -r php
+    fi
+    usermod -a -G php "${USERNAME}"
```
samruddhikhandale, line 213: This logic is redundant in `install_previous_version` . Can we remove?

test/php/test_php_fallback.sh
```diff
@@ -0,0 +1,170 @@
...
+    mkdir -p "${CONF_DIR}";
+
+    PHP_EXT_DIR="${PHP_INSTALL_DIR}/extensions"
+    mkdir -p "${PHP_EXT_DIR}"
+
+    PHP_SRC_DIR="/usr/src/php"
+    mkdir -p $PHP_SRC_DIR
+    cd $PHP_SRC_DIR
+    wget -O php.tar.xz "$PHP_URL"
+    ln -sf /usr/local/php/8.1.27 /usr/local/php/current
```
samruddhikhandale, line 113: This doesn't seem right which might be causing the test failures in https://github.com/devcontainers/features/actions/runs/8256416491/job/22585068181?pr=908#step:4:26268

Also, isn't the binaries installed within `/usr/src/php`?

test/php/test_php_fallback.sh
```diff
@@ -0,0 +1,170 @@
...
+    mkdir -p "${CONF_DIR}";
+
+    PHP_EXT_DIR="${PHP_INSTALL_DIR}/extensions"
+    mkdir -p "${PHP_EXT_DIR}"
+
+    PHP_SRC_DIR="/usr/src/php"
+    mkdir -p $PHP_SRC_DIR
+    cd $PHP_SRC_DIR
+
+    wget -O php.tar.xz "$PHP_URL" || install_previous_version "8.2"
```
samruddhikhandale, line 163: Why do we pass an argument to install_previous_version? This doesn't seem exactly the same like the function in `install.sh`?

Can we make sure the `install_previous_version` are copies are each other?

samruddhikhandale: Also, shouldn't the VERSION be `latest` in this case, why 8.2?

samruddhikhandale (changes requested):

src/php/install.sh
gauravsaini04: done

test/php/test_php_fallback.sh
gauravsaini04: I was not able to have the test show the previous version after it got installed, something to do with symlinking it seems..

test/php/test_php_fallback.sh
gauravsaini04: corrected !

src/php/install.sh
gauravsaini04: made these changes..

test/php/test_php_fallback.sh
samruddhikhandale: I don't understandy why there would be an issue with symlinking as all the variable are reset when we install it with previous version

https://github.com/devcontainers/features/blob/3c6f07e0f50acfc7fc0d544eaa97ef7b1d676264/src/php/install.sh#L309

samruddhikhandale (changes requested):

src/php/install.sh
```diff
@@ -155,7 +193,26 @@ install_php() {
     PHP_SRC_DIR="/usr/src/php"
     mkdir -p $PHP_SRC_DIR
     cd $PHP_SRC_DIR
-    wget -O php.tar.xz "$PHP_URL"
+}
+
+install_previous_version() {
+    if [[ $PHP_VERSION == "latest" ]]; then
```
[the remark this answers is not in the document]
gauravsaini04: corrected, removed

src/php/install.sh
```diff
@@ -196,15 +196,10 @@ init_php_install() {
...
-        echo -e "\nInstalling Previous Version..."
-        find_prev_version_from_git_tags PHP_VERSION https://github.com/php/php-src "tags/php-"
-        init_php_install
-        wget -O php.tar.xz "$PHP_URL"
-    else 
-        echo "Failed to install PHP v${PHP_VERSION}..."
-        exit 1
-    fi
+    echo -e "\nInstalling Previous Version..."
+    find_prev_version_from_git_tags PHP_VERSION https://github.com/php/php-src "tags/php-"
```
samruddhikhandale, line 200: We still that check where we use find_prev_version_from_git_tags only if user asked for `latest` version.
Using `PHP_VERSION` will give incorrect reasons, hence, you need to add another variable which stores value of `PHP_VERSION` before it gets updated.

test/php/test_php_fallback.sh
gauravsaini04: oh it was already given in code `install.sh` to resolve this. Got it working !

src/php/install.sh
gauravsaini04: done

samruddhikhandale (approved):
⚡

tr4200812 (approved):

Issue #899, opened 2024-03-07
Azure-cli feature is broken for Python 3.12 as the distutils module is removed


PeterBosch2:
How to reproduce?

Setup in a fresh VSCode environment a new DevContainer setup ... e.g., using the wizzard. Select Python on Debian Bookworm, add azure-cli feature. This results in the following devcontainer setup:
```json
{
	"name": "Python 3",
	"image": "mcr.microsoft.com/devcontainers/python:1-3.12-bookworm",
	"features": {
		"ghcr.io/devcontainers/features/azure-cli:1": {}
	}
}
```

Re-opening VSCode in the container works fine, installation of the feature succeeds. However, the `azure-cli` does not work at the command line due to the missing `distutils` package, which is removed in Python 3.12:
```text
vscode ➜ /workspaces/test $ az
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/local/pipx/venvs/azure-cli/lib/python3.12/site-packages/azure/cli/__main__.py", line 39, in <module>
    az_cli = get_default_cli()
             ^^^^^^^^^^^^^^^^^
  File "/usr/local/pipx/venvs/azure-cli/lib/python3.12/site-packages/azure/cli/core/__init__.py", line 917, in get_default_cli
    from azure.cli.core.azlogging import AzCliLogging
  File "/usr/local/pipx/venvs/azure-cli/lib/python3.12/site-packages/azure/cli/core/azlogging.py", line 30, in <module>
    from azure.cli.core.commands.events import EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE
  File "/usr/local/pipx/venvs/azure-cli/lib/python3.12/site-packages/azure/cli/core/commands/__init__.py", line 25, in <module>
    from azure.cli.core.extension import get_extension
  File "/usr/local/pipx/venvs/azure-cli/lib/python3.12/site-packages/azure/cli/core/extension/__init__.py", line 11, in <module>
    from distutils.sysconfig import get_python_lib  # pylint: disable=deprecated-module
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'distutils'
```

samruddhikhandale:
Hi 👋 

[...]

PR #909, opened 2024-03-13
[azure-cli] - python3.12 - feature broken due to distutils module removed - code fix

gauravsaini04:
**Feature name**:
 
 * Azure-cli
 
 **Description**:
 
 This PR fixes the following issue:
 
 * `Azure-cli` feature was broken due to `distutils` not provided as default in Python3.12 [#899](https://github.com/devcontainers/features/issues/899)
 
 _Changelog_:
 
 * Updated `install.sh` to add support for `bookworm`
 * Updated tests to validate that the issue was remediated
 
 **Checklist**:
 

src/azure-cli/install.sh
```diff
@@ -17,8 +17,8 @@ AZ_EXTENSIONS=${EXTENSIONS}
 AZ_INSTALLBICEP=${INSTALLBICEP:-false}
 INSTALL_USING_PYTHON=${INSTALL_USING_PYTHON:-true}
```
samruddhikhandale, line 18: Can we bump the patch version for this Feature?

test/azure-cli/scenarios.json
```diff
@@ -38,5 +38,15 @@
...
       }
     }
+  },
+  "install_with_python_3_12_bookworm": {
+    "image": "mcr.microsoft.com/devcontainers/python:1-3.12-bookworm",
+    "user": "vscode",
+    "features": {
+      "azure-cli": {
+        "version": "latest",
+        "installUsingPython": true
```
samruddhikhandale, line 48: Can we add another scenario to reflect the use case in https://github.com/devcontainers/features/issues/899 ? ie with `installUsingPython: false`. (Else, you could also add `"azure-cli": {}` to the `features` object.

samruddhikhandale, line 48: Currently, this scenario should fail as `az` does not work when installed with python (that's the issue we are trying to solve here). However, looking at the code in [here](https://github.com/devcontainers/features/blob/26883bd6a77a4f5b14a5012f884803eb0c81a766/src/azure-cli/install.sh#L191-L195), it still installs with `apt` and doesn't attempt to install via `python`. This is definitely a conflict between the experience vs the Feature option. 

I understand that this bug is not due to your PR change, however, can you fix the code? If user wants to explicitly install with python then it should not be installed with `apt`.

samruddhikhandale: Once you fix the code, this scenario would start failing. Hence, would be good to update the scenario with `"installUsingPython": false`

samruddhikhandale (changes requested):

test/azure-cli/scenarios.json
samruddhikhandale: Defering this to another PR

samruddhikhandale, line 48:
```suggestion
```

src/azure-cli/install.sh
gauravsaini04: done

test/azure-cli/scenarios.json
gauravsaini04: removed

samruddhikhandale (approved):

hhjmmmmmmmm (approved):

PR #922, opened 2024-03-28
[az-cli] - To separate the two methods of installation - using apt, using python

gauravsaini04:
**Feature name**:
 
 * Azure-cli
 
 **Description**:
 
 This PR attempts to do the following :
 
 * Separates the two installation methods using `apt` and using `python`, such that when `installUsingPython` is set as `true` in `scenarios.json`, it won't be attempting to install using `apt` at all.
 * The above has been done to show that just as mentioned in features issue #[899](https://github.com/devcontainers/features/issue/899), when user will try to use `python:1-3.12-bookworm` base image from `devcontainers` then, it still fails due to absence of `distutils` module from upstream
 
 _Changelog_:
 
 * Updated `installation file`
 * Updated tests
 
 **Checklist**:
 

src/azure-cli/install.sh
```diff
@@ -188,13 +188,11 @@ echo "(*) Installing Azure CLI..."
 . /etc/os-release
 architecture="$(dpkg --print-architecture)"
 CACHED_AZURE_VERSION="${AZ_VERSION}" # In case we need to fallback to pip and the apt path has modified the AZ_VERSION variable.
-if [[ "${AZCLI_ARCHIVE_ARCHITECTURES}" = *"${architecture}"* ]] && [[  "${AZCLI_ARCHIVE_VERSION_CODENAMES}" = *"${VERSION_CODENAME}"* ]]; then
-    install_using_apt || use_pip="true"
+if [ "${INSTALL_USING_PYTHON}" != "true" ]; then
+    if [[ "${AZCLI_ARCHIVE_ARCHITECTURES}" = *"${architecture}"* ]] && [[  "${AZCLI_ARCHIVE_VERSION_CODENAMES}" = *"${VERSION_CODENAME}"* ]]; then
+        install_using_apt
```
samruddhikhandale, line 193: We still want to install with python if apt fails, it would need some refactoring.

> install_using_apt || INSTALL_WITH_PIP

samruddhikhandale (changes requested):

src/azure-cli/install.sh
gauravsaini04: done !

test/azure-cli/scenarios.json
```diff
@@ -47,5 +47,15 @@
         "version": "latest"
       }
     }
+  },
+  "install_using_python_with_python_3_12_bookworm": {
```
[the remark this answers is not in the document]
gauravsaini04: thanks..

src/azure-cli/install.sh
```diff
@@ -188,13 +188,11 @@ echo "(*) Installing Azure CLI..."
 . /etc/os-release
 architecture="$(dpkg --print-architecture)"
 CACHED_AZURE_VERSION="${AZ_VERSION}" # In case we need to fallback to pip and the apt path has modified the AZ_VERSION variable.
-if [[ "${AZCLI_ARCHIVE_ARCHITECTURES}" = *"${architecture}"* ]] && [[  "${AZCLI_ARCHIVE_VERSION_CODENAMES}" = *"${VERSION_CODENAME}"* ]]; then
-    install_using_apt || use_pip="true"
+if [ "${INSTALL_USING_PYTHON}" != "true" ]; then
+    if [[ "${AZCLI_ARCHIVE_ARCHITECTURES}" = *"${architecture}"* ]] && [[  "${AZCLI_ARCHIVE_VERSION_CODENAMES}" = *"${VERSION_CODENAME}"* ]]; then
+        install_using_apt || install_using_pip_strategy
```
samruddhikhandale, line 193: This might not behave correctly as we are missing `AZ_VERSION=${CACHED_AZURE_VERSION}`

Also, would be nice to have the below code when we decide to use pip

```
if [ "$?" != 0 ]; then
        echo "Please provide a valid version for your distribution ${ID} ${VERSION_CODENAME} (${architecture})."
        echo
        echo "Valid versions in current apt-cache"
        apt-cache madison azure-cli | awk -F"|" '{print $2}' | grep -oP '^(.+:)?\K.+'
        exit 1
    fi
  ```
  
  You can continue to use `use_pip` variable if that's easier

samruddhikhandale (changes requested):

samruddhikhandale (approved):

PR #926, opened 2024-03-29
[PowerShell]- Fallback method with find_prev_vers_frm_git_tags() implemented

gauravsaini04:
**Feature name**:
 
 * Powershell
 
 **Description**:
 
 * Added a second installation method using `find_prev_version_from_git_tags()` to install if rate limit for the user accessing GitHub api is reached and installation begins to fail using this fallback method as well..
 
 _Changelog_:
 
 * Updated installation file `install.sh`
 * Updated corresponding tests
 
 **Checklist**:
 
### DIFF
--- a/src/powershell/install.sh
+++ b/src/powershell/install.sh
@@ -106,17 +106,83 @@ install_using_apt() {
     apt-get install -yq powershell${version_suffix} || return 1
 }
 
-# Function to fetch the version released prior to the latest version
-get_previous_version() {
-    repo_url=$1
-    curl -s "$repo_url" | jq -r 'del(.[].assets) | .[0].tag_name'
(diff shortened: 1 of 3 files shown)


src/powershell/install.sh
```diff
@@ -106,17 +106,83 @@ install_using_apt() {
     apt-get install -yq powershell${version_suffix} || return 1
 }
```
samruddhikhandale, line 108: Can we bump the Feature version?

samruddhikhandale (changes requested):
One minor comment and we are all set ⚡

src/powershell/install.sh
gauravsaini04: done 👍

samruddhikhandale (approved):

PR #927, opened 2024-03-29
oryx dotnet 8.0.1 cleanup

prathameshzarkar9:
Feature name:

Oryx
Description:

This PR adds the following:

Removed old version of dotnet 8.0.1 files which are not required anymore to reduce the size of an image by 6 MB
Changelog:

Updated installation file install.sh
 Checked that applied changes work as expected

### DIFF
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -241,6 +241,7 @@ if [[ "${PINNED_SDK_VERSION}" != "" ]]; then
     MAJOR_MINOR_PATCH1_VERSION=${PINNED_SDK_VERSION%??}
     rm -rf /usr/share/dotnet/shared/Microsoft.NETCore.App/$MAJOR_MINOR_PATCH1_VERSION
     rm -rf /usr/share/dotnet/shared/Microsoft.AspNetCore.App/$MAJOR_MINOR_PATCH1_VERSION
+    rm -rf /usr/share/dotnet/templates/$MAJOR_MINOR_PATCH1_VERSION
 fi
 
 
--- a/test/oryx/test_python_project.sh
+++ b/test/oryx/test_python_project.sh
@@ -28,5 +28,7 @@ check "oryx-build-python" oryx build --property python_version="${pythonVersion}
 check "oryx-build-python-installed" python3 -m pip list | grep mpmath
 check "oryx-build-python-result" python3 ./src/solve.py
 
+check "templates/8.0.1-does-not-exist" test ! -d "/usr/share/dotnet/templates/8.0.1"
+
 # Report result
 reportResults
--- a/src/oryx/devcontainer-feature.json
+++ b/src/oryx/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "oryx",
-    "version": "1.3.1",
+    "version": "1.3.2",
     "name": "Oryx",
     "description": "Installs the oryx CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/oryx",


src/oryx/install.sh
```diff
@@ -241,6 +241,7 @@ if [[ "${PINNED_SDK_VERSION}" != "" ]]; then
     MAJOR_MINOR_PATCH1_VERSION=${PINNED_SDK_VERSION%??}
     rm -rf /usr/share/dotnet/shared/Microsoft.NETCore.App/$MAJOR_MINOR_PATCH1_VERSION
```
samruddhikhandale, line 242: Can we bump the patch version for this Feature?
https://github.com/devcontainers/features/blob/0d4771c789f0b007b538aaf0922d96f00d574f9c/src/oryx/devcontainer-feature.json#L3

src/oryx/install.sh
```diff
@@ -241,6 +241,7 @@ if [[ "${PINNED_SDK_VERSION}" != "" ]]; then
...
     rm -rf /usr/share/dotnet/shared/Microsoft.AspNetCore.App/$MAJOR_MINOR_PATCH1_VERSION
+    rm -rf /usr/share/dotnet/templates/$MAJOR_MINOR_PATCH1_VERSION
```
samruddhikhandale, line 244: Can you add a test check to validate that `/usr/share/dotnet/templates/8.0.1` no longer exists?

samruddhikhandale (changes requested):

src/oryx/install.sh
prathameshzarkar9: done

src/oryx/install.sh
prathameshzarkar9: done

test/oryx/test_python_project.sh
```diff
@@ -28,5 +28,13 @@ check "oryx-build-python" oryx build --property python_version="${pythonVersion}
 check "oryx-build-python-installed" python3 -m pip list | grep mpmath
 check "oryx-build-python-result" python3 ./src/solve.py
 
+echo "checking 8.0.1 folder is not exist:"
+dirname=/usr/share/dotnet/templates/8.0.1 
+if [ -d "$dirname" ]; then
+    echo "$dirname directory exists"
+else
+    echo "$dirname directory does not exist."
+fi
```
samruddhikhandale, line 37:
```suggestion
check "templates/8.0.1-does-not-exist" ls /usr/share/dotnet/templates/8.0.1 bash -c "test ! -d /usr/share/dotnet/templates/8.0.1"
```

samruddhikhandale (changes requested):

test/oryx/test_python_project.sh
```diff
@@ -28,5 +28,7 @@ check "oryx-build-python" oryx build --property python_version="${pythonVersion}
 check "oryx-build-python-installed" python3 -m pip list | grep mpmath
 check "oryx-build-python-result" python3 ./src/solve.py
 
+check "templates/8.0.1-does-not-exist" ls /usr/share/dotnet/templates/8.0.1 bash -c "test ! -d /usr/share/dotnet/templates/8.0.1"
```
samruddhikhandale, line 31: Hmm.. This isn't working as how I tested in my container. 
Never mind, let's remove this check.

prathameshzarkar9: Now this is working. I have tested in my container and showing correct check. failing if the folder exists and passing if it is not exists.

samruddhikhandale (approved):
Nice!

PR #931, opened 2024-04-02
[Ruby]- rvm - fallback code fix

gauravsaini04:
**Feature name**:
 
 * Ruby
 
 **Description**:
 
 * Added a second installation method using `find_prev_version_from_git_tags()` to install if rate limit for the user accessing GitHub api is reached and installation begins to fail using this fallback method as well
 
 _Changelog_:
 
 * Updated installation file `install.sh`
 * Updated corresponding test
 
 **Checklist**:
 
### DIFF
--- a/src/ruby/install.sh
+++ b/src/ruby/install.sh
@@ -30,8 +30,6 @@ keyserver hkp://keyserver.pgp.com"
 
 set -e
 
-trap 'echo "Last executed command failed at line ${LINENO}"' ERR
-
 # Clean up
 rm -rf /var/lib/apt/lists/*
 
--- a/src/ruby/install.sh
+++ b/src/ruby/install.sh
@@ -218,7 +218,7 @@ fi
 get_previous_version() {
     local url=$1
     local repo_url=$2
-    local variable_name=$3
+    variable_name=$3
     prev_version=${!variable_name}
     
     output=$(curl -s "$repo_url");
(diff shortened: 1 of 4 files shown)


src/ruby/install.sh
```diff
@@ -173,25 +214,44 @@ if ! type git > /dev/null 2>&1; then
...
+
+get_github_api_repo_url() {
+    local url=$1
+    echo "${url/https:\/\/github.com/https:\/\/api.github.com\/repos}/releases/latest"
+}
+
 
 # Figure out correct version of a three part version number is not passed
-find_version_from_git_tags RUBY_VERSION "https://github.com/ruby/ruby" "tags/v" "_"
+ruby_url="https://github.com/ruby/ruby"
```
samruddhikhandale, line 251:
```suggestion
RUBY_URL="https://github.com/ruby/ruby"
```

src/ruby/install.sh
```diff
@@ -210,19 +270,46 @@ else
             DEFAULT_GEMS=""
         fi
     fi
+}
+
+install_previous_version() {
+    repo_url=$(get_github_api_repo_url "$ruby_url")
+    get_previous_version "${ruby_url}" "${repo_url}" RUBY_VERSION
+    set_rvm_install_args
```
samruddhikhandale, line 278: I don't think we need to re-call this? as it doesn't use VERSION

Hence, let's not isolate to a function and keep code intact.

samruddhikhandale (changes requested):

src/ruby/install.sh
gauravsaini04: actually, i have now isolated the `original_ruby_version` if given as "latest" from the `ruby_version` found after `find_version_from_git_tags()` fn, so, in `install_previous_version`, it first checks whether user wanted to install the "latest" version, if yes, then it would install the prev. version if source binary for `ruby_version` for latest requested was not found. so, we do need to add it as a separate fn entity so that it can be called for two different versions got initially from `find_version_from_git_tags` and then later from `get_previous_version`

src/ruby/install.sh
gauravsaini04: done

src/ruby/install.sh
samruddhikhandale: `set_rvm_install_args` only checks if the VERSION is `none` or not. In both cases, latest and prev-of-latest, the VERSION will never be none? 

So we can remove set_rvm_install_args() and keep the flow as is?

gauravsaini04: https://github.com/devcontainers/features/blob/main/src/ruby/install.sh#L218

it is used here, and both the times, its the different version for ruby_version variable and then attempting the above as well, so that's why tried to isolate it so can prevent code reuse. 
But if any other way, let me know. 👍🏻

samruddhikhandale (approved):
Looks good!

hhjmmmmmmmm (approved):

hhjmmmmmmmm (approved):

Issue #941, opened 2024-04-11
Fix tests: nvidia-cuda (liburcu6)


samruddhikhandale:
> Looking at the test failures, looks like we need to install `liburcu6` based on the distro type. 
> Also, I am seeing similar errors on the `main` branch ([ref](https://github.com/devcontainers/features/actions/runs/8633868409/job/23667938174)). This PR does not impact the test failures, hence, I can track them in another issue. 
> 
> >  looks like we need to install `liburcu6` based on the distro type. 
> 
> Can you help revert the `liburcu6` change or feel free to help us with ^

_Originally posted by @samruddhikhandale in https://github.com/devcontainers/features/pull/933#discussion_r1559685937_

samruddhikhandale:
@gauravsaini04 Can you help fix the tests for nvidia-cuda Feature? Thanks! https://github.com/devcontainers/features/actions/runs/8633868409/job/23667938174

prathameshzarkar9:
Hi @samruddhikhandale ,
I have raised a PR to fix this. 
[PR 942](https://github.com/devcontainers/features/pull/942)
`liburcu6` package was not installable so I have used dpkg to download deb package and install. After the changes, test case is passed.
let me know if any change/improvement is need to fix this.

prathameshzarkar9:
Hi @samruddhikhandale ,
Updated the PR with the changes suggested. Test cases are passing.

PR #942, opened 2024-04-12
upgrade cuda version to 11.7 and cudnn to 8.5.0

prathameshzarkar9:
**Dev container name**:
 * nvidia-cuda

 **Description**:
 [Issue 941](https://github.com/devcontainers/features/issues/941)
 This PR incorporate the following changes:
 * liburcu6 is installed manually with the dpkg.

**_Changelog_**:
 * Updated install.sh
   * added change to download deb package and install using `dpkg -i`

 **Checklist**:
 * [x]   Test case `install_cudnn_nvxt_version` passed.

### DIFF
--- a/src/nvidia-cuda/install.sh
+++ b/src/nvidia-cuda/install.sh
@@ -12,6 +12,8 @@ INSTALL_TOOLKIT=${INSTALLTOOLKIT}
 CUDA_VERSION=${CUDAVERSION}
 CUDNN_VERSION=${CUDNNVERSION}
 
+. /etc/os-release 
+
 if [ "$(id -u)" -ne 0 ]; then
     echo -e 'Script must be run as root. Use sudo, su, or add "USER root" to your Dockerfile before running this script.'
     exit 1
@@ -32,6 +34,11 @@ check_packages() {
         apt-get -y install --no-install-recommends "$@"
     fi
 }
+
+if [ $VERSION_CODENAME = "bookworm" ] || [ $VERSION_CODENAME = "jammy" ] && [ $CUDA_VERSION \< 11.7 ]; then  
+    echo "(!) Unsupported distribution version '${VERSION_CODENAME}' for CUDA < 11.7"
+    exit 1
+fi  
 
 export DEBIAN_FRONTEND=noninteractive
 
--- a/src/nvidia-cuda/devcontainer-feature.json
+++ b/src/nvidia-cuda/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "nvidia-cuda",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "name": "NVIDIA CUDA",
   "description": "Installs shared libraries for NVIDIA CUDA.",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/nvidia-cuda",
--- a/test/nvidia-cuda/install_cudnn_nvxt_version.sh
+++ b/test/nvidia-cuda/install_cudnn_nvxt_version.sh
@@ -5,11 +5,11 @@ set -e
 # Optional: Import test library
 source dev-container-features-test-lib
 
-# Check installation of libcudnn8 (8.3.2)
-check "libcudnn.so.8.3.2" test 1 -eq "$(find /usr -name 'libcudnn.so.8.3.2' | wc -l)"
+# Check installation of libcudnn8 (8.5.0)
+check "libcudnn.so.8.5.0" test 1 -eq "$(find /usr -name 'libcudnn.so.8.5.0' | wc -l)"
 
-# Check installation of cuda-nvtx-11-5 (11.5)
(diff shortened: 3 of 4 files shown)


src/nvidia-cuda/install.sh
```diff
@@ -33,6 +33,15 @@ check_packages() {
     fi
 }
```
samruddhikhandale, line 34: Can you bump the patch version of the Feature?

src/nvidia-cuda/install.sh
```diff
@@ -33,6 +33,15 @@ check_packages() {
...
 
+install_debian_package() {
```
samruddhikhandale, line 36: This approach is a bit complex and doesn't handle dependencies automatically. Hence, this might fail down the line.

Also, `liburcu6` is not available in `bookworm` and `jammy`. Hence, we shouldn't attempt to force install it.

Taking another look at this issue, looks like the root cause of the test failures is https://github.com/devcontainers/features/blob/391ab32f6969c904133a2eeb845192ef63dc60e2/test/nvidia-cuda/scenarios.json#L19

Apparently for `bookworm` and `jammy`, one needs a cuda version >= 11.7.

Can you remove your changes and update the test with `cudaVersion:11.7` ? Let's see if this works! 

Ref: https://askubuntu.com/questions/1407962/unable-to-install-cuda-on-ubuntu-22-04-wsl2

src/nvidia-cuda/install.sh
```diff
@@ -33,6 +33,15 @@ check_packages() {
...
+    local url=$1
```
samruddhikhandale, line 37: If the tests pass, then can you add following code?

```
// psuedocode
If [ distro is `bookworm` and `jammy`] & [ cuda version < 11. 7 ]  ; then
    echo "(!) Unsupported distribution version '${VERSION_CODENAME}' for CUDA < 11.7"
    exit 1
fi
```

samruddhikhandale (changes requested):

src/nvidia-cuda/install.sh
prathameshzarkar9: done

src/nvidia-cuda/install.sh
prathameshzarkar9: done, removed the complex code as per the suggestion.

src/nvidia-cuda/install.sh
prathameshzarkar9: done. Added the correct code.

src/nvidia-cuda/install.sh
```diff
@@ -12,6 +12,9 @@ INSTALL_TOOLKIT=${INSTALLTOOLKIT}
 CUDA_VERSION=${CUDAVERSION}
 CUDNN_VERSION=${CUDNNVERSION}
 
+. /etc/os-release 
+
+
```
samruddhikhandale, line 17:
```suggestion
```

src/nvidia-cuda/install.sh
```diff
@@ -33,6 +36,11 @@ check_packages() {
     fi
 }
 
+if [ $VERSION_CODENAME = `bookworm` || $VERSION_CODENAME = `jammy`] && [ $CUDA_VERSION < 11.7 ]; then
```
samruddhikhandale, line 39: Can you fix https://github.com/devcontainers/features/actions/runs/8666785264/job/23768360746?pr=942#step:4:125

samruddhikhandale (changes requested):
Can you fix

src/nvidia-cuda/install.sh
prathameshzarkar9: done

src/nvidia-cuda/install.sh
prathameshzarkar9: done

src/nvidia-cuda/install.sh
samruddhikhandale: It's still failing https://github.com/devcontainers/features/actions/runs/8667039013/job/23769166488?pr=942#step:4:125

prathameshzarkar9: done

src/nvidia-cuda/install.sh
```diff
@@ -33,6 +35,11 @@ check_packages() {
     fi
 }
 
+if [ $VERSION_CODENAME = `bookworm` ] || [ $VERSION_CODENAME = `jammy` ] && [ $CUDA_VERSION \< 11.7 ]; then
```
samruddhikhandale, line 38:
```suggestion
if [ $VERSION_CODENAME = "bookworm" ] || [ $VERSION_CODENAME = "jammy" ] && [ $CUDA_VERSION \< 11.7 ]; then  
```

samruddhikhandale (changes requested):

src/nvidia-cuda/install.sh
prathameshzarkar9: done

samruddhikhandale (approved):

PR #943, opened 2024-04-13
[anaconda] support ARM architecture and install latest version

r3dpoint:
- The anaconda installer is available for ARM and dev containers work on ARM; therefore, elaborate on architecture detection in the anaconda feature to support ARM.
- Update to default to latest version of anaconda.

### DIFF
--- a/src/anaconda/install.sh
+++ b/src/anaconda/install.sh
@@ -47,7 +47,15 @@ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then
 fi
 
 architecture="$(uname -m)"
-if [ "${architecture}" != "x86_64" ]; then
+if [ "${architecture}" == "amd64" ]; then
+    CONDA_ARCH="x86_64"
+elif [ "${architecture}" == "x86_64" ]; then
+    CONDA_ARCH="x86_64"
+elif [ "${architecture}" == "arm64" ]; then
+    CONDA_ARCH="aarch64"
+elif [ "${architecture}" == "aarch64" ]; then
+    CONDA_ARCH="aarch64"
+else
     echo "(!) Architecture $architecture unsupported"
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


r3dpoint:

src/anaconda/install.sh
```diff
@@ -47,7 +47,15 @@ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then
...
+elif [ "${architecture}" == "x86_64" ]; then
+    CONDA_ARCH="x86_64"
+elif [ "${architecture}" == "arm64" ]; then
+    CONDA_ARCH="aarch64"
+elif [ "${architecture}" == "aarch64" ]; then
+    CONDA_ARCH="aarch64"
+else
     echo "(!) Architecture $architecture unsupported"
     exit 1
 fi
```
samruddhikhandale, line 61:
```suggestion
case "${architecture}" in
        amd64) architecture=x86_64 ;;
        arm64) architecture=aarch64 ;;
        *)
            echo "Anaconda does not support machine architecture '$architecture'. Please use an x86-64 or ARM64 machine."
            exit 1
    esac
```

How about we simplify the code?

src/anaconda/install.sh
```diff
@@ -94,11 +102,11 @@ if ! conda --version &> /dev/null ; then
 
     CONDA_VERSION=$VERSION
     if [ "${VERSION}" = "latest" ] || [ "${VERSION}" = "lts" ]; then
-        CONDA_VERSION="2021.11"
+        CONDA_VERSION="2024.02-1"
     fi
 
     su --login -c "export http_proxy=${http_proxy:-} && export https_proxy=${https_proxy:-} \
-        && wget -q https://repo.anaconda.com/archive/Anaconda3-${CONDA_VERSION}-Linux-x86_64.sh -O /tmp/anaconda-install.sh \
+        && wget -q https://repo.anaconda.com/archive/Anaconda3-${CONDA_VERSION}-Linux-${CONDA_ARCH}.sh -O /tmp/anaconda-install.sh \
```
samruddhikhandale, line 109:
```suggestion
        && wget -q https://repo.anaconda.com/archive/Anaconda3-${CONDA_VERSION}-Linux-${architecture}.sh -O /tmp/anaconda-install.sh \
```

samruddhikhandale (changes requested):
Thank you so much for contributing these changes, they look great. 

Left some minor comments!

src/anaconda/install.sh
r3dpoint: happy to implement though as i read your suggestion it accomplishes a different goal. the goal is to support amd64, x86_64, arm64, and aarch64. and in the case of amd64 use the x86_64 URL and in the case of the arm64 use the aarch64 URL

samruddhikhandale: I was hoping to keep the goal in tact but didn't realize that we are using `architecture="$(uname -m)"` to compute `architecture`. Thanks for catching that! 

How about we use 👇 instead? That way I believe it only returns `amd64` of `arm64`. Hence, we'd skip another conditions. 

```
architecture=$(dpkg --print-architecture)
```

r3dpoint: nice that makes sense. i'll work up a commit.

PR #950, opened 2024-04-17
[Node]-NVM-Fallback scenario added for 2 way fallback

gauravsaini04:
**Feature name**:
 
 * Node
 
 **Description**:
 
 * When the latest tag for `nvm` doesn't have a source binary released yet, then it will continue installing the previous working version for which a source binary will be available
 
 _Changelog_:
 
 * Updated `install.sh`
 * Updated tests to validate that when a fake `nvm` version is tried to be downloaded with curl command thus representing trying to access a source binary which is not deployed yet, then will pick up the binary for the tag which will eventually work successfully.
 
 **Checklist**:
 
### DIFF
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -387,8 +387,6 @@ elif [ "${NODE_VERSION}" = "latest" ]; then
     export NODE_VERSION="node"
 fi
 
-trap 'echo "Error on line $LINENO"' ERR
-
 # Install snipppet that we will run as the user
 nvm_install_snippet="$(cat << EOF
     set -e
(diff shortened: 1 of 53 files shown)


src/node/install.sh
```diff
@@ -187,6 +188,103 @@ find_version_from_git_tags() {
...
+)
+
+target_directory="/tmp/nvm_installation_script"
+mkdir -p "$target_directory"
+
+# Define your target script file
+target_script="$target_directory/main_script.sh"
+
+# Append the functions content to the target script
+echo "$functions_content" > "$target_script"
```
samruddhikhandale, line 286: Can we rename the functions and file names to be more descriptive?

samruddhikhandale (changes requested):

Issue #961, opened 2024-05-06
bug(rust): Fails to install 


bearrito:
Here is my devcontainer.json

```
{
	"name": "Ubuntu",
	"build": {
		"dockerfile": "Dockerfile"
	},
	"features": {
		"ghcr.io/devcontainers/features/rust:latest": {}
	},
	"postCreateCommand": [
		".devcontainer/install_corrosion.sh",
		"cargo install cargo-deb"
	]
}
```

Log output

```

11.23 Installing Rust...
11.83 sha256sum: ./rustup-init: No such file or directory
11.83 sha256sum: WARNING: 1 listed file could not be read
[2024-05-06T20:33:07.604Z] 11.83 ./rustup-init: FAILED open or read
11.83 ERROR: Feature "Rust" (ghcr.io/devcontainers/features/rust) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/rust for help troubleshooting this error.
------
Dockerfile-with-features:28
```

andrewn617:
I'm also encountering this.

samruddhikhandale:
I am able to reproduce the error as well. At first glance, it looks like an upstream issue. Opened https://github.com/rust-lang/rustup/issues/3809

bearrito:
The bug seems to be in `install.sh` I have a fix ready if you want it?

The gist is the `rustup-init` binary isn't landing in the same directory as the checksum file.


```
curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init" -o /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init
curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init.sha256" -o /tmp/rustup/rustup-init.sha256
cp /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init  /tmp/rustup/rustup-init 
echo "Checking sum"
sha256sum -c rustup-init.sha256
```

samruddhikhandale:
@bearrito Thanks for the quick analysis. Do you mind creating a PR?

bearrito:
@samruddhikhandale Doing it now. I'll have it out in about 15 mins

samruddhikhandale:
Thanks @bearrito for the quick fix, I have left minor comments.

PR #962, opened 2024-05-06
Fix rustup-init sha256sum check

bearrito:
Description:

`sha256sum rustup-init.sha256sum` expects there to be a file name `rustup-init` in the same directory as ` rustup-init.sha256sum`
Instead of messing with the original download path, I just copy the binary to the path where the sha256sum is placed, check it and then delete it. There might be a more efficient way with pipes or similar. 

Tests passing on my machine
```
Test Passed!
🧪 Executing duplicate test for feature 'rust'...
⚠️ Skipping duplicate test for rust because '/home/bstrausser/Git/features/test/rust/duplicate.sh' does not exist.
🧹 Cleaning up 2 test containers...
🧹 Removing container a7c1a0e866f3...
🧹 Removing container 01a2b8770ce2...



✅ Passed:      'rust'
✅ Passed:      'rust_at_pinned_version'

```

closes: https://github.com/devcontainers/features/issues/961

### DIFF
--- a/src/rust/install.sh
+++ b/src/rust/install.sh
@@ -186,7 +186,9 @@ else
     curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init" -o /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init
     curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init.sha256" -o /tmp/rustup/rustup-init.sha256
     cd /tmp/rustup
+    cp /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init  /tmp/rustup/rustup-init
     sha256sum -c rustup-init.sha256
+    rm /tmp/rustup/rustup-init
     chmod +x target/${download_architecture}-unknown-linux-gnu/release/rustup-init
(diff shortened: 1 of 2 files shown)


src/rust/install.sh
```diff
@@ -186,7 +186,9 @@ else
     curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init" -o /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init
```
samruddhikhandale, line 186: Can you bump the patch version for this Feature? Thanks!

https://github.com/devcontainers/features/blob/75602ab7eee94992198fbdb1a3b0d52beb62399a/src/rust/devcontainer-feature.json#L3

src/rust/install.sh
```diff
@@ -186,7 +186,9 @@ else
...
     curl -sSL --proto '=https' --tlsv1.2 "https://static.rust-lang.org/rustup/dist/${download_architecture}-unknown-linux-gnu/rustup-init.sha256" -o /tmp/rustup/rustup-init.sha256
     cd /tmp/rustup
+    cp /tmp/rustup/target/${download_architecture}-unknown-linux-gnu/release/rustup-init  /tmp/rustup/rustup-init
     sha256sum -c rustup-init.sha256
+    rm /tmp/rustup/rustup-init
```
samruddhikhandale, line 191:
```suggestion
```

This file would eventually get cleaned up in L195

samruddhikhandale (changes requested):
Left minor comments, appreciate the quick fix 🙌

bearrito:
@samruddhikhandale Addressed your comments.

bearrito:

PR #969, opened 2024-05-12
Support docker-in-docker on Ubuntu Noble (24.04)

jjh42:
This tiny changed seem to work for me.

https://github.com/jjh42/docker-in-docker-noble

as a temporary workaround you can use
      "ghcr.io/jjh42/docker-in-docker-noble/docker-in-docker:latest": {},
instead of the official docker-in-docker feature.

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -18,8 +18,8 @@ USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
 INSTALL_DOCKER_BUILDX="${INSTALLDOCKERBUILDX:-"true"}"
 INSTALL_DOCKER_COMPOSE_SWITCH="${INSTALLDOCKERCOMPOSESWITCH:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
-DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy"
-DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy"
+DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy noble"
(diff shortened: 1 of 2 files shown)


src/docker-in-docker/install.sh
```diff
@@ -18,8 +18,8 @@ USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
 INSTALL_DOCKER_BUILDX="${INSTALLDOCKERBUILDX:-"true"}"
 INSTALL_DOCKER_COMPOSE_SWITCH="${INSTALLDOCKERCOMPOSESWITCH:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
-DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy"
-DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy"
+DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy noble"
+DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy noble"
```
samruddhikhandale, line 22: Can you help write a similar test 👇 for the `noble` image? Thanks! 
https://github.com/devcontainers/features/blob/18c182385da24fde37768ea187612aa5e5063d57/test/docker-in-docker/scenarios.json#L2-L10

samruddhikhandale (changes requested):
This makes sense, left one minor comment. Thanks!

src/docker-in-docker/install.sh
samruddhikhandale: Better idea, I think we should update the 👇 workflow and add the `mcr.microsoft.com/devcontainers/base:noble` image to the `baseImage` 

https://github.com/devcontainers/features/blob/18c182385da24fde37768ea187612aa5e5063d57/.github/workflows/test-all.yaml#L43-L51

samruddhikhandale:
Thank you so much @jjh42 for the PR, we appreciate it ✨ 

Closing this PR as we have merged in the fix with https://github.com/devcontainers/features/pull/971

PR #982, opened 2024-05-26
[desktop-lite] Allow password-less VNC connections

### LINKED ISSUE #611: [desktop-lite] Option to have password-less, autoconnecting noVNC
For a local devcontainer with the noVNC only accessible from localhost, it would be great to have an option to remove the password and configure noVNC to automatically connect. It could alternatively leave the password for the TigerVNC but just configure noVNC to automatically supply the password. However, I think keepping the password-iness of the web and VNC in sync would be good.
To turn off the password in tigerVNC, you can just add `-SecurityTypes None` to the command launching the server. I have not looked into noVNC configuration to automatically connect or a default password beside the URL params.

Currently to automatically connect and use the password, I have a launch configuration to launch `http://localhost:6080/?autoconnect=true&password=vscode` but this is suboptimal.

hoechenberger:
This allows to use:

```json
    "ghcr.io/devcontainers/features/desktop-lite:1": {
      "password": "noPassword"
    }
```

to start the VNC server such that it won't require authentication when trying to connect.

Closes #611

### DIFF
--- a/src/desktop-lite/README.md
+++ b/src/desktop-lite/README.md
@@ -16,10 +16,10 @@ Adds a lightweight Fluxbox based desktop to the container that can be accessed u
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Currently Unused! | string | latest |
-| noVncVersion | NoVnc Version | string | 1.2.0 |
-| password | Enter a password for desktop connections | string | vscode |
-| webPort | Enter a port for the VNC web client | string | 6080 |
-| vncPort | Enter a port for the desktop VNC server | string | 5901 |
+| noVncVersion | The noVNC version to use | string | 1.2.0 |
+| password | Enter a password for desktop connections. If `noPassword`, connections from the local host can be established without entering a password | string | vscode |
+| webPort | Enter a port for the VNC web client (noVNC) | string | 6080 |
+| vncPort | Enter a port for the desktop VNC server (TigerVNC) | string | 5901 |
 
 ## Connecting to the desktop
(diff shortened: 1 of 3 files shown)


samruddhikhandale (changes requested):
Thanks for taking the time in contributing this PR. The changes make sense to me ✨ 

Left some minor comments

hoechenberger:
Thanks @samruddhikhandale, I've adopted your proposal (needed to apply a small fix to make it work)

samruddhikhandale (approved):
Looks great, thank you!

hoechenberger:
Thank you, @samruddhikhandale!

One question: The feature has apparently indeed been built and published. However, the bot didn't add a new tag on GH. Is this expected?

samruddhikhandale:
@hoechenberger Which bot are you mentioning here, where do you expect the new tag? 

https://github.com/devcontainers/features/pkgs/container/features%2Fdesktop-lite looks good to me as you already mentioned.

hoechenberger:
@samruddhikhandale I don't know which bot would be responsible, but I see that there are numerous tags for (older?) versions of many features in the repo, see: https://github.com/devcontainers/features/tags

samruddhikhandale:
> @samruddhikhandale I don't know which bot would be responsible, but I see that there are numerous tags for (older?) versions of many features in the repo, see: https://github.com/devcontainers/features/tags

Got it, @joshspicer do you mind reminding us why we saw tags before, but we stopped couple of months ago? I wonder if this was intentional or a regression.

PR #987, opened 2024-05-29
[Java] - Document additionalVersions functionality

gauravsaini04:
**Feature name**:
 
 * Java
 
 **Description**:
 
 This PR documents the following existing functionality :
 
 * `Java` feature already had the functionality for installing `additional versions` that will be kept other than the `version` property's default installed version
 
 _Changelog_:
 
 * Updated `Readme.md` file

### DIFF
--- a/src/java/README.md
+++ b/src/java/README.md
@@ -16,6 +16,7 @@ Installs Java, SDKMAN! (if not installed), and needed dependencies.
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Select or enter a Java version to install | string | latest |
+| additionalVersions | Comma separated list of additional versions to install | string | empty |
 | jdkDistro | Select or enter a JDK distribution | string | ms |
 | installGradle | Install Gradle, a build automation tool for multi-language software development | boolean | false |
 | gradleVersion | Select or enter a Gradle version | string | latest |
--- a/src/java/README.md
+++ b/src/java/README.md
@@ -16,7 +16,6 @@ Installs Java, SDKMAN! (if not installed), and needed dependencies.
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Select or enter a Java version to install | string | latest |
-| additionalVersions | Comma separated list of additional versions to install | string | empty |
 | jdkDistro | Select or enter a JDK distribution | string | ms |
 | installGradle | Install Gradle, a build automation tool for multi-language software development | boolean | false |
 | gradleVersion | Select or enter a Gradle version | string | latest |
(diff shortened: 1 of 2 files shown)


src/java/README.md
```diff
@@ -16,6 +16,7 @@ Installs Java, SDKMAN! (if not installed), and needed dependencies.
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
 | version | Select or enter a Java version to install | string | latest |
+| additionalVersions | Comma separated list of additional versions to install | string | empty |
```
samruddhikhandale, line 19: We need to expose `additionalVersions` as a Feature option.

samruddhikhandale (changes requested):

src/java/README.md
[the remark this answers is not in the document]
gauravsaini04: tests are already in place. tests like `install_additional_java`.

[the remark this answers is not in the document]
gauravsaini04: I have reverted changes to this file.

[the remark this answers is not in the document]
gauravsaini04: what changes need to be made in this pr, changing readme.md was the only change and tests are already written for installing additional versions of java ?

gauravsaini04: it's already exposed it seems. Pls do check the tests written for it in the scenario - `install_additional_java`

samruddhikhandale: The option is created, but it is not listed in https://github.com/devcontainers/features/blob/main/src/java/devcontainer-feature.json

Hence, it won't be exposed to other IDEs. Instead of directly updating the README, we'd need to update https://github.com/devcontainers/features/blob/main/src/java/devcontainer-feature.json which publicizes the option.

gauravsaini04: thanks for elaborating

src/java/devcontainer-feature.json
```diff
@@ -17,6 +17,14 @@
       "default": "latest",
       "description": "Select or enter a Java version to install"
     },
+    "additionalVersions": {
+      "type": "string",
+      "proposals": [
+        "17, 8"
```
samruddhikhandale, line 23: Can we remove `proposals` property?

src/java/devcontainer-feature.json
```diff
@@ -17,6 +17,14 @@
...
+      ],
+      "default": "",
+      "description": "Select or enter additional Java versions to install"
```
samruddhikhandale, line 26:
```suggestion
      "description": "Enter additional Java versions, separated by commas."
```

samruddhikhandale (changes requested):

src/java/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.4.1",
+  "version": "1.4.2",
```
samruddhikhandale, line 3:
```suggestion
  "version": "1.5.0",
```

samruddhikhandale (approved):

PR #988, opened 2024-05-29
[Node] - pnpmVersion as option added

gauravsaini04:
**Feature name**:
 
 * Node
 
 **Description**:
 
 This PR introduces the following functionality :
 
 * `pnpnVersion` is added as an option as it has the compatibility requirements with specific node versions
 
 _Changelog_:
 
 * Updated `install.sh`
 * Updated tests to validate the above
 
 **Checklist**:
 

src/node/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.4.1",
+    "version": "1.4.2",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.5.0",
```

src/node/install.sh
```diff
@@ -16,6 +16,9 @@ INSTALL_TOOLS_FOR_NODE_GYP="${NODEGYPDEPENDENCIES:-true}"
 # alongside NODE_VERSION, but not set as default.
 ADDITIONAL_VERSIONS="${ADDITIONALVERSIONS:-""}"
 
+#pnpm Version
```
samruddhikhandale, line 19: remove?

src/node/install.sh
```diff
@@ -16,6 +16,9 @@ INSTALL_TOOLS_FOR_NODE_GYP="${NODEGYPDEPENDENCIES:-true}"
...
+PNPM_VERSION="${PNPMVERSION:-"7"}"
```
samruddhikhandale, line 20:
```suggestion
PNPM_VERSION="${PNPMVERSION:-"latest"}"
```

Let's install latest version when use does not pass a version

src/node/install.sh
```diff
@@ -377,7 +380,7 @@ else
             [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
             [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
             [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
-            npm install -g pnpm
+            npm install -g pnpm@"${PNPM_VERSION}"
```
samruddhikhandale, line 383: Need updates when once you resolve my previous comment

samruddhikhandale (changes requested):

src/node/devcontainer-feature.json
gauravsaini04: done

src/node/install.sh
gauravsaini04: done

PR #994, opened 2024-06-04
Ruby: Install `rustc` to enable `yjit`

runephilosof-abtion:
Ruby 3.2 will install with `yjit` enabled when `rustc` is installed.

### DIFF
--- a/src/ruby/install.sh
+++ b/src/ruby/install.sh
@@ -209,7 +209,7 @@ fi
 # Install dependencies
 check_packages curl ca-certificates software-properties-common build-essential gnupg2 libreadline-dev \
     procps dirmngr gawk autoconf automake bison libffi-dev libgdbm-dev libncurses5-dev \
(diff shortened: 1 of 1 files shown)


andrewn617:
Hi @samruddhikhandale I saw this issue an wanted to clarify a few things.

YJIT is no longer experimental. In Rails 7.2, when running ruby 3.3, YJIT [is enabled by default](https://github.com/rails/rails/pull/49947).

It's not exactly right to say that YJIT will be disabled when `rustc` is not installed - YJIT will not be installed at all, so it cannot be enabled. With `rustc` installed, it will give user the option to use YJIT if they choose (or if it is enabled by default for their rails/ruby version). So IMO it would be better to have `rustc` installed always, otherwise the user cannot benefit from yjit.

eitsupi:
Just a question: I am not familiar with the Ruby source code, but I wonder if installing Rust in the OS package manager is really desirable here.
I mean, is Ruby's MSRV small enough and is a few years old Rustc like Debian stable provides enough?

If not, then perhaps we need to use rustup and configure it to install the latest rustc.

PR #997, opened 2024-06-10
dotnet: add ability to install workloads

### LINKED ISSUE #996: [dotnet] install workloads
Hi,

I was playing with Aspire, which is shipped as a dotnet CLI workload, an optional download that is not included in the base SDK. I would like the ability to specify workloads to be installed in my devcontainer.json file.

``` diff
"ghcr.io/devcontainers/features/dotnet": {
  "version": "8.0",
+  "workloads": "aspire"
}
```

Multiple workloads would be implemented as a comma-separated list.

``` json
  "workloads": "aspire, wasm-tools"
```

The install script should then run `dotnet workload install` on each of them, after finishing the installation of the base SDK / dotnet CLI.

## Additional context

`dotnet workload install` requires root.

image

Running the command as root triggers the "first time" experience, which is possibly bad. For example, it generates an HTTPS development certificate, which should never become a part of the container image, because there are security implications regarding pre-built images which might be shared publicly.

This behavior should be disabled by setting `DOTNET_GENERATE_ASPNET_CERTIFICATE=false`
https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_generate_aspnet_certificate

Since the output is quite noisy, maybe `DOTNET_NOLOGO=true` should also be set.

image

sliekens:
This PR adds a new configuration item `"workloads"` for dotnet, which is a comma-separated list of .NET workload ids.

Run `dotnet workload search` to see which workloads are available to install. Note that some workloads like MAUI cannot be installed in Ubuntu.

This is a non-breaking change, I increased only the minor version of the feature.

(I also improved the README a little bit while adding a new example.)

Resolves #996

### DIFF
--- a/src/dotnet/README.md
+++ b/src/dotnet/README.md
@@ -31,19 +31,18 @@ This Feature installs the latest .NET SDK, which includes the .NET CLI and the s
 Installing only the latest .NET SDK version (the default).
 
 ``` json
-{
 "features": {
-    "ghcr.io/devcontainers/features/dotnet:2": "latest" // or "" or {}
+    "ghcr.io/devcontainers/features/dotnet:2": "latest"
 }
 ```
 
 Installing an additional SDK version. Multiple versions can be specified as comma-separated values.
 
 ``` json
-{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
-        "additionalVersions": "lts"
+      "version": "latest",
+      "additionalVersions": "lts"
     }
 }
 ```
@@ -51,7 +50,6 @@ Installing an additional SDK version. Multiple versions can be specified as comm
 Installing specific SDK versions.
 
 ``` json
-{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "6.0",
@@ -63,7 +61,6 @@ Installing specific SDK versions.
 Installing a specific SDK feature band.
 
 ``` json
-{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "6.0.4xx",
@@ -74,7 +71,6 @@ Installing a specific SDK feature band.
 Installing a specific SDK patch version.
 
 ``` json
-{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "6.0.412",
@@ -85,7 +81,6 @@ Installing a specific SDK patch version.
 Installing only the .NET Runtime or the ASP.NET Core Runtime. (The SDK includes all runtimes so this configuration is only useful if you need to run .NET apps without building them from source.)
 
 ``` json
-{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "none",
@@ -95,12 +90,22 @@ Installing only the .NET Runtime or the ASP.NET Core Runtime. (The SDK includes 
 }
 ```
 
+Installing .NET workloads. Multiple workloads can be specified as comma-separated values.
+
+``` json
+"features": {
+    "ghcr.io/devcontainers/features/dotnet:2": {
+      "version": "latest",
+      "workloads": "aspire, wasm-tools"
+    }
+}
+```
+
 ## OS Support
 
 This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
 
 `bash` is required to execute the `install.sh` script.
-
 
 ---
 
--- a/src/dotnet/README.md
+++ b/src/dotnet/README.md
@@ -31,18 +31,19 @@ This Feature installs the latest .NET SDK, which includes the .NET CLI and the s
 Installing only the latest .NET SDK version (the default).
 
 ``` json
+{
 "features": {
-    "ghcr.io/devcontainers/features/dotnet:2": "latest"
+    "ghcr.io/devcontainers/features/dotnet:2": "latest" // or "" or {}
 }
 ```
 
 Installing an additional SDK version. Multiple versions can be specified as comma-separated values.
 
 ``` json
+{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
-      "version": "latest",
-      "additionalVersions": "lts"
+        "additionalVersions": "lts"
     }
 }
 ```
@@ -50,6 +51,7 @@ Installing an additional SDK version. Multiple versions can be specified as comm
 Installing specific SDK versions.
 
 ``` json
+{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "6.0",
@@ -61,6 +63,7 @@ Installing specific SDK versions.
 Installing a specific SDK feature band.
 
 ``` json
+{
 "features": {
     "ghcr.io/devcontainers/features/dotnet:2": {
         "version": "6.0.4xx",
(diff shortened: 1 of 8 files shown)


src/dotnet/README.md
```diff
@@ -95,13 +90,23 @@ Installing only the .NET Runtime or the ASP.NET Core Runtime. (The SDK includes
 }
 ```
 
+Installing .NET workloads. Multiple workloads can be specified as comma-separated values.
```
samruddhikhandale, line 93: Once this PR is merged, https://github.com/devcontainers/action will run automatically ; which opens a PR that updates the README file based on `devcontainer-feature.json` and `NOTES.md`.

Hence, can we revert the changes to this file to avoid merge conflicts? You can make these changes in `NOTES.md` instead. Thanks!

test/dotnet/scenarios.json
```diff
@@ -81,5 +81,15 @@
                 "aspnetcoreRuntimeVersions": "latest"
             }
         }
+    },
+    "install_dotnet_workloads": {
```
samruddhikhandale, line 85: Looks like the tests are failing, can you help fix them?

src/dotnet/README.md
```diff
@@ -31,27 +31,25 @@ This Feature installs the latest .NET SDK, which includes the .NET CLI and the s
 Installing only the latest .NET SDK version (the default).
 
 ``` json
-{
 "features": {
-    "ghcr.io/devcontainers/features/dotnet:2": "latest" // or "" or {}
+    "ghcr.io/devcontainers/features/dotnet:2": "latest"
```
samruddhikhandale, line 35: Can we keep other changes to README file as is? I think the current changes makes sense.

src/dotnet/devcontainer-feature.json
```diff
@@ -32,6 +32,11 @@
             "type": "string",
             "default": "",
             "description": "Enter additional ASP.NET Core runtime versions, separated by commas. Use 'latest' for the latest version, 'lts' for the latest LTS version, 'X.Y' or 'X.Y.Z' for a specific version."
+        },
+        "workloads": {
```
samruddhikhandale, line 36: 🎉

src/dotnet/install.sh
```diff
@@ -111,6 +117,10 @@ for version in "${aspNetCoreRuntimeVersions[@]}"; do
     install_runtime "aspnetcore" "$version"
 done
 
+for workload in "${workloads[@]}"; do
```
samruddhikhandale, line 120: Can we optimize the changes in this file? (untested)

```
IFS=',' read -ra workloads <<< "$WORKLOADS"
for workload in "${workloads[@]}"; do
    if [[ -n "$workload" ]]; then
        install_workload "$workload"
    fi
done
```

src/dotnet/scripts/dotnet-helpers.sh
```diff
@@ -76,12 +75,11 @@ install_sdk() {
     fi
     
     # Currently this script does not make it possible to qualify the version, 'GA' is always implied
-    echo "Executing $DOTNET_INSTALL_SCRIPT --version $version --channel $channel --install-dir $DOTNET_INSTALL_DIR --no-path"
```
samruddhikhandale, line 79: Curious, why was `--no-path` removed?

src/dotnet/scripts/dotnet-helpers.sh
```diff
@@ -117,3 +115,22 @@ install_runtime() {
...
+install_workload() {
+    local workload_id="$1"
+
+    # Prevent "Welcome to .NET" message from dotnet
+    local DOTNET_NOLOGO=true
+
+    # Prevent generating a development certificate
+    local DOTNET_GENERATE_ASPNET_CERTIFICATE=false
+
+    echo "Executing dotnet workload install $workload_id --temp-dir /tmp/dotnet-workload-temp-dir"
```
samruddhikhandale, line 131:
```suggestion
    echo "Installing .NET workload $workload_id"
```

src/dotnet/scripts/dotnet-helpers.sh
```diff
@@ -117,3 +115,22 @@ install_runtime() {
...
+    dotnet workload install "$workload_id" --temp-dir /tmp/devcontainer-nuget-cache
```
samruddhikhandale, line 132: Suggestion to improve the code: (untested)

```
if ! dotnet workload list | grep -q "$workload_id"; then
    echo "Installing .NET workload: $workload_id"
    dotnet workload install "$workload_id" --temp-dir /tmp/devcontainer-nuget-cache
else
    echo "Workload $workload_id is already installed."
fi
```

samruddhikhandale, line 132: Should this be `/tmp/dotnet-workload-temp-dir` to fix the tests?

samruddhikhandale (changes requested):
This would be a pretty nice improvement to the Feature, thank you so much.

We appreciate your constant efforts in improving the .NET Feature ✨

src/dotnet/README.md
sliekens: Sure, I mainly wanted to fix this code highlighting issue

image

Fixed it by changing the language to `jsonc` and also removed the unnecessary brace at the top.

src/dotnet/scripts/dotnet-helpers.sh
sliekens: The default behavior is to add `dotnet` to the PATH for the current shell session. That didn't seem very useful to me, so I initially added `--no-path` to disable modifying the PATH.

Now, I do need to run `dotnet` in the same shell session (to install workloads), so I re-enabled modifying PATH.

src/dotnet/scripts/dotnet-helpers.sh
sliekens: JimcareyDuhGIF

sliekens: Fixed it, thanks for the hint

test/dotnet/scenarios.json
sliekens: Thanks for the hint

src/dotnet/README.md
sliekens: Good point, forgot about that, I moved the changes to NOTES.

src/dotnet/scripts/dotnet-helpers.sh
sliekens: As I was testing this, I just realized you can pass multiple workloads in one line.

```
dotnet workload install <workload_id_1> <workload_id_2> <workload_id_n>
```

The install command already checks for duplicates, so I think that is the more preferred approach.

src/dotnet/install.sh
sliekens: I changed the code to run `dotnet workload install <workload_id1> [<workload_id2> ...]` in a single invocation

src/dotnet/scripts/dotnet-helpers.sh
sliekens: This is now done

sliekens:
@samruddhikhandale thanks for the review, I made some improvements and fixed the test, can you review again?

src/dotnet/install.sh
samruddhikhandale: Superb!

samruddhikhandale:
@gauravsaini04 Can you prioritize and help fix https://github.com/devcontainers/features/actions/runs/9490744002/job/26193280306?pr=997#step:4:2676 tests? Thanks!

src/dotnet/scripts/dotnet-helpers.sh
samruddhikhandale: 🎉

samruddhikhandale (approved):
Wonderful 🎉

Issue #1000, opened 2024-06-13
[Dotnet] Fix failing test "install_dotnet_lts"


samruddhikhandale:
@gauravsaini04 Can you prioritize and help fix https://github.com/devcontainers/features/actions/runs/9490744002/job/26193280306?pr=997#step:4:2676 tests? Thanks!

_Originally posted by @samruddhikhandale in https://github.com/devcontainers/features/issues/997#issuecomment-2166976456_

prathameshzarkar9:
Hi @samruddhikhandale ,

from resources it is found that , as of now, there is no .NET 8.0.302 designated as an LTS version. The latest LTS version is .NET 6.0. If .NET 8.0 were to become LTS in the future, it would be reflected in the URL: `https://dotnetcli.azureedge.net/dotnet/Sdk/**LTS**/latest.version`
whereas , $channel = "LTS"
output : **`8.0.301`**

However, at present, there is no LTS version 8.0.302 available to retrieve the latest version of the dotnet.

I found one fix to pass the test case `install_dotnet_lts` as below.
when I changed the $channel = 8.0 , it is correctly retrieving the latest version of dotnet **`8.0.302`**
the url will be: `https://dotnetcli.azureedge.net/dotnet/Sdk/**8.0**/latest.version`

PR is created here: https://github.com/devcontainers/features/pull/1003

PR #1003, opened 2024-06-16
dotnet lts fix

prathameshzarkar9:
[Issue 1000](https://github.com/devcontainers/features/issues/1000)
**Feature Name:**
- Dotnet

**Description:**
This PR does the following -
- changed the channel for the test from LTS to 8.0 to retrieve the latest sdk version 8.0.302

_Changelog:_
- Updated install_dotnet_lts.sh
    - changed channel to 8.0 

**Checklist:**

- [x] It should pass the test as the test install latest version 8.0.302

### DIFF
--- a/test/dotnet/install_dotnet_lts.sh
+++ b/test/dotnet/install_dotnet_lts.sh
@@ -13,7 +13,7 @@ source dev-container-features-test-lib
 source dotnet_env.sh
 source dotnet_helpers.sh
 
-expected=$(fetch_latest_version_in_channel "LTS")
+expected=$(fetch_latest_version_in_channel "8.0")
 
 check "Latest LTS version installed" \
 is_dotnet_sdk_version_installed "$expected"


test/dotnet/install_dotnet_lts.sh
```diff
@@ -13,7 +13,7 @@ source dev-container-features-test-lib
 source dotnet_env.sh
 source dotnet_helpers.sh
 
-expected=$(fetch_latest_version_in_channel "LTS")
+expected=$(fetch_latest_version_in_channel "8.0")
```
samruddhikhandale, line 16: This approach doesn't make sense.

As discussed this offline,

URL used by tests is https://github.com/devcontainers/features/blob/main/test/dotnet/dotnet_helpers.sh#L16C9-L16C87

However, for LTS we use  https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/dotnet-helpers.sh#L57-L61 & https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/dotnet-helpers.sh#L79-L82

We don’t seem to directly utilize any URL for finding LTS by the Feature but depend on the vendor script (which we have no control over).
Check if URLs mentioned in https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/vendor/dotnet-install.sh#L1301-L1302 or https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/vendor/dotnet-install.sh#L1311-L1312 might be making any difference and update the tests if we could sync!

samruddhikhandale (changes requested):

test/dotnet/install_dotnet_lts.sh
prathameshzarkar9: both the urls mentioned here:
https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/vendor/dotnet-install.sh#L1301-L1302 or https://github.com/devcontainers/features/blob/main/src/dotnet/scripts/vendor/dotnet-install.sh#L1311-L1312 
are not working to retrieve the latest sdk version of dotnet i.e. 8.0.302

can you suggets what we should update in the test?

sliekens: Hi, I also had a look at this problem and it seems the install-script uses `https://aka.ms/dotnet/$channel`, e.g.

```
https://aka.ms/dotnet/LTS/dotnet-sdk-linux-x64.tar.gz
```

This starts the download of version SDK 8.0.302 while [latest.version](https://dotnetcli.azureedge.net/dotnet/Sdk/LTS/latest.version) from the dotnetcli CDN is stuck on 8.0.301.

I would like to report this issue, but I don't know which github repo is best for this. Maybe [dotnet ](https://github.com/dotnet/core)?

sliekens: Actually they don't really want anyone to use the `latest.version` file (see https://github.com/dotnet/core/issues/9280#issuecomment-2082867900).

The recommended API for retrieving versions is
<https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json>

You can parse it with `jq`, for example:

``` sh
curl -s https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json |
    jq --raw-output '[."releases-index"[] | select(."release-type"=="lts" and ."support-phase"=="active") ."latest-sdk"] | first'```

Comment (prathameshzarkar9) [reply test/dotnet/install_dotnet_lts.sh:16]: Thanks @@sliekens. I made changes accordingly and now the test is not failing.

File: test/dotnet/dotnet_helpers.sh [hunk 2]
```diff
@@ -12,6 +12,9 @@ fetch_latest_version_in_channel() {
         wget -qO- "https://dotnetcli.azureedge.net/dotnet/Runtime/$channel/latest.version"
     elif [ "$runtime" = "aspnetcore" ]; then
         wget -qO- "https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$channel/latest.version"
+    elif [ "$runtime" = "LTS" ]; then
```
sliekens, line 15: shouldn't this be `"$channel" = "LTS"`?

prathameshzarkar9: Done

src/dotnet/install.sh
```diff
@@ -6,7 +6,7 @@
 #
 # Docs: https://github.com/devcontainers/features/tree/main/src/dotnet
 # Maintainer: The Dev Container spec maintainers
-DOTNET_VERSION="${VERSION:-"latest"}"
+DOTNET_VERSION="${VERSION:-"lts"}"
```
sliekens, line 9: I'm not sure if this was just for testing but otherwise I wouldn't change the default version

src/dotnet/install.sh
```diff
@@ -105,7 +105,7 @@ done
 
 # Install .NET versions and dependencies
 # icu-devtools includes dependencies for .NET
-check_packages wget ca-certificates icu-devtools
+check_packages wget ca-certificates icu-devtools curl jq
```
sliekens, line 108: Perhaps you could drop `curl` and only use `wget` for consistency with existing code

test/dotnet/dotnet_helpers.sh
```diff
@@ -12,6 +12,9 @@ fetch_latest_version_in_channel() {
         wget -qO- "https://dotnetcli.azureedge.net/dotnet/Runtime/$channel/latest.version"
     elif [ "$runtime" = "aspnetcore" ]; then
         wget -qO- "https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/$channel/latest.version"
+    elif [ "$channel" = "LTS" ]; then
+        echo $(curl -s https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json |
```
sliekens, line 16: I am thinking it would be good to replace all usages of `latest.version` with `release-index.json` for consistency, and because using the `latest.version` file in scripts is not endorsed by the .NET team

test/dotnet/scenarios.json
```diff
@@ -87,7 +87,7 @@
         "remoteUser": "vscode",
         "features": {
             "dotnet": {
-                "version": "latest",
+                "version": "lts",
```
sliekens, line 90: Same remark, if not needed for testing, I would revert this.

sliekens (changes requested):

PR #1006, opened 2024-06-18
Fix a typo shared among many installation scripts

hoechenberger:
`retring` -> `retrying`

Should I bump the version numbers of all features I touched here?

### DIFF
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -82,7 +82,7 @@ receive_gpg_keys() {
         echo "(*) Downloading GPG key..."
         ( echo "${keys}" | xargs -n 1 gpg --recv-keys) 2>&1 && gpg_ok="true"
         if [ "${gpg_ok}" != "true" ]; then
-            echo "(*) Failed getting key, retring in 10s..."
+            echo "(*) Failed getting key, retrying in 10s..."
             (( retry_count++ ))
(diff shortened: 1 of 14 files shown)


src/git-lfs/install.sh
```diff
@@ -82,7 +82,7 @@ receive_gpg_keys() {
         echo "(*) Downloading GPG key..."
         ( echo "${keys}" | xargs -n 1 gpg --recv-keys) 2>&1 && gpg_ok="true"
         if [ "${gpg_ok}" != "true" ]; then
-            echo "(*) Failed getting key, retring in 10s..."
+            echo "(*) Failed getting key, retrying in 10s..."
```
samruddhikhandale, line 85: Can we also bump patch versions for all the updated Features? Thank you!

samruddhikhandale (changes requested):
Thanks for fixing the typo, left one comment!

src/git-lfs/install.sh
hoechenberger: Certainly, I've addressed this in [500562b](https://github.com/devcontainers/features/pull/1006/commits/500562b6394645cc4f7db3d9a3f027224da07ab5)

samruddhikhandale (approved):
Thank you ⚡

PR #1007, opened 2024-06-18
dotnet: make 'dotnet' available to 'sudo' commands

sliekens:
Hello,

This PR adds a symbolic link from `/usr/bin/dotnet` to `/usr/share/dotnet/dotnet`.

This is necessary to make `sudo dotnet` commands work properly, because `sudo` resets the `PATH` and dotnet will not be found without this symlink.

For example, when setting up HTTPS developer certificates, you need `sudo` to save the certificate to a location owned by root.

``` sh
sudo -E dotnet dev-certs https --export-path /usr/local/share/ca-certificates/https.crt --format pem
sudo update-ca-certificates
```

(Example from https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-8.0&tabs=visual-studio%2Clinux-ubuntu#ubuntu-trust-the-certificate-for-service-to-service-communication)

### DIFF
--- a/src/dotnet/devcontainer-feature.json
+++ b/src/dotnet/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "dotnet",
-    "version": "2.1.0",
+    "version": "2.1.1",
     "name": "Dotnet CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/dotnet",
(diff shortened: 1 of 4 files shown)


samruddhikhandale (approved):
Looks good, thanks!

PR #1009, opened 2024-06-19
Allow installing `yarn` via `corepack` on Debian-based systems, too

### LINKED ISSUE #1004: [Node] Allow installing `yarn` on Debian-based systems via `corepack`, or disable global yarn installation altogether
The yarn docs state:

> The preferred way to manage Yarn is by-project and through [Corepack](https://yarnpkg.com/corepack), a tool shipped by default with Node.js. Modern releases of Yarn aren't meant to be installed globally, or from npm.

https://yarnpkg.com/getting-started/install

Currently, there is special-handling for Debian systems in the Node feature to install yarn from a .deb, stating it's for backward-compatibility. Is this really still needed? 

In any case, I'd like to install `yarn` via `corepack` in my Debian dev container to make use of the latest yarn release -- or be able to disable the global yarn installation altogether, as suggested in the yarn docs.

I'm thinking about adding a config setting that would allow me to do this. Would such a PR be accepted or are there objections?

hoechenberger:
The default behavior remains unchanged, but Debian and Ubuntu users can now opt-out from the APT-based installation to get the latest `yarn` via `corepack`.

Closes #1004

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -40,7 +40,7 @@
         "yarnFromApt": {
             "type": "boolean",
             "default": true,
-            "description": "Whether to install yarn globally via APT on Debian-based systems (Debian, Ubuntu). If false, will set up yarn via corepack. This setting is ignored on other Linux images, where yarn is always installed via corepack."
+            "description": "On Debian & Ubuntu systems, you can choose whether to install Yarn globally via APT. If this option is set to false, Yarn will be set up via Corepack instead. This setting does not apply to other Linux distributions, where Yarn is always installed using Corepack."
         }
     },
     "customizations": {
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.4.1",
+    "version": "1.5.0",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
@@ -36,6 +36,11 @@
             ],
             "default": "latest",
             "description": "Version of NVM to install."
+        },
+        "yarnFromApt": {
+            "type": "boolean",
+            "default": true,
+            "description": "Whether to install yarn globally via APT on Debian-based systems (Debian, Ubuntu). If false, will set up yarn via corepack. This setting is ignored on other Linux images, where yarn is always installed via corepack."
         }
     },
     "customizations": {
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -40,7 +40,7 @@
         "installYarnUsingApt": {
             "type": "boolean",
             "default": true,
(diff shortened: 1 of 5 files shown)


src/node/devcontainer-feature.json
```diff
@@ -36,6 +36,11 @@
             ],
             "default": "latest",
             "description": "Version of NVM to install."
+        },
+        "yarnFromApt": {
+            "type": "boolean",
+            "default": true,
+            "description": "Whether to install yarn globally via APT on Debian-based systems (Debian, Ubuntu). If false, will set up yarn via corepack. This setting is ignored on other Linux images, where yarn is always installed via corepack."
```
samruddhikhandale, line 43:
```suggestion
            "description": "On Debian & Ubuntu systems, you can choose whether to install Yarn globally via APT. If this option is set to false, Yarn will be set up via Corepack instead. This setting does not apply to other Linux distributions, where Yarn is always installed using Corepack."
```

src/node/devcontainer-feature.json
```diff
@@ -36,6 +36,11 @@
             ],
             "default": "latest",
             "description": "Version of NVM to install."
+        },
+        "yarnFromApt": {
```
samruddhikhandale, line 40:
```suggestion
        "installYarnUsingApt": {
```

src/node/install.sh
```diff
@@ -11,6 +11,7 @@ export NODE_VERSION="${VERSION:-"lts"}"
 export NVM_VERSION="${NVMVERSION:-"latest"}"
 export NVM_DIR="${NVMINSTALLPATH:-"/usr/local/share/nvm"}"
 INSTALL_TOOLS_FOR_NODE_GYP="${NODEGYPDEPENDENCIES:-true}"
+export YARN_FROM_APT="${YARNFROMAPT:-true}"  # only concerns Debian-based systems
```
samruddhikhandale, line 14: Need to update post rename the Feature option.

src/node/install.sh
```diff
@@ -188,7 +189,7 @@ find_version_from_git_tags() {
 }
 
 install_yarn() {
-    if [ "${ADJUSTED_ID}" = "debian" ]; then
+    if [ "${ADJUSTED_ID}" = "debian" ] && [ "${YARN_FROM_APT}" = "true" ]; then
```
samruddhikhandale, line 192: Can we add a test to validate these changes? Thanks!

samruddhikhandale (changes requested):
This makes sense, thanks! 

Left couple of comments

hoechenberger:
@samruddhikhandale I believe I've addressed all your comments. Unfortunately I currently cannot run the test case locally, so we'll have to rely on the CI output

test/node/debian_yarn_from_corepack.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
+# Definition specific tests
+YARN_VERSION="4.3.0"
+
+# Corepack provides shims for package managers like yarn. When yarn is invoked by the "yarn"
```
hoechenberger, line 11:
```suggestion
# Corepack provides shims for package managers like yarn. When yarn is invoked via the "yarn"
```

samruddhikhandale (approved):
Looks good, thanks!

samruddhikhandale (approved):

Issue #686, opened 2023-09-09
Terraform devcontainer feature v1.3.4 gpg error


f0o:
```
0.410 ===========================================================================
0.410 Feature       : Terraform, tflint, and TFGrunt
0.410 Description   : Installs the Terraform CLI and optionally TFLint and Terragrunt. Auto-detects latest version and installs
[2023-09-09T17:26:31.353Z]  needed dependencies.
0.410 Id            : ghcr.io/devcontainers/features/terraform
0.410 Version       : 1.3.4
0.410 Documentation : https://github.com/devcontainers/features/tree/main/src/terraform
0.410 Options       :
0.410     VERSION="latest"
0.410     TFLINT="latest"
0.410     TERRAGRUNT="latest"
0.410     INSTALLSENTINEL="false"
0.410     INSTALLTFSEC="false"
0.410     INSTALLTERRAFORMDOCS="false"
0.410     HTTPPROXY=""
0.410 ===========================================================================
0.863 TERRAFORM_VERSION=1.5.7
1.285 TFLINT_VERSION=0.48.0
1.760 TERRAGRUNT_VERSION=0.50.14
1.762 Downloading terraform...
2.451 (*) Downloading GPG key...
2.549 gpg: keyserver receive failed: Server indicated a failure
2.551 (*) Failed getting key, retring in 10s...
12.55 (*) Downloading GPG key...
12.62 gpg: keyserver receive failed: Server indicated a failure
12.62 (*) Failed getting key, retring in 10s...
22.62 (*) Downloading GPG key...
22.69 gpg: keyserver receive failed: Server indicated a failure
22.69 (*) Failed getting key, retring in 10s...
32.70 (*) Downloading GPG key...
32.78 gpg: keyserver receive failed: Server indicated a failure
32.78 (*) Failed getting key, retring in 10s...
42.78 (*) Downloading GPG key...
42.85 gpg: keyserver receive failed: Server indicated a failure
42.85 (*) Failed getting key, retring in 10s...
52.86 (!) Failed to get gpg key.
52.86 ERROR: Feature "Terraform, tflint, and TFGrunt" (ghcr.io/devcontainers/features/terraform) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/terraform for help troubleshooting this error.
------
```

how to go from here?

samruddhikhandale:
Hi 👋 

[...]

Issue #337, opened 2022-12-15
Installing git feature is slow


mataslib:
Is it normal that git feature takes about ~10 minutes to install? 

![image](https://user-images.githubusercontent.com/5154014/207907291-30752f6e-d5b4-4387-be3a-95e15b7cf4a2.png)

---
Win11
Wsl2
Docker for desktop
Dell XPS 9570, Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz, 2208 Mhz, 6 Core(s), 12 Logical Processor(s), 64 GB ram

samruddhikhandale:
Are you using the default options for the `git` Feature? If not, then can you paste the values in here?

With default options, it takes 12.7 sec - [action run](https://github.com/devcontainers/features/actions/runs/3709027564/jobs/6287205570#step:4:412)

With `ppa:false`, git is built from source and it takes ~4.5 mins - [action run](https://github.com/devcontainers/features/actions/runs/3680936945/jobs/6227071713#step:4:1012)

mataslib:
Screenshot is taken having this configuration
```json
"ghcr.io/devcontainers/features/git": {
      "version": "latest",
      "ppa": false
}
```
However, I had try also with `ppa: true`, I stopped the installation when there was ~800 seconds and counting - that takes even more time on my machine than `ppa: false`.

mataslib:
Alright `"version": "latest"` is culprit. With it deleted and `ppa: false` it is done in 0.5 seconds. Thank you for help! I don't know what is wrog with `"version": "latest"` however It seems to me that it can happen to lots of devs.

Chuxel:
@mataslib Linux distributions ship git inside them, but only one version. The PPA option on Ubuntu allows the use of pre-built versions of git other than the one in the distro.  In other cases, it has to be compiled to be installed, which takes a while.

You can pre-build your image if you really want the latest version of git to avoid this problem. https://containers.dev/implementors/reference/#prebuilding

samruddhikhandale:
> With `ppa:true` It hangs on `=> => # (*) Downloading GPG key... `
> 
> ![image](https://user-images.githubusercontent.com/5154014/208102153-8567f0aa-5fee-4d63-864f-74d1516ff067.png)

ℹ️  This is a separate issue tracked in https://github.com/devcontainers/features/issues/323

samruddhikhandale:
> @samruddhikhandale Should we keep this open to track potentially pre-building git ahead of time?

[...]

PR #1016, opened 2024-06-25
Resolve slow key retrieval when some GPG key servers are not reachable

h3xcat:
At the moment, if the host environment cannot reach first key server over port 11371, it takes 1 minute to retrieve individual keys from the other listed key servers. For the Python feature, the installation script retrieves 13 keys, so the total time to retrieve all the keys is 13 minutes. The delay could be further extended if subsequent servers also cannot be reached.

The reason it takes 1 minute rather than the 15 second default timeout in `dirmngr` is that I believe `dirmngr` internally tries the connection to the same keyserver 4 times before it moves on to a different server. Essentially, this multiplies the timeout delay by 4.

Instead of relying on the `dirmngr` timeout mechanism, the change implements a check for keyserver reachability using `curl` before the list is passed on to `dirmngr`.

If no keyservers are identified, the script will intentionally fail. If some keyservers are not reachable, the script will log this information, enabling users to more easily identify and resolve network issues in their environments.

Related issues: 
https://github.com/devcontainers/features/issues/686
https://github.com/devcontainers/features/issues/337
https://github.com/devcontainers/features/issues/323

**Affected features:**
- git-lfs
- git
- github-cli
- kubectl-helm-minikube
- python
- ruby
- terraform

I ran tests for all the features locally; most succeeded without issues. The only test that failed was `autoPullDisabled` from `git-lfs`, but it also fails with unmodified code, so I suspect something else in my environment is causing the tests to fail, not specific to my changes.

### DIFF
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -62,11 +62,12 @@ find_version_from_git_tags() {
 
 # Get the list of GPG key servers that are reachable
 get_gpg_key_servers() {
-    declare -A keyservers_curl_map
-    keyservers_curl_map["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com:11371"
-    keyservers_curl_map["hkp://keyserver.ubuntu.com:80"]="http://keyserver.ubuntu.com"
-    keyservers_curl_map["hkps://keys.openpgp.org"]="https://keys.openpgp.org:443"
-    keyservers_curl_map["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
+    declare -A keyservers_curl_map=(
+        ["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com:11371"
+        ["hkp://keyserver.ubuntu.com:80"]="http://keyserver.ubuntu.com"
+        ["hkps://keys.openpgp.org"]="https://keys.openpgp.org"
+        ["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
+    )
 
     local curl_args=""
     local keyserver_reachable=false  # Flag to indicate if any keyserver is reachable
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -69,34 +69,36 @@ get_gpg_key_servers() {
     keyservers_curl_map["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
 
     local curl_args=""
+    local keyserver_reachable=false  # Flag to indicate if any keyserver is reachable
+
     if [ ! -z "${KEYSERVER_PROXY}" ]; then
         curl_args="--proxy ${KEYSERVER_PROXY}"
     fi
 
-    local keyserver_list=""
     for keyserver in ${!keyservers_curl_map[@]}; do
         local keyserver_curl_url="${keyservers_curl_map[${keyserver}]}"
-        if curl -s ${curl_args} --max-time 3 ${keyserver_curl_url} > /dev/null; then
-            keyserver_list="${keyserver_list}keyserver ${keyserver}"
-            keyserver_list+=$'\n'
+        if curl -s ${curl_args} --max-time 5 ${keyserver_curl_url} > /dev/null; then
+            echo "keyserver ${keyserver}"
+            keyserver_reachable=true
         else
             echo "(*) Keyserver ${keyserver} is not reachable." >&2
         fi
     done
 
-    if [ -z "${keyserver_list}" ]; then
+    if ! $keyserver_reachable; then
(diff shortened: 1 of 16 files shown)


h3xcat:
Alternatively I'm wondering if a quick check could be performed with a curl command to use or generate alternative keyserver list.

h3xcat:

src/git-lfs/install.sh
```diff
@@ -64,15 +60,51 @@ find_version_from_git_tags() {
...
+
+    if [ ! -z "${KEYSERVER_PROXY}" ]; then
+        curl_args="--proxy ${KEYSERVER_PROXY}"
+    fi
+
+    for keyserver in "${!keyservers_curl_map[@]}"; do
+        local keyserver_curl_url="${keyservers_curl_map[${keyserver}]}"
+        if curl -s ${curl_args} --max-time 5 ${keyserver_curl_url} > /dev/null; then
+            echo "keyserver ${keyserver}"
+            keyserver_reachable=true
```
samruddhikhandale, line 82: Should we return from the loop once the keyserver is reachable? 

Else, it might try to loop over all the keyservers

src/git-lfs/install.sh
```diff
@@ -64,15 +60,51 @@ find_version_from_git_tags() {
     echo "${variable_name}=${!variable_name}"
 }
 
+# Get the list of GPG key servers that are reachable
+get_gpg_key_servers() {
+    declare -A keyservers_curl_map
+    keyservers_curl_map["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com:11371"
+    keyservers_curl_map["hkp://keyserver.ubuntu.com:80"]="http://keyserver.ubuntu.com"
+    keyservers_curl_map["hkps://keys.openpgp.org"]="https://keys.openpgp.org:443"
+    keyservers_curl_map["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
```
samruddhikhandale, line 69:
```suggestion
    keyservers_curl_map["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com"
    keyservers_curl_map["hkp://keyserver.ubuntu.com:80"]="http://keyserver.ubuntu.com:80"
    keyservers_curl_map["hkps://keys.openpgp.org"]="https://keys.openpgp.org"
    keyservers_curl_map["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com"
```

We have seen issues where the default ports (eg. 11371) are blocked. Hence, can we have the http URLs resemble the values we currently have?

samruddhikhandale, line 69: For improving readability, we can also declare as follows :

```
declare -A keyservers_curl_map=(
        ["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com:11371"
        ["hkp://keyserver.ubuntu.com:80"]="http://keyserver.ubuntu.com"
        ["hkps://keys.openpgp.org"]="https://keys.openpgp.org:443"
        ["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
    )
```

src/git-lfs/install.sh
```diff
@@ -15,10 +15,6 @@ GIT_LFS_ARCHIVE_GPG_KEY_URI="https://packagecloud.io/github/git-lfs/gpgkey"
 GIT_LFS_ARCHIVE_ARCHITECTURES="amd64 arm64"
 GIT_LFS_ARCHIVE_VERSION_CODENAMES="stretch buster bullseye bionic focal jammy"
```
samruddhikhandale, line 16: Can we bump patch versions for the updated Features? thanks! 

https://github.com/devcontainers/features/blob/a10c59fbee9fff661d3310f58e95dba6f0056bdb/src/git-lfs/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
thanks for contributing, definitely a helpful improvement.

src/git-lfs/install.sh
h3xcat: Looping through other servers is intentional. Key servers do not guarantee they all share the same identical keys. So it's useful to include other reachable keyservers as fallback for some specific keys.

src/git-lfs/install.sh
h3xcat: Default port for `hkp://` is 11371. Hence, I'm using this map to store `http://` equivalent destinations to be used with `curl`, as `curl` does not accept `hkp(s)://` addresses as destinations.

samruddhikhandale (approved):
Looks good, thanks! Let's merge this on Monday to ensure we accidentally don't break anything.

h3xcat:
@samruddhikhandale 
Could you merge the pull request when you are able to? If there are any concerns, please let me know.

Issue #1011, opened 2024-06-20
`java` installed `21.x-tem` instead of `-ms` by default


justinyoo:
Hi, Team.

I've found a few issues while using the `java` feature:

1. **Document discrepancy**:  According to this doc: https://github.com/devcontainers/features/blob/main/src/java/README.md, the default value of the `version` is `latest`. But `install.sh` uses `lts` as a default value. As you know `latest` and `lts` may or may not be the same as each other.
2. **Default OpenJDK version**: In the `sdk_install()` function from `install.sh`, if the requested version is `lts` (default value), it seems to be defaulted to `21.0.3-tem` regardless the distro is set to `ms` (default). I was expecting to install the latest version of Microsoft distro, which is `21.0.3-ms` for this case. I wonder if it's by design or not?

    https://github.com/devcontainers/features/blob/7953318c38a98d6bba8c8e78f7fb5bce095d7619/src/java/install.sh#L210-L211

samruddhikhandale:
Hi 👋 

Thanks for reporting the bugs, @gauravsaini04 can you help fix these? thanks!

gauravsaini04:
Have raised as a probable solution in pr [#1021](https://github.com/devcontainers/features/pull/1021)

gauravsaini04:
pr was merged!

PR #1021, opened 2024-06-28
[Java] - features issue #1101 - installs -tem instead -ms for lts installation

gauravsaini04:
**Feature name**:
 
 * Java
 
 **Description**:
 
 This PR is for solution to following issue:
 
 * Features issue [#1011](https://github.com/devcontainers/features/issues/1011) 
     - jdk lts resulted in installation of 21.x-tem instead of 21.x.x-ms, which is not the intended behaviour. 
 
 _Changelog_:
 
 * Updated `install.sh`
 
 **Checklist**:
 
### DIFF
--- a/src/java/devcontainer-feature.json
+++ b/src/java/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.5.0",
+  "version": "1.5.1",
   "name": "Java (via SDKMAN!)",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/java",
   "description": "Installs Java, SDKMAN! (if not installed), and needed dependencies.",
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -214,7 +214,7 @@ sdk_install() {
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
     if [ "${requested_version}" = "none" ]; then return; fi
-    if [ "${requested_version}" = "default" ] || [ ( "${install_type}" = "maven" -o "${install_type}" = "groovy" ) -a ( "${requested_version}" = "latest" ) ]; then
+    if [ "${requested_version}" = "default" ] || { { [ "${install_type}" = "maven" ] || [ "${install_type}" = "groovy" ] } && [ "${requested_version}" = "latest" ]; }; then
         requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -229,7 +229,7 @@ sdk_install() {
     if [ "${set_as_default}" = "true" ]; then
         JAVA_VERSION=${requested_version}
     fi
-    
+
     su ${USERNAME} -c "umask 0002 && . ${SDKMAN_DIR}/bin/sdkman-init.sh && sdk install ${install_type} ${requested_version} && sdk flush archives && sdk flush temp"
 }
 
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -213,10 +213,12 @@ sdk_install() {
     local suffix="${4:-"\\s*"}"
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
+    pkgs=("maven" "gradle" "ant" "groovy")
+    pkg_vals="${pkgs[@]}"
     if [ "${requested_version}" = "none" ]; then return; fi
     if [ "${requested_version}" = "default" ]; then
         requested_version=""
-    elif ( [ "${install_type}" = "maven" ] || [ "${install_type}" = "groovy" ] ) && [ "${requested_version}" = "latest" ]; then
+    elif [[ "${pkg_vals}" =~ "${install_type}" ]] && [ "${requested_version}" = "latest" ]; then
         requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -214,7 +214,7 @@ sdk_install() {
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
     if [ "${requested_version}" = "none" ]; then return; fi
-    if [ "${requested_version}" = "default" ] || { { [ "${install_type}" = "maven" ] || [ "${install_type}" = "groovy" ] } && [ "${requested_version}" = "latest" ]; }; then
+    if [ "${requested_version}" = "default" ] || [ { [ "${install_type}" = "maven" ] || [ "${install_type}" = "groovy" ] } && [ "${requested_version}" = "latest" ] ]; then
         requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -205,6 +205,22 @@ get_jdk_distro() {
     fi
 }
 
+find_version_list() {
+    prefix="$1"
+    suffix="$2"
+    install_type=$3
+    ifLts="$4"
+    if [ "${ifLts}" = "true" ]; then 
+        all_lts_versions=$(curl -s https://api.adoptium.net/v3/info/available_releases)
+        major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_lts')
+        regex="${prefix}\\K${major_version}\\.?[0-9]*\\.?[0-9]*${suffix}"
+    else 
+        regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
+    fi
+    version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
+    echo $version_list
+}
+
 # Use SDKMAN to install something using a partial version match
 sdk_install() {
     local install_type=$1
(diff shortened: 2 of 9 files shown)


src/java/install.sh
```diff
@@ -206,15 +206,15 @@ sdk_install() {
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
     if [ "${requested_version}" = "none" ]; then return; fi
-    # Blank will install latest stable version SDKMAN has
-    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "lts" ] || [ "${requested_version}" = "default" ]; then
+     # Blank will install latest stable version SDKMAN has
+    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "default" ]; then
```
justinyoo, line 210: Still, are we assuming `latest` to use the `tem` distro? Shouldn't it be the latest version of the given distro?

src/java/install.sh
```diff
@@ -206,15 +206,15 @@ sdk_install() {
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
```
samruddhikhandale, line 207: Can we bump patch version?

src/java/install.sh
samruddhikhandale: @justinyoo  Why do you think it's always using `tem` ? 

It's using `tem` only if `jdkDistro` is `ms` and version is 8 or 18. @gauravsaini04 Can we check is MS started supporting those two versions as well?

src/java/install.sh
```diff
@@ -206,15 +206,15 @@ sdk_install() {
...
+    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "default" ]; then
          requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
         local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
-        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ]; then
+        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ] || [ "${requested_version}" = "lts" ]; then
             requested_version="$(echo "${version_list}" | head -n 1)"
```
samruddhikhandale, line 218: I am not sure if this version would give us `lts`. Can we double check>

samruddhikhandale (changes requested):

src/java/install.sh
gauravsaini04: no `ms` haven't yet started supporting these 2 versions yet i.e. ones starting with 8 or 18..

src/java/install.sh
gauravsaini04: actually i had tested scenarios where version was given as `lts` and fed some additionalVersions: '17, 8' like this, could see installation taking ms for default version and then for 17 also ms and for 8 it would pick up tem..

src/java/install.sh
gauravsaini04: done

src/java/install.sh
justinyoo, line 210: Based on the README,

https://github.com/devcontainers/features/blob/9387225ac7ada315d94511848fd490a8519ed591/src/java/README.md?plain=1#L18

This `install.sh` should receive `latest`, if nothing is passed. However, it automatically sets 'lts' as default, which is different from README.

That was mentioned in #1011

src/java/install.sh
```diff
@@ -206,15 +206,15 @@ sdk_install() {
     local full_version_check=${5:-".*-[a-z]+"}
     local set_as_default=${6:-"true"}
     if [ "${requested_version}" = "none" ]; then return; fi
-    # Blank will install latest stable version SDKMAN has
-    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "lts" ] || [ "${requested_version}" = "default" ]; then
+     # Blank will install latest stable version SDKMAN has
+    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "default" ]; then
          requested_version=""
```
justinyoo, line 211: And, if this `requested_version` value is `""` (empty or blank), it automatically picks up `21.0.3-tem`, regardless the `jdkDistro` is set to `ms`.

src/java/install.sh
samruddhikhandale: AH, thank's for pointing this out. It's definitely a bug, @gauravsaini04 can we update to match with the default `latest`? 

[features/src/java/install.sh](https://github.com/devcontainers/features/blob/9387225ac7ada315d94511848fd490a8519ed591/src/java/install.sh#L12)

src/java/install.sh
samruddhikhandale: @gauravsaini04 Can we fix this bug as well? The Feature should consider the `jdkDistro` while computing `latest` or `lts`

src/java/install.sh
gauravsaini04: corrected

src/java/install.sh
gauravsaini04: corrected

src/java/install.sh
```diff
@@ -206,15 +206,15 @@ sdk_install() {
...
+     # Blank will install latest stable version SDKMAN has
+    if [ "${requested_version}" = "default" ]; then
          requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
         local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
-        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ]; then
+        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ] || [ "${requested_version}" = "lts" ] || [ "${requested_version}" = "latest" ]; then
```
samruddhikhandale, line 217: `[ "${requested_version}" = "latest"` is duplicate

samruddhikhandale, line 217: How is this condition block finding `lts` version?

samruddhikhandale (changes requested):

src/java/install.sh
gauravsaini04: earlier, whenever it would find `lts` or `latest` as the required version, then it would simply skip the finding of version from this condition block and would go on installing `tem` instead of `ms`. So, checking for `lts` or `latest` and then entering this code block, makes consider anything coming out of `get_jdk_distros` function, where the execution checks in earlier than the `sdk_install` function

src/java/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.5.0",
+  "version": "1.5.1",
```
samruddhikhandale, line 3:
```suggestion
  "version": "1.6.0",
```

src/java/install.sh
samruddhikhandale: > requested_version="$(echo "${version_list}" | head -n 1)"

I am talking about ^

LTS is not necessarily the latest or current version. Hence, the way this PR is now finding the LTS version is not correct. 

Do we have a way to determine the lts version for the given distro?

samruddhikhandale (changes requested):

src/java/install.sh
gauravsaini04: corrected logic for lts finding

src/java/install.sh
```diff
@@ -206,15 +206,21 @@ sdk_install() {
...
          requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
         local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
-        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ] || [ "${requested_version}" = "lts" ]; then
+        if [ "${requested_version}" = "lts" ]; then
+            check_packages jq
+            all_lts_versions=$(curl -s https://api.adoptopenjdk.net/v3/info/available_releases)
```
samruddhikhandale, line 218: Won't this only work for the "Temurin" distribution?

I am concerned that a user might encounter failures if they select a different distribution that hasn't yet released the LTS version returned by this API.

Is there an easier way to find the LTS version for each JDK distribution?

samruddhikhandale, line 218: I like the approach of how you are finding the LTS version 🎉 

image


Quick question, who hosts this API? Is it https://adoptium.net/ or someone else?
We need to ensure it is safe to use before we use this approach

src/java/install.sh
```diff
@@ -206,15 +206,21 @@ sdk_install() {
...
+            major_version=$(echo "$all_lts_versions" | jq -r '.available_lts_releases | sort | .[-1]')
```
samruddhikhandale, line 219: Why don't we directly use `most_recent_lts` property?

src/java/install.sh
gauravsaini04: [The API at api.adoptopenjdk.net is hosted by AdoptOpenJDK, which has now moved to the Eclipse Foundation](https://api.adoptopenjdk.net/)[1](https://api.adoptopenjdk.net/). [As part of this transition, AdoptOpenJDK has rebranded to Adoptium, and it’s recommended to use the new API endpoint at api.adoptium.net](https://api.adoptopenjdk.net/)

src/java/install.sh
gauravsaini04: yeah right

src/java/install.sh
samruddhikhandale: Great, thanks!

src/java/install.sh
```diff
@@ -205,16 +213,26 @@ sdk_install() {
...
-    if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "lts" ] || [ "${requested_version}" = "default" ]; then
-         requested_version=""
+    if [ "${requested_version}" = "default" ]; then
+        requested_version=""
+    elif [[ "${pkg_vals}" =~ "${install_type}" ]] && [ "${requested_version}" = "latest" ]; then
+        requested_version=""
     elif echo "${requested_version}" | grep -oE "${full_version_check}" > /dev/null 2>&1; then
         echo "${requested_version}"
     else
         local regex="${prefix}\\K[0-9]+\\.?[0-9]*\\.?[0-9]*${suffix}"
```
samruddhikhandale, line 226: Should we move definition of `regex` to the `else` block?

src/java/install.sh
```diff
@@ -205,16 +213,26 @@ sdk_install() {
...
         local version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
```
samruddhikhandale, line 227: Should we move this to `else` block?

src/java/install.sh
```diff
@@ -205,16 +213,26 @@ sdk_install() {
...
-        if [ "${requested_version}" = "latest" ] || [ "${requested_version}" = "current" ]; then
+        if [ "${requested_version}" = "lts" ]; then
+            check_packages jq
+            all_lts_versions=$(curl -s https://api.adoptium.net/v3/info/available_releases)
+            major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_lts')
+            regex="${prefix}\\K${major_version}\\.?[0-9]*\\.?[0-9]*${suffix}"
+            version_list=$(su ${USERNAME} -c ". \${SDKMAN_DIR}/bin/sdkman-init.sh && sdk list ${install_type} 2>&1 | grep -oP \"${regex}\" | tr -d ' ' | sort -rV")
```
samruddhikhandale, line 233: looks like this is repetitive? Can we move to a function?

samruddhikhandale (dismissed):

src/java/install.sh
gauravsaini04: sure, done

src/java/install.sh
gauravsaini04: ok

samruddhikhandale (approved):
⚡

Issue #945, opened 2024-04-15
[desktop-lite] Entrypoint script does not use VNC_RESOLUTION defined by containerEnv


Huoleit:
## Summary

As written in the [`desktop-init.sh`](https://github.com/devcontainers/features/blob/e7dd9fafd9aeede11d0a59a0ace819e3b774de2a/src/desktop-lite/install.sh#L293) , the entrypoint script will use user-defined `VNC_RESOLUTION` if provided. But even if I include the env definition in both `containerEnv` and `remoteEnv`, the entrypoint script always uses the default setting. 

## Screenshot
![Screenshot from 2024-04-15 15-10-01](https://github.com/devcontainers/features/assets/29969110/ba334c26-4893-4bfa-a047-b876e27d3d98)

## Steps to reproduce

`.devcontainer` used.
```
{
	"name": "Ubuntu",
	"image": "mcr.microsoft.com/devcontainers/base:jammy",
	"features": {
		"ghcr.io/devcontainers/features/desktop-lite:1": {}
	},
	"containerEnv": {
		"VNC_RESOLUTION": "1920x1080x32"
	},
	"remoteEnv": {
		"VNC_RESOLUTION": "1920x1080x32"
	},
	"forwardPorts": [
		5901,
		6080
	],
	"containerUser": "vscode"
}
```
## Log
```
[2024-04-15T13:06:53.341Z] Start: Run: docker run --sig-proxy=false -a STDOUT -a STDERR --mount type=bind,source=/home/docker_test,target=/workspaces/docker_test --mount type=volume,src=vscode,dst=/vscode -e VNC_RESOLUTION=1920x1080x32 -u vscode --init --entrypoint /bin/sh vsc-docker_test-07c641cc479b026b86df51ee3bad1935608dd8988b8d00265001f60892599974-features-uid -c echo Container started
```

Huoleit:
After investigating, it appears that `docker run` doesn't pass environment variables to a non-interactive shell. One workaround is setting those variables in the Dockerfile using `ENV`. However, since `VNC_RESOLUTION` is user-specific and I prefer not to expose it in the Dockerfile, is there a better approach to configure it in `devcontainer.json`?

samruddhikhandale:
Ideally, using `containerEnv` to update `VNC_RESOLUTION` in `devcontainer.json` should work out of the box. It's strange that it's not working for you. 

[...]

PR #1024, opened 2024-06-28
[Desktop-lite] - VNC_RESOLUTION unreadable fix - issue #945 fix

gauravsaini04:
**Feature name**:
 
 * Desktop-lite
 
 **Description**:
 
 This PR introduces the following functionality :
 
 * This PR is as a solution to Features issue #945 
 * VNC_RESOLUTION's value was not able to be read when set inside `containerEnv` as a configured option, appearing to be available as an Environment variable during build time, but it can only be read at runtime of container.
  
 
 _Changelog_:
 
 * Updated `install.sh`
 * Added a single quotes wrapper around EOF so as to disable variable expansion during build time, and such that the variable's values can be read when they are needed i.e. during runtime of container.
 
 **Checklist**:
 
### DIFF
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -329,10 +329,26 @@ keepRunningInBackground()
     ($2 bash -c "while :; do echo [\\$(date)] Process started.; $3; echo [\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
 }
 
+apt_get_update()
+{
+    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
+        echo "Running apt-get update..."
+        apt-get update -y
+    fi
+}
+
+check_packages() {
+    if ! dpkg -s "$@" > /dev/null 2>&1; then
+        apt_get_update
+        apt-get -y install --no-install-recommends "$@"
+    fi
+}
+
 # Use sudo to run as root when required
 sudoIf()
 {
     if [ "$(id -u)" -ne 0 ]; then
+        check_packages sudo
         sudo "$@"
     else
         "$@"
@@ -343,6 +359,7 @@ sudoIf()
 sudoUserIf()
 {
     if [ "$(id -u)" -eq 0 ] && [ "${user_name}" != "root" ]; then
+        check_packages sudo
         sudo -u "${user_name}" "$@"
     else
         "$@"
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -309,24 +309,24 @@ export LANGUAGE="${LANGUAGE:-"en_US.UTF-8"}"
 # Execute the command it not already running
 startInBackgroundIfNotRunning()
 {
-    log "Starting \$1."
-    echo -e "\n** \$(date) **" | sudoIf tee -a /tmp/\$1.log > /dev/null
-    if ! pgrep -x \$1 > /dev/null; then
-        keepRunningInBackground "\$@"
-        while ! pgrep -x \$1 > /dev/null; do
+    log "Starting $1."
+    echo -e "\n** $(date) **" | sudoIf tee -a /tmp/\$1.log > /dev/null
+    if ! pgrep -x $1 > /dev/null; then
+        keepRunningInBackground "$@"
+        while ! pgrep -x $1 > /dev/null; do
             sleep 1
         done
-        log "\$1 started."
+        log "$1 started."
     else
-        echo "\$1 is already running." | sudoIf tee -a /tmp/\$1.log > /dev/null
-        log "\$1 is already running."
+        echo "$1 is already running." | sudoIf tee -a /tmp/\$1.log > /dev/null
+        log "$1 is already running."
     fi
 }
 
 # Keep command running in background
 keepRunningInBackground()
 {
-    ($2 bash -c "while :; do echo [\\\$(date)] Process started.; \$3; echo [\\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
+    ($2 bash -c "while :; do echo [\\$(date)] Process started.; $3; echo [\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
 }
 
 # Use sudo to run as root when required
@@ -352,7 +352,7 @@ sudoUserIf()
 # Log messages
 log()
 {
-    echo -e "[\$(date)] \$@" | sudoIf tee -a \$LOG > /dev/null
+    echo -e "[$(date)] $@" | sudoIf tee -a $LOG > /dev/null
 }
 
 log "** SCRIPT START **"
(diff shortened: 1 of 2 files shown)


src/desktop-lite/install.sh
```diff
@@ -293,7 +293,7 @@ echo -e "\nSuccess!\n"
 EOF
 
 # Container ENTRYPOINT script
-cat << EOF > /usr/local/share/desktop-init.sh
+cat << 'EOF' > /usr/local/share/desktop-init.sh
```
samruddhikhandale, line 296: Can we update patch version?

src/desktop-lite/install.sh
```diff
@@ -293,7 +293,7 @@ echo -e "\nSuccess!\n"
 EOF
 
 # Container ENTRYPOINT script
```
samruddhikhandale, line 295: Did you test if it works locally?

samruddhikhandale (changes requested):

src/desktop-lite/install.sh
gauravsaini04: done

src/desktop-lite/install.sh
gauravsaini04: yes

samruddhikhandale (approved):
Thanks! ⚡

samruddhikhandale:
@gauravsaini04 Can you look into test failures?

gauravsaini04:
> @gauravsaini04 Can you look into test failures?

Hi @samruddhikhandale ,
have tested locally with a couple of these scenarios and they seem to be passing the `log-exists` test there in the codespace

src/desktop-lite/install.sh
samruddhikhandale, line 295: How are the tests passing locally? 

From https://github.com/devcontainers/features/actions/runs/9719274381/job/27190519697?pr=1024#step:4:3497

It looks like these PR changes are failing the init script run. Can you investigate here? 
 
image

samruddhikhandale (changes requested):

src/desktop-lite/install.sh
```diff
@@ -309,50 +309,50 @@ export LANGUAGE="${LANGUAGE:-"en_US.UTF-8"}"
...
+        echo "$1 is already running." | sudoIf tee -a /tmp/\$1.log > /dev/null
+        log "$1 is already running."
     fi
 }
 
 # Keep command running in background
 keepRunningInBackground()
 {
-    (\$2 bash -c "while :; do echo [\\\$(date)] Process started.; \$3; echo [\\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
+    ($2 bash -c "while :; do echo [\\$(date)] Process started.; $3; echo [\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
```
samruddhikhandale, line 329: Good to see that the tests are passing, thanks!  🎉 

Can you now look into why we are seeing this? 👇 
https://github.com/devcontainers/features/actions/runs/10202059742/job/28225358778?pr=1024#step:4:3407
image

Also, can you paste the output of the /tmp/container-init.log file? Also, update the tests to do a cat on the file, would like the see the status there.

samruddhikhandale (changes requested):

src/desktop-lite/install.sh
gauravsaini04:
```
#!/bin/bash

user_name="${USERNAME}"
group_name="$(id -gn ${USERNAME})"
LOG=/tmp/container-init.log

export DBUS_SESSION_BUS_ADDRESS="${DBUS_SESSION_BUS_ADDRESS:-"autolaunch:"}"
export DISPLAY="${DISPLAY:-:1}"
export VNC_RESOLUTION="${VNC_RESOLUTION:-1440x768x16}" 
export LANG="${LANG:-"en_US.UTF-8"}"
export LANGUAGE="${LANGUAGE:-"en_US.UTF-8"}"

# Execute the command it not already running
startInBackgroundIfNotRunning()
{
    log "Starting $1."
    echo -e "\n** $(date) **" | sudoIf tee -a /tmp/\$1.log > /dev/null
    if ! pgrep -x $1 > /dev/null; then
        keepRunningInBackground "$@"
        while ! pgrep -x $1 > /dev/null; do
            sleep 1
        done
        log "$1 started."
    else
        echo "$1 is already running." | sudoIf tee -a /tmp/\$1.log > /dev/null
        log "$1 is already running."
    fi
}

# Keep command running in background
keepRunningInBackground()
{
    ($2 bash -c "while :; do echo [\\$(date)] Process started.; $3; echo [\\$(date)] Process exited!; sleep 5; done 2>&1" | sudoIf tee -a /tmp/\$1.log > /dev/null & echo "\$!" | sudoIf tee /tmp/\$1.pid > /dev/null)
}

# Use sudo to run as root when required
sudoIf()
{
    if [ "$(id -u)" -ne 0 ]; then
        sudo "$@"
    else
        "$@"
    fi
}

# Use sudo to run as non-root user if not already running
sudoUserIf()
{
    if [ "$(id -u)" -eq 0 ] && [ "${user_name}" != "root" ]; then
        sudo -u "${user_name}" "$@"
    else
        "$@"
    fi
}

# Log messages
log()
{
    echo -e "[$(date)] $@" | sudoIf tee -a $LOG > /dev/null
}

log "** SCRIPT START **"

# Start dbus.
log 'Running "/etc/init.d/dbus start".'
if [ -f "/var/run/dbus/pid" ] && ! pgrep -x dbus-daemon  > /dev/null; then
    sudoIf rm -f /var/run/dbus/pid
fi
sudoIf /etc/init.d/dbus start 2>&1 | sudoIf tee -a /tmp/dbus-daemon-system.log > /dev/null
while ! pgrep -x dbus-daemon > /dev/null; do
    sleep 1
done

# Startup tigervnc server and fluxbox
sudoIf rm -rf /tmp/.X11-unix /tmp/.X*-lock
mkdir -p /tmp/.X11-unix
sudoIf chmod 1777 /tmp/.X11-unix
sudoIf chown root:${group_name} /tmp/.X11-unix
if [ "$(echo "${VNC_RESOLUTION}" | tr -cd 'x' | wc -c)" = "1" ]; then VNC_RESOLUTION=${VNC_RESOLUTION}x16; fi
screen_geometry="${VNC_RESOLUTION%*x*}"
screen_depth="${VNC_RESOLUTION##*x}"

# Check if VNC_PASSWORD is set and use the appropriate command
common_options="tigervncserver \${DISPLAY} -geometry ${screen_geometry} -depth ${screen_depth} -rfbport ${VNC_PORT} -dpi ${VNC_DPI:-96} -localhost -desktop fluxbox -fg"

if [ -n "${VNC_PASSWORD+x}" ]; then
    startInBackgroundIfNotRunning "Xtigervnc" sudoUserIf "${common_options} -passwd /usr/local/etc/vscode-dev-containers/vnc-passwd"
else
    startInBackgroundIfNotRunning "Xtigervnc" sudoUserIf "${common_options} -SecurityTypes None"
fi

# Spin up noVNC if installed and not running.
if [ -d "/usr/local/novnc" ] && [ "$(ps -ef | grep /usr/local/novnc/noVNC*/utils/launch.sh | grep -v grep)" = "" ]; then
    keepRunningInBackground "noVNC" sudoIf "/usr/local/novnc/noVNC*/utils/launch.sh --listen ${NOVNC_PORT} --vnc localhost:${VNC_PORT}"
    log "noVNC started."
else
    log "noVNC is already running or not installed."
fi

# Run whatever was passed in
log "Executing \"\$@\"."
exec "$@"
log "** SCRIPT EXIT **"
```

gauravsaini04:
```
root@e41698c17c02:/# cat /tmp/container-init.log
[Fri 02 Aug 2024 08:16:09 AM UTC] ** SCRIPT START **
[Fri 02 Aug 2024 08:16:09 AM UTC] Running "/etc/init.d/dbus start".
[Fri 02 Aug 2024 08:16:09 AM UTC] Starting Xtigervnc.
```

samruddhikhandale: Hmm.. the log seems fine. Thanks! 

@gauravsaini04 Then do you know why we are seeing `/usr/local/share/desktop-init.sh: line 50: sudo: command not found` ? 

Can you check if we see the same message when we run tests on `main`? Also, can you compare the `/tmp/container-init.log` logs as well? 

Want to make sure we don't break things due to this change!

samruddhikhandale (approved):
Looks great to me!

Issue #1012, opened 2024-06-20
common-utils: Red Hat distros are missing various commands but are available


StephenGennard:
The following commands are missing but are available:

- tree
- bzip2
- dialog

````shell
root ➜ /workspaces/HelloWorld $ sudo microdnf -y install tree bzip2 dialo
g
Package                                            Repository        Size
Installing:                                                              
 bzip2-1.0.8-8.el9.x86_64                          baseos         53.1 kB
 dialog-1.3-32.20210117.el9.0.1.x86_64             appstream     244.6 kB
 tree-1.8.0-10.el9.x86_64                          baseos         56.1 kB
Transaction Summary:
 Installing:        3 packages
 Reinstalling:      0 packages
 Upgrading:         0 packages
 Obsoleting:        0 packages
 Removing:          0 packages
 Downgrading:       0 packages
Downloading packages...
Running transaction test...
Installing: dialog;1.3-32.20210117.el9.0.1;x86_64;appstream
Installing: tree;1.8.0-10.el9;x86_64;baseos
Installing: bzip2;1.0.8-8.el9;x86_64;baseos
Complete.
````

Validated on:

````shell
root ➜ /workspaces/HelloWorld $ sudo microdnf -y install tree bzip2 dialo
g
Package                                            Repository        Size
Installing:                                                              
 bzip2-1.0.8-8.el9.x86_64                          baseos         53.1 kB
 dialog-1.3-32.20210117.el9.0.1.x86_64             appstream     244.6 kB
 tree-1.8.0-10.el9.x86_64                          baseos         56.1 kB
Transaction Summary:
 Installing:        3 packages
 Reinstalling:      0 packages
 Upgrading:         0 packages
 Obsoleting:        0 packages
 Removing:          0 packages
 Downgrading:       0 packages
Downloading packages...
Running transaction test...
Installing: dialog;1.3-32.20210117.el9.0.1;x86_64;appstream
Installing: tree;1.8.0-10.el9;x86_64;baseos
Installing: bzip2;1.0.8-8.el9;x86_64;baseos
Complete.
````

samruddhikhandale:
Hi 👋 

The utilities `tree, bzip2, dialog, and ncdu` are commonly used in Unix-like operating systems for various tasks related to file and directory management, compression, and user interface creation in shell scripts. Feel free to open a PR if you are interested. Thanks!

StephenGennard:
see https://github.com/devcontainers/features/pull/1029

PR #1029, opened 2024-07-03
Pull request for issue #1012

StephenGennard:
I have added the files and conditionally installed ncdu is present

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -196,7 +196,10 @@ install_redhat_packages() {
             grep \
             which \
             man-db \
-            strace"
+            strace \
+            tree \
+            bzip2 \
+            dialog"
 
         # rockylinux:9 installs 'curl-minimal' which clashes with 'curl'
         # Install 'curl' for every OS except this rockylinux:9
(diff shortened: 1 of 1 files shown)


StephenGennard:

src/common-utils/main.sh
```diff
@@ -196,7 +196,10 @@ install_redhat_packages() {
             grep \
             which \
             man-db \
-            strace"
+            strace \
+            tree \
+            bzip2 \
```
samruddhikhandale, line 201: Makes sense to install them as this is widely used.

samruddhikhandale (commented):
Thanks for the contribution. 

Let's hold off this PR until we hear more feedback or interest from the community for these packages.

We'd be curious to hear your use case and how frequently do you use these packages. Thanks!

src/common-utils/main.sh
spgennard: Hi, 

The tools we noticed were missing was tree and bzip2.

bzip2 was found threw some build scripts, so it could be changed.

The rest were found throw comparison rather than a specific need for them.

PR #1040, opened 2024-07-14
[Rust] Update devcontainer-feature.json to include rust versions until 1.79

anoopknayak:
I was getting errors when trying to build a devcontainer in my project as one of the dependencies require rustc == 1.79. More details below:

```
18.03 error: failed to compile `cargo-binstall v1.7.4`, intermediate artifacts can be found at `/tmp/cargo-installZIsWeo`.
18.03 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
18.03 
18.03 Caused by:
18.03   package `cargo-binstall v1.7.4` cannot be built because it requires rustc 1.79.0 or newer, while the currently active rustc version is 1.74.0
18.03   Try re-running cargo install with `--locked`
```

Ran `devcontainer features test -f rust` and all tests passed:
```
✅ Passed:      'rust'
✅ Passed:      'rust_at_pinned_version'
✅ Passed:      'rust_with_target'
```

I'm hoping see if this would resolve it as I'd be able to pin rust version 1.79 after this change.


src/rust/devcontainer-feature.json
```diff
@@ -10,6 +10,9 @@
             "proposals": [
                 "latest",
                 "none",
+                "1.79",
```
samruddhikhandale, line 13: Can you help bump patch version of this Feature? thanks! 

https://github.com/devcontainers/features/blob/a7c5141c6888d90d5fa1daa33b217f37cee39b36/src/rust/devcontainer-feature.json#L3

samruddhikhandale (changes requested):

src/rust/devcontainer-feature.json
anoopknayak: Updated. Could you have a look again. Thanks in advance!

samruddhikhandale (approved):

PR #1045, opened 2024-07-18
[rust] - remove crates deprecated ext. - install dependi ext. instead

gauravsaini04:
**Feature Name**

* Rust

**Changes**

This PR brings in the following changes to the features repo / rust feature :  

* Crates extension is deprecated now, as given in issue [#1131](https://github.com/devcontainers/images/issues/1131), so instead of that installing dependi extension (fill-labs.dependi).

**Checklist**

### DIFF
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.2.0",
+    "version": "1.2.1",
     "name": "Rust",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/rust",
     "description": "Installs Rust, common Rust utilities, and their required dependencies",
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -56,8 +56,7 @@
             "extensions": [
                 "vadimcn.vscode-lldb",
                 "rust-lang.rust-analyzer",
-                "tamasfe.even-better-toml",
-                "fill-labs.dependi"
+                "tamasfe.even-better-toml"
             ],
             "settings": {
                 "files.watcherExclude": {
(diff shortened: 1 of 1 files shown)


src/rust/devcontainer-feature.json
```diff
@@ -57,7 +57,7 @@
                 "vadimcn.vscode-lldb",
                 "rust-lang.rust-analyzer",
                 "tamasfe.even-better-toml",
-                "serayuzgur.crates"
```
samruddhikhandale, line 60: Can we bump the minor version for this Feature?

src/rust/devcontainer-feature.json
```diff
@@ -57,7 +57,7 @@
...
+                "fill-labs.dependi"
```
samruddhikhandale, line 60: Instead of replacing, can we simply remove the crates extension? 

We should wait for user feedback and do a security check before adding a dependency on `fill-labs.dependi`

samruddhikhandale (changes requested):

src/rust/devcontainer-feature.json
gauravsaini04: ok sure

src/rust/devcontainer-feature.json
gauravsaini04: sure

samruddhikhandale: Looks like you bumped patch version, can we do minor instead?

src/rust/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.2.0",
+    "version": "1.2.1",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.3.0",
```

samruddhikhandale: Looks like you bumped patch version, can we do minor instead?

samruddhikhandale (approved):

PR #1048, opened 2024-07-19
common-utils: add bash-completion

### LINKED ISSUE #1043: bash completion not working out of the box
Since the dev containers' main purpose is to provide an interactive development environment, I think that shell autocomplete should be enabled by default in dev containers. The `bash-completion` package should be included with `common-utils` feature.

The `bash-completion` package could be included conditionally, if INSTALL_ZSH is not set.

dmytrokyrychuk:
This change increases the size of the base images by about 1MB. Below is the size comparison of the `base-alpine` image with and without the `bash-completion` package:

```
$ docker images --format "{{.Repository}}:{{.Tag}}\t{{.Size}}"
without-bash-completion:latest  664MB
with-bash-completion:latest     665MB
```

Resolve #1043

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.4.4",
+    "version": "2.4.5",
     "name": "Common Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",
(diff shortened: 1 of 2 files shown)


dmytrokyrychuk:

samruddhikhandale (approved):
Thanks!

dmytrokyrychuk:
The pipeline failure appears to be due to a DNS problem, unrelated to the changes made in this PR. Here is the relevant line from the build log:

> curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"

samruddhikhandale:
@dmytrokyrychuk Would you be able to help fix the tests? You'd need to add similar changes as https://github.com/devcontainers/features/blob/main/src/git/install.sh#L35-L41

That would be very helpful, thanks!

dmytrokyrychuk:
@samruddhikhandale, I applied @trxcllnt's fix to the common-utils feature and verified that the feature installs on centos by running `devcontainer features test -f common-utils --filter centos-7 --skip-autogenerated .`.

samruddhikhandale (approved):
Great, thanks!

PR #1049, opened 2024-07-19
Fixed Unsupported Debian distro

gvatsal60:
Updated the conditional check to support some other debian based distro as well, which identifies as Debian-like but doesn't match the exact string comparison for "debian". This change allows the script to correctly identify and execute on Linux Mint systems, ensuring compatibility.

**Also trimmed TrailingWhitespace**

**Changes Made**
Modified the conditional statement in the script from:
```
if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = "debian" ]; then
```
to:

```
if [ "${ID}" = "debian" ] || [ "${ID_LIKE}" = *"debian"* ]; then
```
**Reason for Change**
While based on Ubuntu and Debian, does not set "ID_LIKE" to "debian" but includes "debian" in its list of like distributions. This change ensures the script recognizes distros as a valid Debian-like distribution and performs necessary actions accordingly.

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -104,7 +104,7 @@ install_debian_packages() {
 
     # Needed for adding manpages-posix and manpages-posix-dev which are non-free packages in Debian
     if [ "${ADD_NON_FREE_PACKAGES}" = "true" ]; then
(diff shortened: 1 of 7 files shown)


gvatsal60:

gvatsal60:
Plenty of fixes are required in what I have done, will do it soon.

Issue #1041, opened 2024-07-17
Powershell feature fails to install version 7.4.3 OR 7.4.4 properly for linux arm64


DonPwrShellHunt:
# PowerShell feature install of 7.4.3 for linux arm64 Issue

## Issue with install.sh

If an install using GitHub is attempted, install.sh logic for devcontainers/features/src/powershell does not explicitly set pwsh as executable. Successful installation currently depends on the permissions in the tar file.

### Symptoms of Failure

If I pinned the version to 7.4.2, the powershell feature would function correctly. Looking at the permissions inside the tar file showed the execute bit was set on **pwsh**.

```bash
tar -tvf /Users/donhunt/local/powershell-7.4.2-linux-arm64.tar.gz pwsh
-rwxr-xr-x  0 cloudtest cloudtest 72544 Apr  9 18:48 pwsh
```

Using an explicit version of 7.4.3, latest or not specifying a version, **pwsh** could not be run because permissions were not set to execute. The install did not return an error code. It just did not function at runtime.

```bash
tar -tvf /Users/donhunt/local/powershell-7.4.3-linux-arm64.tar.gz pwsh
-rw-rw-r--  0 root   root    72544 Jun 18 10:15 pwsh
```

NOTE: The instructions from the powershell team to install powershell into linux using the binaries **does** include setting pwsh as an executable. In my case I am running MacOS on a MacBook Pro with an M1 processor. Docker Desktop is installed using a linux arm64 virtual machine. When installing devcontainer features, such as powershell, install scripts are running as root in a linux arm64 environment.

In the case of installing from github, powershell-7.4.3-linux-arm64.tar.gz is extracted into /opt/microsoft/powershell/7 but the pwsh executable does not have proper permissions to run. 

## Fix install.sh

Add chmod after the extract to make pwsh executable.

```bash

223	tar xf "${powershell_filename}" -C "${powershell_target_path}"
224	chmod 755 "${powershell_target_path}/pwsh"		
225	ln -s "${powershell_target_path}/pwsh" /usr/local/bin/pwsh
```

## Fixed as the vscode user with the commands

```bash
cd /opt/microsoft/powershell/7
sudo chmod 755 pwsh
```

## Log snippets showing before and after fix

```bash
vscode ➜ /opt/microsoft/powershell/7 $ ls -l pwsh
-rw-rw-r-- 1 root root  72544 Jun 18 14:15 pwsh
...
[...]

PR #1050, opened 2024-07-19
`[powershell]`: Fix manual PowerShell installation binary execution permissions

jpawlowski:
This patch aligns the manual setup of `pwsh` with the settings the official Debian module uses:

- symlink to `/usr/bin/pwsh` instead of `/usr/bin/local/pwsh`
- register as shell in `/etc/shells` (using `add-shell`)

Also, because of #1041, this will also enforce execution permissions of the `/opt/microsoft/powershell/7/pwsh` binary.

The package version minor version number is increased to `1.5.0`. Also, version suggestions for PowerShell installation were updated to remove deprecated version `7.1` and add version `7.2`, `7.3` and `7.4` to the selection (they can still be installed as this is only the suggestions presented to the user).

Some other maintenance work:

- Add Debian Bookworm and Ubuntu Noble archive names to allow install of packaged versions from them
- Make the feature idempotent to improve when the feature is installed multiple times
- Repaired broken profile test
- Added a new `validate_powershell_installation` test to validate general `pwsh` binary installation

### DIFF
--- a/src/powershell/devcontainer-feature.json
+++ b/src/powershell/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "powershell",
-    "version": "1.4.0",
+    "version": "1.5.0",
     "name": "PowerShell",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/powershell",
     "description": "Installs PowerShell along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like gpg.",
@@ -10,7 +10,9 @@
             "proposals": [
                 "latest",
                 "none",
-                "7.1"
+                "7.4",
+                "7.3",
+                "7.2"
             ],
             "default": "latest",
             "description": "Select or enter a version of PowerShell."
--- a/src/powershell/install.sh
+++ b/src/powershell/install.sh
@@ -224,6 +224,7 @@ install_using_github() {
     chmod 755 "${powershell_target_path}/pwsh"
     ln -sf "${powershell_target_path}/pwsh" /usr/bin/pwsh
     add-shell "/usr/bin/pwsh"
+    cd /tmp
     rm -rf /tmp/pwsh
 }
 
--- a/src/powershell/install.sh
+++ b/src/powershell/install.sh
@@ -221,7 +221,9 @@ install_using_github() {
         echo "${powershell_archive_sha256} *${powershell_filename}" | sha256sum -c -
     fi
     tar xf "${powershell_filename}" -C "${powershell_target_path}"
-    ln -s "${powershell_target_path}/pwsh" /usr/local/bin/pwsh
+    chmod 755 "${powershell_target_path}/pwsh"
+    ln -s "${powershell_target_path}/pwsh" /usr/bin/pwsh
+    add-shell "/usr/bin/pwsh"
     rm -rf /tmp/pwsh
 }
 
--- a/src/powershell/install.sh
+++ b/src/powershell/install.sh
@@ -222,26 +222,30 @@ install_using_github() {
     fi
     tar xf "${powershell_filename}" -C "${powershell_target_path}"
     chmod 755 "${powershell_target_path}/pwsh"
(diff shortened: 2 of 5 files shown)


src/powershell/install.sh
```diff
@@ -221,25 +221,31 @@ install_using_github() {
         echo "${powershell_archive_sha256} *${powershell_filename}" | sha256sum -c -
     fi
     tar xf "${powershell_filename}" -C "${powershell_target_path}"
-    ln -s "${powershell_target_path}/pwsh" /usr/local/bin/pwsh
```
samruddhikhandale, line 224: Should we keep this symbolic linking intact? I am afraid if folks might be relying on this.

src/powershell/install.sh
```diff
@@ -221,25 +221,31 @@ install_using_github() {
...
+    chmod 755 "${powershell_target_path}/pwsh"
+    ln -sf "${powershell_target_path}/pwsh" /usr/bin/pwsh
```
samruddhikhandale, line 225: We have intentionally attempted to install binaries in `/usr/local/bin` due to following differences 👇 

/usr/local/bin:

- Typically used for software installed manually by the system administrator.
- Preferred for custom or third-party software installations.
- Keeps the system's default binaries (/usr/bin) separate from manually installed binaries.
- Higher precedence in the PATH variable on many systems, meaning binaries here are found first.

/usr/bin:

- Used for binaries provided by the system's package manager.
- Standard location for system-wide binaries.
- Modifying this directory can potentially interfere with system updates or package management.

Can you help me understand your thoughts for this change? Thanks!

src/powershell/devcontainer-feature.json
```diff
@@ -10,7 +10,9 @@
             "proposals": [
                 "latest",
                 "none",
-                "7.1"
+                "7.4",
```
samruddhikhandale, line 13: Can we add an explicit test to validate these changes work for the issue mentioned in https://github.com/devcontainers/features/issues/1041? Thank you!

samruddhikhandale (changes requested):
Thank you so much for taking the time to contribute these changes!

src/powershell/install.sh
jpawlowski: The intention is to mimic the exact same setup from Debian repository packages.
I am familiar with the differences between `/usr/local/bin` and `/usr/bin`. In my opinion, we are not a typical user installing something manually, we are acting similar to package maintainers for what `/usr/bin` is exactly for.

On top of that, the GitHub installation is more an automatic fallback where people might still expect the very same outcome as if they had been lucky to have a packaged version available.

src/powershell/install.sh
jpawlowski: it is down in line 225, with added "force" to overwrite any potentially existing symlink. The reason for that is that otherwise, when other features use `dependsOn` to refer to the PowerShell feature, and they also explicitly define the PowerShell feature to install modules, it does not play nice with existing symlinks, leading to a build failure.

src/powershell/devcontainer-feature.json
jpawlowski: I am going to add a test for that particular version.

DonPwrShellHunt:
Is there something blocking completion of this pull request?
I'm just trying to gauge how I can assist in getting #1041 resolved...

samruddhikhandale:
@DonPwrShellHunt Thanks for the nudge! We are waiting for the author to add test coverage validating his change as well as fix the broken test.

src/powershell/devcontainer-feature.json
jpawlowski: Added a new test with 2ea64a6b0df8859ae9c697885743a465970ad330

jpawlowski:
@samruddhikhandale would you be kind enough to review this PR and, if all is well, merge it into the `main` branch along with a new build of the `powershell` feature package?

samruddhikhandale (approved):
Thanks you, looks good!

PR #1056, opened 2024-07-23
Fix installing latest git from PPA in bionic

### LINKED ISSUE #1055: Installing latest git from PPA in bionic fails fetching GPG keys
After https://github.com/devcontainers/features/commit/f45518999e598127c46ece79a607a241eb5e38d6, installing the latest git from PPA in `ubuntu:bionic` fails fetching GPG keys:

```
 > [dev_containers_target_stage  5/18] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=git_1,target=/tmp/build-features-src/git_1     cp -ar /tmp/build-features-src/git_1 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/git_1  && cd /tmp/dev-container-features/git_1  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/git_1:
gpg: key A1715D88E1DF1F24: no user ID
(*) Failed getting key, retrying in 10s...
(*) Downloading GPG key...
gpg: key A1715D88E1DF1F24: no user ID
(*) Failed getting key, retrying in 10s...
(*) Downloading GPG key...
gpg: key A1715D88E1DF1F24: no user ID
(*) Failed getting key, retrying in 10s...
(!) Failed to get gpg key.
ERROR: Feature "Git (from source)" (ghcr.io/devcontainers/features/git) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/git for help troubleshooting this error.
```

The issue can be replicated by changing the default from "os-provided" to "latest" in `src/git/devcontainer-feature.json`, then running the generated test via
```shell
BUILDKIT_PROGRESS=plain devcontainer features test --skip-scenarios -f git --base-image ubuntu:18.04 --project-folder .
```

trxcllnt:
`keys.openpgp.org` strips user IDs from keys unless the key owner grants permission to share them. This [leads to](https://superuser.com/a/1485255) gpg rejecting the keys instead of importing them, and failing to install the `git` feature.

This PR enforces an explicit order when testing and importing keys from the keyservers list, ensuring `keys.openpgp.org` is the last keyserver to be tried. Previously, keyservers were selected based on bash's associative array key enumeration order, but this order is different between Bash v4 (in Ubuntu Bionic) and v5 (Ubuntu Focal+).

I believe the issue isn't seen in Bash v5 because the ubuntu keyservers are enumerated first, not because the gpg version in Focal accepts keys without user IDs.

If this is the case, we should either stop attempting to import keys from `keys.openpgp.org`, or encourage the owner of `E1DD270288B4E6030699E45FA1715D88E1DF1F24` to grant `openpgp.org` permission to share their user ID.

Fixes #1055 and adds a new scenario test.

Note: I believe this fix may need to be applied to all features modified in https://github.com/devcontainers/features/pull/1016.

### DIFF
--- a/src/git/install.sh
+++ b/src/git/install.sh
@@ -266,17 +266,7 @@ elif [ "${ADJUSTED_ID}" = "alpine" ]; then
     check_packages asciidoc curl-dev expat-dev g++ gcc openssl-dev pcre2-dev perl-dev perl-error python3-dev tcl tk xmlto
 
 elif [ "${ADJUSTED_ID}" = "rhel" ]; then
-
-    if [ $VERSION_CODENAME = "centos7" ]; then
-        check_packages centos-release-scl
-        check_packages devtoolset-11
-        source /opt/rh/devtoolset-11/enable
-    else
-        check_packages gcc
-    fi
-
-
-    check_packages libcurl-devel expat-devel gettext-devel openssl-devel perl-devel zlib-devel cmake pcre2-devel tar gzip ca-certificates
+    check_packages gcc libcurl-devel expat-devel gettext-devel openssl-devel perl-devel zlib-devel cmake pcre2-devel tar gzip ca-certificates
     if ! type curl > /dev/null 2>&1; then
         check_packages curl
     fi
--- a/src/git/install.sh
+++ b/src/git/install.sh
@@ -26,10 +26,18 @@ elif [ "${ID}" = "alpine" ]; then
     ADJUSTED_ID="alpine"
 elif [[ "${ID}" = "rhel" || "${ID}" = "fedora" || "${ID}" = "mariner" || "${ID_LIKE}" = *"rhel"* || "${ID_LIKE}" = *"fedora"* || "${ID_LIKE}" = *"mariner"* ]]; then
     ADJUSTED_ID="rhel"
-    VERSION_CODENAME="${ID}{$VERSION_ID}"
+    VERSION_CODENAME="${ID}${VERSION_ID}"
 else
     echo "Linux distro ${ID} not supported."
     exit 1
+fi
+
+if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
+    # As of 1 July 2024, mirrorlist.centos.org no longer exists.
+    # Update the repo files to reference vault.centos.org.
+    sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
+    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo
+    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
 fi
 
 if type apt-get > /dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


trxcllnt:
[It appears](https://serverfault.com/a/1161847) `mirrorlist.centos.org` no longer exists, and that's why the centos-7 scenario tests are failing.

src/git/install.sh
```diff
@@ -254,17 +266,7 @@ elif [ "${ADJUSTED_ID}" = "alpine" ]; then
     check_packages asciidoc curl-dev expat-dev g++ gcc openssl-dev pcre2-dev perl-dev perl-error python3-dev tcl tk xmlto
 
 elif [ "${ADJUSTED_ID}" = "rhel" ]; then
-
-    if [ $VERSION_CODENAME = "centos7" ]; then
-        check_packages centos-release-scl 
-        check_packages devtoolset-11
-        source /opt/rh/devtoolset-11/enable
```
samruddhikhandale, line 261: Was this removed because `VERSION_CODENAME` could have never been `centos7` due to the bug in https://github.com/devcontainers/features/pull/1056/files#diff-3dc3a1c6b68eed85dc6e030494337eeb7dacc6203f0db79595f4152878712508L29?

src/git/install.sh
```diff
@@ -66,37 +74,41 @@ clean_up
...
+        if curl -s ${curl_args} --max-time 5 "${keyserver_curl_url}" > /dev/null; then
             echo "keyserver ${keyserver}"
             keyserver_reachable=true
         else
             echo "(*) Keyserver ${keyserver} is not reachable." >&2
         fi
-    done
+    }
+
+    # Explicitly test these in order because Bash v4.4.20 (Ubuntu Bionic)
```
samruddhikhandale, line 95: Nice 🎉

src/git/install.sh
```diff
@@ -26,12 +26,20 @@ elif [ "${ID}" = "alpine" ]; then
...
     ADJUSTED_ID="rhel"
-    VERSION_CODENAME="${ID}{$VERSION_ID}"
+    VERSION_CODENAME="${ID}${VERSION_ID}"
 else
     echo "Linux distro ${ID} not supported."
     exit 1
 fi
 
+if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
+    # As of 1 July 2024, mirrorlist.centos.org no longer exists.
```
samruddhikhandale, line 36: Thanks for fixing the tests!

samruddhikhandale (approved):
Thank you for the fix, appreciate it.

Let's merge once the tests are green!

samruddhikhandale:
> Note: I believe this fix may need to be applied to all features modified in https://github.com/devcontainers/features/pull/1016.

Most likely, yes! 
Let's see how this PR works and we could add similar fix to others.

src/git/install.sh
trxcllnt: Yes, exactly. Once I fixed the value of `$VERSION_CODENAME`, the tests failed attempting to install `devtoolset-11`.  I removed this to align with the previously accidentally-working behavior :slightly_smiling_face:.

Issue #1023, opened 2024-06-28
docker-in-docker feature breaks with Codespaces+Kind and docker v27+ (docker's ipv6 breaking changes)


jeremysf:
We have been using Visual Studio Code, GitHub Codespaces and the docker-in-docker feature with Kubernetes' kind project (https://kind.sigs.k8s.io/) for several years now.

Recently, with the release of docker version 27, things broke. When trying to use the kind command line utility to create a new kind cluster (i.e. launch the kind docker container which encapsulates a Kubernetes cluster) we get the following error:

```
Creating cluster "kind" ...
ERROR: failed to create cluster: failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1
Command Output: Error response from daemon: Failed to Setup IP tables: Unable to enable NAT rule:  (iptables failed: ip6tables --wait -t nat -I POSTROUTING -s fc00:f853:ccd:e793::/64 ! -o br-94aea5e559a6 -j MASQUERADE: ip6tables v1.8.7 (legacy): can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)
Perhaps ip6tables or your kernel needs to be upgraded.
 (exit status 3))
```

The issue is I think related to this change by docker:

https://docs.docker.com/engine/release-notes/27.0/#ipv6

I think what is needed is the ability to do this from the release notes:

```
To restore the behavior of earlier releases, no ip6tables at all, set "ip6tables": false in daemon.json, or use the CLI option --ip6tables=false. Alternatively, leave ip6tables enabled, publish ports, and enable direct routing.
```

Temporarily, I am able to get things working again with the following feature configuration in our devcontainer.json:
```
    "features": {
        "ghcr.io/devcontainers/features/common-utils:1": {
            "installZsh": true,
            "upgradePackages": false,
            "uid": "1000",
            "gid": "1000",
            "installOhMyZsh": "true",
            "nonFreePackages": "true"
        },
        "ghcr.io/devcontainers/features/docker-in-docker:2.2.1": {
            "version": "26.1.3",
            "enableNonRootDocker": true,
            "moby": false
        }
    },
```

Is it possible to extend the feature with an option to disable `ip6tables` and/or to pass in additional command line options to the launch of docker?

samruddhikhandale:
[...]

PR #1068, opened 2024-07-29
[docker-in-docker] - toggle ip6tables settings value as option

gauravsaini04:
**Feature Name**

* Docker - in - Docker

**_Changelog_**

* Changes to `install.sh` and `devcontainer-feature.json` files for features' issue [#1023](https://github.com/devcontainers/features/issues/1023)

**Checklist**


src/docker-in-docker/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "docker-in-docker",
-    "version": "2.11.0",
+    "version": "2.11.1",
```
samruddhikhandale, line 3:
```suggestion
    "version": "2.12.0",
```

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -55,6 +55,11 @@
             "type": "boolean",
             "default": true,
             "description": "Install Compose Switch (provided docker compose is available) which is a replacement to the Compose V1 docker-compose (python) executable. It translates the command line into Compose V2 docker compose then runs the latter."
+        },
+        "ip6tables": {
+            "type": "boolean",
+            "default": true,
+            "description": "toggle ip6tables setting"
```
samruddhikhandale, line 62:
```suggestion
            "description": "Disable ip6tables"
```

src/docker-in-docker/install.sh
```diff
@@ -20,6 +20,7 @@ INSTALL_DOCKER_COMPOSE_SWITCH="${INSTALLDOCKERCOMPOSESWITCH:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy noble"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy noble"
+TOGGLE_IP6TABLES_SETTING="${IP6TABLES:-"true"}"
```
samruddhikhandale, line 23: Do we not need any other changes to the Feature? How is the  `TOGGLE_IP6TABLES_SETTING` applied?

samruddhikhandale: From https://github.com/devcontainers/features/issues/1023#issuecomment-2221238973

> (Preferable) Test if we can pass --ip6tables=false on the condition when we start docker with dockerd
If not, let's look at creating or modifying the daemon.json file to include the ip6tables setting

samruddhikhandale, line 23: Can we add a test with `ip6tables:false` and test if the scenario mentioned by user works with these changes?

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: done

gauravsaini04: done

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -55,6 +55,11 @@
             "type": "boolean",
             "default": true,
             "description": "Install Compose Switch (provided docker compose is available) which is a replacement to the Compose V1 docker-compose (python) executable. It translates the command line into Compose V2 docker compose then runs the latter."
+        },
+        "disableIp6tables": {
+            "type": "boolean",
+            "default": "",
```
samruddhikhandale, line 61:
```suggestion
            "default": "false",
```

The schema check will error out while releasing this Feature

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -55,6 +55,11 @@
...
+            "description": "Disable ip6tables"
```
samruddhikhandale, line 62:
```suggestion
            "description": "Disable ip6tables (this option is only applicable for Docker versions 27 and greater)"
```

src/docker-in-docker/install.sh
```diff
@@ -20,6 +20,7 @@ INSTALL_DOCKER_COMPOSE_SWITCH="${INSTALLDOCKERCOMPOSESWITCH:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy noble"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy noble"
+DISABLE_IP6_TABLES="${DISABLEIP6TABLES:-""}"
```
samruddhikhandale, line 23:
```suggestion
DISABLE_IP6_TABLES="${DISABLEIP6TABLES:-"false"}"
```

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,35 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
     find "${docker_home}" -type d -print0 | xargs -n 1 -0 chmod g+s
 fi
 
+if [ -z "$DISABLE_IP6_TABLES" ]; then
+    DOCKER_DEFAULT_IP6_TABLES=""
```
samruddhikhandale, line 473: Update according to previous comment

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,35 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+else
+    version_to_check=""
+    semver_regex="^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$"
+    if echo "$DOCKER_VERSION" | grep -Eq "$semver_regex"; then
+        version_to_check=$(echo $DOCKER_VERSION | cut -d. -f1)
+    elif echo "$DOCKER_VERSION" | grep -Eq "^-?[0-9]+$"; then
+        version_to_check=$DOCKER_VERSION
+    fi
+
+    if [ -n "$version_to_check" ] && [ "$version_to_check" -ge 27 ] || [ "$DOCKER_VERSION" = "latest" ]; then
```
samruddhikhandale, line 483:
```suggestion
    if [[ -n "$version_to_check" ] && [ "$version_to_check" -ge 27 ]] || [ "$DOCKER_VERSION" = "latest" ]; then
```

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,35 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+            if [ "$DISABLE_IP6_TABLES" == false ]; then
+                DOCKER_DEFAULT_IP6_TABLES=""
```
samruddhikhandale, line 485: We don't need to handle this separately for > v27 and lower. Let's just set `DOCKER_DEFAULT_IP6_TABLES` to empty by default and only handle `true` cases

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: for the two cases such that v26 and lower, script exits out seeing the lesser version and disableIp6tables value as `false` whereas for v27 and up, script simply does add the flag --ip6tables=false when disableIp6tables equalling `true` is provided by user as the config option, the if conditional statement shall be used

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
-else
-    version_to_check=""
-    semver_regex="^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$"
-    if echo "$DOCKER_VERSION" | grep -Eq "$semver_regex"; then
-        version_to_check=$(echo $DOCKER_VERSION | cut -d. -f1)
-    elif echo "$DOCKER_VERSION" | grep -Eq "^-?[0-9]+$"; then
-        version_to_check=$DOCKER_VERSION
-    fi
+DOCKER_DEFAULT_IP6_TABLES=""
+version_to_check=""
```
samruddhikhandale, line 473:
```suggestion
requested_version=""
```

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+semver_regex="^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$"
```
samruddhikhandale, line 474: Curious, where did you grab this from? 👀

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+if echo "$DOCKER_VERSION" | grep -Eq "$semver_regex"; then
```
samruddhikhandale, line 475: Can you add a comment on what's happening here?

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
-    else 
+if [[ -n "$version_to_check" && "$version_to_check" -ge 27 ]] || [ "$DOCKER_VERSION" = "latest" ]; then
         if [ "$DISABLE_IP6_TABLES" == true ]; then
-            echo "Info: Ip6tables feature is already disabled in docker engine v26 & lower."
-            DOCKER_DEFAULT_IP6_TABLES=""
-        else
-            echo "Error: Ip6tables in docker engine v26 & lower is an experimental feature. Enabling is not supported."
-            exit 1
+            echo "Info: Ip6tables feature is enabled by default in docker engine v27 & later. Disabling.."
+            DOCKER_DEFAULT_IP6_TABLES="--ip6tables=false"
```
samruddhikhandale, line 484:
```suggestion
            DOCKER_DEFAULT_IP6_TABLES="--ip6tables=false"
            echo "(!) As requested, passing '${DOCKER_DEFAULT_IP6_TABLES}'"
```

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
         fi
+else 
+    if [ "$DISABLE_IP6_TABLES" == false ]; then
```
samruddhikhandale, line 487: Shouldn't this be `true` ?

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+        echo "Error: Ip6tables in docker engine v26 & lower is an experimental feature. Enabling is not supported. Aborting.."
```
samruddhikhandale, line 488:
```suggestion
        echo "ERR: Passing --ip6tables=false is not supported for Docker v26 and below... Remove 'disableIp6tables:true' from Feature options..."
```

src/docker-in-docker/install.sh
```diff
@@ -469,32 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+        exit 1
```
samruddhikhandale, line 489: Is it possible to add tests for Docker >= 27 and `disableIp6tables:true` ?

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: actually, v26 and lower already has this disabled by default so testing true doesn't make sense, added a condition check if the user actually wanted it to be false i.e. trying to enable the setting of `--ip6tables=true`, only this condition makes sense for v26 and lower

src/docker-in-docker/install.sh
gauravsaini04: it's already there `dockerIp6tablesDisabledTest`

src/docker-in-docker/install.sh
gauravsaini04: changed !

src/docker-in-docker/install.sh
gauravsaini04: changed !

src/docker-in-docker/install.sh
gauravsaini04: comment added.

src/docker-in-docker/install.sh
gauravsaini04: from some web resource

src/docker-in-docker/install.sh
gauravsaini04: changed

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,29 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
     find "${docker_home}" -type d -print0 | xargs -n 1 -0 chmod g+s
 fi
 
+DOCKER_DEFAULT_IP6_TABLES=""
+requested_version=""
+# checking whether the version requested either is in semver format or just a number denoting the major version
+# extracting the major version number out of the two scenarios
+semver_regex="^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$"
```
samruddhikhandale, line 476: Why do we need such a complex semver? Why can't we use `^(latest|none|lts|[0-9]+)`?

src/docker-in-docker/install.sh
samruddhikhandale: Is passing `--ip6tables` supported for docker v26 and lower?

gauravsaini04: image

yeah it is supported as is written in the docs

src/docker-in-docker/install.sh
gauravsaini04: ok will change..

src/docker-in-docker/install.sh
samruddhikhandale: Got it, thanks for looking into it!  

So given the complimentary situation, let's only bother for v27 and up, and avoid conflicts and changing default behavior. 

Do nothing if `DOCKER_DEFAULT_IP6_TABLES=false`, and only pass `--ip6tables=true` if `DOCKER_DEFAULT_IP6_TABLES=true`

@gauravsaini04 Does that make sense?

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: changed code

src/docker-in-docker/install.sh
gauravsaini04: this semver regex checks for entire semver i.e. 1.2.3 pattern, which will confirm if semver has actually been passed so can pick the major version from it

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,24 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
```
samruddhikhandale, line 476:
```suggestion
semver_regex="^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
```

Can we use ^ and comment that it's taken from https://semver.org/

samruddhikhandale, line 476: Can we move the new code to the block only if `DISABLE_IP6_TABLES = true`?

This would avoid unwanted time wasted in computations

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: I did use it but had to convert to a bash compatible regex

gauravsaini04: sure, thanks !

src/docker-in-docker/install.sh
```diff
@@ -468,6 +469,23 @@ if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
...
+    requested_version=""
+    # checking whether the version requested either is in semver format or just a number denoting the major version
+    # and, extracting the major version number out of the two scenarios
+    semver_regex="^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?$"
+    if echo "$DOCKER_VERSION" | grep -Eq $semver_regex; then
+        requested_version=$(echo $DOCKER_VERSION | cut -d. -f1)
+    elif echo "$DOCKER_VERSION" | grep -Eq "^[1-9][0-9]*$"; then
+        requested_version=$DOCKER_VERSION
+    fi
+    if [[ -n "$requested_version" && "$requested_version" -ge 27 ]] || [ "$DOCKER_VERSION" = "latest" ]; then
```
samruddhikhandale, line 483:
```suggestion
    if [ "$DOCKER_VERSION" = "latest" ] || [[ -n "$requested_version" && "$requested_version" -ge 27 ]] ; then
```

samruddhikhandale (dismissed):
Looks wonderful ⚡

Issue #1035, opened 2024-07-10
`common-utils` overwrites `.zshrc` coming from image built with `common-utils` already


AliOsm:
I'm using [`rails/ruby`](https://github.com/rails/devcontainer/pkgs/container/devcontainer%2Fimages%2Fruby) image, which already [uses](https://github.com/rails/devcontainer/blob/main/images/ruby/.devcontainer/devcontainer.json#L7-L13) `common-utils`. But I want to override some options like making Zsh the default shell and install OhMyZsh. When I add:

```json
"ghcr.io/devcontainers/features/common-utils:latest": {
  "installOhMyZsh": true,
  "configureZshAsDefaultShell": true
}
```

To my `devcontainer.json`, this `common-utils` usage overwrites the `.zshrc` coming from `rails/ruby` image, which should not happen based on [this](https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#L438-L449) `common-utils` script.

Any thoughts why this could happen?

AliOsm:
I think I found something, when I add `common-utils` as follows:

```json
"ghcr.io/devcontainers/features/common-utils:latest": {
  "installOhMyZsh": false,
  "installOhMyZshConfig": false,
  "configureZshAsDefaultShell": true
}
```

IT is not overwriting the `.zshrc` file, but enabling OhMyZsh stuff as follows:

```json
"ghcr.io/devcontainers/features/common-utils:latest": {
  "installOhMyZsh": true,
  "installOhMyZshConfig": true,
  "configureZshAsDefaultShell": true
}
```

Overwrites it. So, I think the bug lives in [this](https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#L505-L545) snippet.

AliOsm:
Specifically, [this](https://github.com/devcontainers/features/blob/main/src/common-utils/main.sh#L534) line overwrites the `.zshrc` file.

AliOsm:
So, the issue is as follows: in `rails/ruby` feature, an `rbenv` initialization line is added to the `.zshrc` file ([ref](https://github.com/rails/devcontainer/blob/main/features/ruby/install.sh#L25-L27)). When I enable OhMyZsh, `.zshrc` file changes to OhMyZsh template and the initialization line disappears, so I need to add the following to my post create script in my DevContainer setup to add the initialization again:

```bash
echo 'eval "$(rbenv init -)"' >> /home/vscode/.zshrc
eval "$(rbenv init -)"
```

[...]

PR #1069, opened 2024-07-31
common-utils overwrites .zshrc coming from image built

prathameshzarkar9:
[Issue 1035](https://github.com/devcontainers/features/issues/1035)
common-utils overwrites .zshrc coming from image built with common-utils already

**Feature Name**
- common-utils

**Changelog**
- Changes to main.sh for features issue

**Checklist**
- [x] user provided commands appended to .zshrc file should not overwrite


src/common-utils/main.sh
```diff
@@ -491,7 +491,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
         echo 'source $HOME/.profile' >> "${user_home}/.zprofile" # TODO: Reconsider adding '.profile' to '.zprofile'
```
samruddhikhandale, line 491: Let's bump the Feature version and add a new test scenario (instead of updating the existing one) which validates the issue mentioned by the user in https://github.com/devcontainers/features/issues/1035

prathameshzarkar9: version bumped to : 2.4.6
Also created a scenario `configure_zshrc_without_overwrite` to validate the issue mentioned by the user, kindly review it once.

src/common-utils/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.4.5",
+    "version": "2.4.6",
```
samruddhikhandale, line 3: Looks like we have multiple test failures due to this PR change, can you help fix those? 

https://github.com/devcontainers/features/actions/runs/10196886412/job/28208566441?pr=1069#step:4:1167

samruddhikhandale (changes requested):

test/common-utils/scenarios.json
```diff
@@ -189,6 +189,22 @@
             }
         }
     },
+    "configure_zshrc_without_overwrite": {
+        "image": "ghcr.io/rails/devcontainer/images/ruby:3.2.5",
```
samruddhikhandale, line 193:
```suggestion
        "image": "mcr.microsoft.com/devcontainers/ruby:3.2",
```

src/common-utils/devcontainer-feature.json
prathameshzarkar9: Done. resolved

src/common-utils/main.sh
```diff
@@ -518,6 +518,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
     # See https://github.com/ohmyzsh/ohmyzsh/blob/master/tools/install.sh for official script.
     if [ "${INSTALL_OH_MY_ZSH}" = "true" ]; then
         user_rc_file="${user_home}/.zshrc"
+        copy_user_rc_file="${user_home}/.zshrc-copy"
```
samruddhikhandale, line 521: Isn't `user_home` a temporary file? So can we create a new temp folder and add it there? 

Once the copying over is completed, let's delete the folder

src/common-utils/main.sh
```diff
@@ -543,8 +544,14 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
 
         # Add devcontainer .zshrc template
         if [ "$INSTALL_OH_MY_ZSH_CONFIG" = "true" ]; then
+            if [ -f "$user_rc_file" ]; then
+                cp -f ${user_rc_file} ${copy_user_rc_file}
+            fi
             echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
             sed -i -e 's/ZSH_THEME=.*/ZSH_THEME="devcontainers"/g' ${user_rc_file}
+            if [ -f "$user_rc_file" ]; then
+                diff ${user_rc_file} ${copy_user_rc_file} | grep '^>' | sed 's/^>\ //' >> ${user_rc_file}
```
samruddhikhandale, line 553: I don't think we should use `diff` as it has potential drawbacks 👇 

image

Are there any concerns in appending the new contents to the end of the existing file?

test/common-utils/scenarios.json
samruddhikhandale: We don't want our tests to depend on a third party and less well-known image as they have the potential of breaking on top of having security concerns.

Can we use a Dockerfile and recreate the scenario?

samruddhikhandale (changes requested):

src/common-utils/main.sh
prathameshzarkar9: @samruddhikhandale 
There is no concern appending a new contents to the end of the existing file , the concern is, when user is appending a content to the existing file `~/.zshrc` the file is overwritten by the zshrc_template file and all the content added by user is getting disappeared from the file. 

before [this](https://github.com/devcontainers/features/pull/1069/files/6f61ee04b67b8f8bdb382455c8b4e607f12e5633#diff-3e71c0a0669a0410f7dd0d8f2b83f3b6bf6b525d3eabd354f19f32822da669fcR550) the user_rc_file has user provided content already appended, but after this it is overwritten by the template file.

can you suggest alternate option if diff is not recommended

test/common-utils/scenarios.json
prathameshzarkar9: I have discarded this scenario and created a new one as suggested by you.
name of the new scenario with Dockerfile is called `devcontainer-ruby-zshrc`
Kindly review the new code commit

src/common-utils/main.sh
prathameshzarkar9: discarding this too as copying a file is not required now

src/common-utils/main.sh
prathameshzarkar9: I have introduced a new logic without affecting the existing logic.
now, if the `.zshrc` file exist then we will just update the necessary attributes from the file and if the named file is not exist then only we will create a file with the default zshrc-template file
hope this logic will work as expected. Kindly review it once

samruddhikhandale: Yep, makes sense! Looking

src/common-utils/main.sh
```diff
@@ -543,8 +542,11 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
 
         # Add devcontainer .zshrc template
         if [ "$INSTALL_OH_MY_ZSH_CONFIG" = "true" ]; then
-            echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
-            sed -i -e 's/ZSH_THEME=.*/ZSH_THEME="devcontainers"/g' ${user_rc_file}
+            if [ ! -f "$user_rc_file" ]; then
+                echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
```
samruddhikhandale, line 546: Now if the user has `user_rc_file` file, then this will not add the zshrc template. This doesn't make sense! 

Looking at https://github.com/devcontainers/features/issues/1035, the user can control and avoid this overwriting by setting `installOhMyZshConfig ` to `false`.

Instead of appending, let's prepend the contents to the beginning of the file. This ensures, the file will still prioritize user contents, but will have the necessary contents from this repo.

test/common-utils/devcontainer-ruby-zshrc/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/devcontainers/ruby:3.2
+
```
samruddhikhandale, line 2: Can we fix the failing tests?

samruddhikhandale (changes requested):

src/common-utils/main.sh
prathameshzarkar9: what if user has this file but does not follow the template? in that case, still user has control over the content irrespective of appending the content in this file.

samruddhikhandale: I think that should be fine! User contents take priority over the content the Feature adds in

test/common-utils/devcontainer-ruby-zshrc/Dockerfile
prathameshzarkar9: fixed

src/common-utils/main.sh
prathameshzarkar9: as suggested, kindly review the changes done for this.

src/common-utils/main.sh
```diff
@@ -543,7 +542,7 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
 
         # Add devcontainer .zshrc template
         if [ "$INSTALL_OH_MY_ZSH_CONFIG" = "true" ]; then
-            echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
+            echo -e "$(cat "${template_path}")\n$(cat ${user_rc_file})\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
```
samruddhikhandale, line 545: Won't this throw file not found error?

samruddhikhandale, line 545: Also, is it possible to append before `template_path` ?

samruddhikhandale (changes requested):

src/common-utils/main.sh
prathameshzarkar9: appended before template_path

prathameshzarkar9: if statement again introduced to check user_rc_file exist.
if yes, appending the file content to template_path file content and overwritting the user_rc_file
if no, just using the template_path content and overwritting the user_rc_file which creates the file too

src/common-utils/main.sh
```diff
@@ -543,7 +542,11 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
 
         # Add devcontainer .zshrc template
         if [ "$INSTALL_OH_MY_ZSH_CONFIG" = "true" ]; then
-            echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > ${user_rc_file}
+            if [ -f "${user_rc_file}" ]; then
+                echo -e "$(cat "${user_rc_file}")\n$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > "${user_rc_file}"
+            else
+                echo -e "$(cat "${template_path}")\nDISABLE_AUTO_UPDATE=true\nDISABLE_UPDATE_PROMPT=true" > "${user_rc_file}"
+            fi
```
samruddhikhandale, line 549: There are many duplicates in this snippet, can we optimize the code?

prathameshzarkar9: optimisation done. kindly review once.

samruddhikhandale (approved):
Looks great, thanks!

Mark-J-Lawrence:
I think this change has broken devcontainers using `ghcr.io/devcontainers/features/node:1`. `node` & `npm` are not found after building the container

I have had to specify `ghcr.io/devcontainers/features/common-utils:2.4.5` to get it building with `node` & `npm` again.

Issue #1062, opened 2024-07-25
Features: Update logic for PPA in bionic


samruddhikhandale:
> Note: I believe this fix may need to be applied to all features modified in https://github.com/devcontainers/features/pull/1016.

Most likely, yes! 
Let's see how this PR works and we could add similar fix to others.

_Originally posted by @samruddhikhandale in https://github.com/devcontainers/features/issues/1056#issuecomment-2251572313_

samruddhikhandale:
@prathameshzarkar9 Can you add this [code snippet](https://github.com/devcontainers/features/blob/main/src/git/install.sh#L35-L41) (if missing) to [these Features](https://github.com/search?q=repo%3Adevcontainers%2Ffeatures+Redhat+language%3AMarkdown&type=code) that support CentOs? 

Also, make sure to bump patch version for the Features in your PR.

prathameshzarkar9:
Hi @samruddhikhandale 
Created PR for this https://github.com/devcontainers/features/pull/1079

prathameshzarkar9:
Hi @samruddhikhandale ,
All the review comments have been addressed. Test scenarios also modified accordingly to check the required pnpm is getting installed.
kindly review the same

PR #1079, opened 2024-08-06
Features update logic for PPA in bionic

prathameshzarkar9:
[Issue 1062](https://github.com/devcontainers/features/issues/1062)
Features: Update logic for PPA in bionic

**Feature Name**
- common-utils
- go
- java
- node
- python

**Changelog**
- Changes to install.sh and devcontainer-feature.json for features update

**Checklist**
- [x] code snippet is added at specific position in all the install.sh of mentioned features
- [x] Bumped all the version for the mentioned features

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -362,6 +362,14 @@ if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
     sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
 fi
 
+if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
+    # As of 1 July 2024, mirrorlist.centos.org no longer exists.
+    # Update the repo files to reference vault.centos.org.
+    sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
+    sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo
+    sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
+fi
+
 # Install packages for appropriate OS
 case "${ADJUSTED_ID}" in
     "debian")
--- a/src/java/devcontainer-feature.json
+++ b/src/java/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.6.1",
+  "version": "1.6.0",
   "name": "Java (via SDKMAN!)",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/java",
   "description": "Installs Java, SDKMAN! (if not installed), and needed dependencies.",
--- a/src/java/devcontainer-feature.json
+++ b/src/java/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "name": "Java (via SDKMAN!)",
   "documentationURL": "https://github.com/devcontainers/features/tree/main/src/java",
   "description": "Installs Java, SDKMAN! (if not installed), and needed dependencies.",
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -31,6 +31,7 @@
         "pnpmVersion": {
             "type": "string",
             "proposals": [
+                "latest",
                 "8.8.0",
                 "8.0.0",
                 "7.30.0",
@@ -38,8 +39,8 @@
                 "5.18.10",
                 "none"
             ],
-            "default": "8.8.0",
-            "description": "Version of PNPM to install."
+            "default": "latest",
+            "description": "Select or enter the PNPM version to install"
         },
         "nvmVersion": {
             "type": "string",
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.5.0",
+    "version": "1.5.1",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.5.1",
+    "version": "1.6.0",
     "name": "Node.js (via nvm), yarn and pnpm",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -38,7 +38,7 @@
                 "5.18.10",
                 "none"
             ],
-            "default": "none",
+            "default": "8.8.0",
             "description": "Version of PNPM to install."
         },
         "nvmVersion": {
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -28,6 +28,19 @@
             "default": "/usr/local/share/nvm",
             "description": "The path where NVM will be installed."
         },
+        "pnpmVersion": {
+            "type": "string",
+            "proposals": [
+                "8.8.0",
+                "8.0.0",
+                "7.30.0",
+                "6.14.8",
+                "5.18.10",
+                "none"
+            ],
+            "default": "none",
+            "description": "Version of PNPM to install."
+        },
         "nvmVersion": {
             "type": "string",
             "proposals": [
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -8,7 +8,7 @@
 # Maintainer: The Dev Container spec maintainers
 
 export NODE_VERSION="${VERSION:-"lts"}"
-export PNPM_VERSION="${PNPMVERSION:-"none"}"
+export PNPM_VERSION="${PNPMVERSION:-"latest"}"
 export NVM_VERSION="${NVMVERSION:-"latest"}"
 export NVM_DIR="${NVMINSTALLPATH:-"/usr/local/share/nvm"}"
 INSTALL_TOOLS_FOR_NODE_GYP="${NODEGYPDEPENDENCIES:-true}"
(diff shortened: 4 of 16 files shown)


src/java/install.sh
```diff
@@ -53,6 +53,14 @@ else
     exit 1
 fi
 
+if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
```
samruddhikhandale, line 56: Let's resolve merge conflicts!

src/node/install.sh
```diff
@@ -379,7 +387,7 @@ else
             [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
             [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
             [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
-            npm install -g pnpm
+            npm install -g @pnpm/exe
```
samruddhikhandale, line 390: I think this is updated to fix the tests? However, both of these packages are different. 

Latest version of `pnpm` is not supported for node 16. Then, only for node 16, can we condition to install `pnpm 7` ? Then keep the rest as is! 

See https://pnpm.io/installation#compatibility

samruddhikhandale (changes requested):

src/node/install.sh
prathameshzarkar9: Done

src/java/install.sh
prathameshzarkar9: resolved

src/node/install.sh
```diff
@@ -370,20 +378,20 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
 fi
 
 # Install pnpm
-if bash -c ". '${NVM_DIR}/nvm.sh' && type pnpm >/dev/null 2>&1"; then
-    echo "pnpm already installed."
```
samruddhikhandale, line 374: Why was this removed?

src/node/install.sh
```diff
@@ -370,20 +378,20 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
...
+    (
+        . "${NVM_DIR}/nvm.sh"
+        [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
+        [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
+        [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
+        if [ "${NODE_VERSION}" = "16" ]; then
+            npm install -g --force pnpm@7
+        else
+            npm install -g --force pnpm
+        fi
```
samruddhikhandale, line 391:
```
if [ "${NODE_VERSION}" = "16" ]; then
    PNPM_VERSION="pnpm@7"
else
    PNPM_VERSION="pnpm"
fi

npm install -g --force ${PNPM_VERSION}
```

Can we optimize?

samruddhikhandale (changes requested):

src/node/install.sh
prathameshzarkar9: few images like `centos:centos7` has `pnpm` pre-installed which does not support the node version 16 so we should eliminate this condition and force install pnpm with the installed npm to support the compatible version of pnpm for node 16.

src/node/install.sh
prathameshzarkar9: done

src/node/install.sh
samruddhikhandale: Thanks for the thoughts here, however, we shouldn't override pnpm version if it already exists. This would have side effects on other distros which are neither centos nor are attempting to install node 16 

> few images like centos:centos7 has pnpm pre-installed which does not support the node version 16 

Did you find our tests failing?

src/node/install.sh
```diff
@@ -370,20 +378,21 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
...
-if bash -c ". '${NVM_DIR}/nvm.sh' && type pnpm >/dev/null 2>&1"; then
-    echo "pnpm already installed."
+if bash -c ". '${NVM_DIR}/nvm.sh' && type npm >/dev/null 2>&1"; then
+    (
+        . "${NVM_DIR}/nvm.sh"
+        [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
+        [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
+        [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
+        PNPM_VERSION="pnpm"
+        if [ "${NODE_VERSION}" = "16" ]; then
```
samruddhikhandale, line 388: Rethinking about this, in future it will end up piling too many conditions.

Is there a dynamic way to find out the compatibility between node versions and pnpm?

src/node/install.sh
prathameshzarkar9: kindly check below, if we use the default pnpm which is pre-installed in centos:7 scenaio. it is failing for pnpm -v check
```
The current version of Node.js is v16.20.2
Visit https://r.pnpm.io/comp to see the list of past pnpm versions with respective Node.js version support.


❌ pnpm check failed.
🧪 Executing duplicate test for feature 'node'...
⚠️ Skipping duplicate test for node because '/workspaces/features/test/node/duplicate.sh' does not exist.



❌ Failed:      'centos-7'

@prathameshzarkar9 ➜ /workspaces/features (update-logic-for-ppa) $ docker exec -it 2f bash
[root@2f6f7f406396 /]# pnpm -v
ERROR: This version of pnpm requires at least Node.js v18.12
The current version of Node.js is v16.20.2
Visit https://r.pnpm.io/comp to see the list of past pnpm versions with respective Node.js version support.
[root@2f6f7f406396 /]# 
```

prathameshzarkar9: @samruddhikhandale, reverting this change. let me know your thought on this.
adding a code to install pnpm specific version compatible for the installed node version

src/node/install.sh
```diff
@@ -370,6 +378,32 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
 fi
 
 # Install pnpm
+# Declare an associative array
+declare -A node_pnpm_map
+
+# Populate the associative array with Node.js version to pnpm version mappings
+node_pnpm_map["10.13.0"]="5.18.10"
```
samruddhikhandale, line 385: Thanks for adding it. I was hoping if there was a way to to dynamically find the compatible version without hardcoding, however, seems like there isn't a way.

Can we revert these changes? Let's add a new "pnpmVersion" Feature option where users can pin the `pnpm` version if they need to fix the compatibility issues.

Hence, in the test case which fails, we would use `pnpmVersion` to fix it.

samruddhikhandale (changes requested):

src/node/install.sh
prathameshzarkar9: removed

src/common-utils/main.sh
```diff
@@ -362,6 +362,14 @@ if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
     sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
 fi
 
+if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
```
samruddhikhandale, line 365: Can we bump patch version? Looks like it got lost due to merging with main

src/java/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "java",
-  "version": "1.6.0",
+  "version": "1.6.1",
```
samruddhikhandale, line 3: Why do we need to bump this given that we are not making any Java Feature changes?

src/node/devcontainer-feature.json
```diff
@@ -28,6 +28,19 @@
...
+            "type": "string",
+            "proposals": [
+                "8.8.0",
+                "8.0.0",
+                "7.30.0",
+                "6.14.8",
+                "5.18.10",
+                "none"
+            ],
+            "default": "8.8.0",
```
samruddhikhandale, line 41:
```suggestion
            "default": "latest",
```

Can you update to allow this ^ ?

src/node/devcontainer-feature.json
```diff
@@ -28,6 +28,19 @@
...
+            "description": "Version of PNPM to install."
```
samruddhikhandale, line 42:
```suggestion
            "description": "Select or enter the PNPM version to install"
```

src/node/install.sh
```diff
@@ -8,6 +8,7 @@
 # Maintainer: The Dev Container spec maintainers
 
 export NODE_VERSION="${VERSION:-"lts"}"
+export PNPM_VERSION="${PNPMVERSION:-"none"}"
```
samruddhikhandale, line 11:
```suggestion
export PNPM_VERSION="${PNPMVERSION:-"latest"}"
```

samruddhikhandale: This should match with the defaults

src/node/install.sh
```diff
@@ -370,7 +379,7 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
 fi
 
 # Install pnpm
-if bash -c ". '${NVM_DIR}/nvm.sh' && type pnpm >/dev/null 2>&1"; then
+if bash -c ". '${NVM_DIR}/nvm.sh' && [ ! -z "${PNPM_VERSION}" ] && [ "${PNPM_VERSION}" = "none" ]"; then
     echo "pnpm already installed."
```
samruddhikhandale, line 383:
```suggestion
if [ ! -z "${PNPM_VERSION}" ] && [ "${PNPM_VERSION}" = "none" ]"; then
    echo "Ignoring installation of PNPM"
```

test/node/scenarios.json
```diff
@@ -21,7 +22,8 @@
         "features": {
             "node": {
                 "version": "lts",
-                "additionalVersions": "v17.9.1,v14.19.3"
+                "additionalVersions": "v17.9.1,v14.19.3",
+                "pnpmVersion": "6.16.0"
```
samruddhikhandale, line 26: Can we update the tests to verify the required version is installed?

samruddhikhandale (changes requested):

src/common-utils/main.sh
prathameshzarkar9: patch version bumped to "2.5.1"

src/node/devcontainer-feature.json
prathameshzarkar9: done

src/node/devcontainer-feature.json
prathameshzarkar9: done

src/node/install.sh
prathameshzarkar9: done

src/node/install.sh
prathameshzarkar9: done

test/node/scenarios.json
prathameshzarkar9: required version is verified in all the test cases for pnpmVersion

src/node/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.5.0",
+    "version": "1.5.1",
```
samruddhikhandale, line 3:
```suggestion
    "version": "1.6.0",
```

samruddhikhandale (approved):
Looks good!

PR #1081, opened 2024-08-07
fix(aws-cli): 🔧 specify installation directory

rubensa:
Despite the [official doc](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) states:

<quote>
--install-dir or -i – This option specifies the directory to copy all of the files to.

The default value is /usr/local/aws-cli
</quote>

It looks like the [default value](https://github.com/aws/aws-cli/blob/2.17.24/scripts/install#L201) is `~/.local/lib/aws` which makes it to be innacesible if the package is installed by **root** and the devcointainer is later run as **non-root** user.


rubensa:
My mistake.

This is not needed.  It is installing in the right place.

I realized that my problem was related with the `verify_aws_cli_gpg_signature` is calling `gpg` that generates some files under _~/.gnupg_ folder that, if you have a `USER` and `ENV HOME` defined in your **Dockerfile** ends up with access errors (as the files are owned by **root**, cause the feature is always installed using **root**, but they are created under **USER HOME** directory).

rubensa:
Just in case someone gets the same problem...

I had to add this to my Dockerfile, so the _~/.gnupg_ folder is created with the right permissions before the feature installation is run by **root** (the **Fix** is optional but useful to avoid ipv6 network problems).

```Docker
# In case it's the first time that the user runs gpg and the directory ~/.gnupg/ doesn't exist yet
RUN <<EOT
#!/bin/bash -l
gpg -k
# Fix: "gpg: keyserver receive failed: Cannot assign requested address"
# see: https://github.com/usbarmory/usbarmory-debian-base_image/issues/9#issuecomment-451635505
echo "disable-ipv6" >> ~/.gnupg/dirmngr.conf
EOT
```

PR #1082, opened 2024-08-08
Desktop lite: Fix quote error

mdahshan:
Corrected a syntax error in line 414 

log "Executing "\$@\"."

was corrected to

log "Executing \"$@\"."

### DIFF
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -411,7 +411,7 @@ else
 fi
 
 # Run whatever was passed in
-log "Executing "\$@\"."
+log "Executing \"$@\"."
 exec "$@"
 log "** SCRIPT EXIT **"
 EOF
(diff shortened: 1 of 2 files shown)


mdahshan:
```
/usr/local/share/desktop-init.sh: line 120: unexpected EOF while looking for matching `"'
```

src/desktop-lite/install.sh
```diff
@@ -411,7 +411,7 @@ else
 fi
 
 # Run whatever was passed in
-log "Executing "\$@\"."
+log "Executing \"$@\"."
```
samruddhikhandale, line 414: Can you bump the patch version of this Feature? Thanks! 

https://github.com/devcontainers/features/blob/01c0c4b76721aaedfaf450a1d6c0b497eaf6d916/src/desktop-lite/devcontainer-feature.json#L3

samruddhikhandale (dismissed):
Thanks for the quick fix, left one minor comment and we are all set to merge this in!

mdahshan:
Thank you!

src/desktop-lite/install.sh
samruddhikhandale: Let me know when it's ready for another review

samruddhikhandale (approved):

PR #1095, opened 2024-08-15
Modify scripts to stream terminal title changes

osortega:
Adding the capabilities to signal terminal title changes based on the command that is being executed


src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -1,4 +1,3 @@
-
 # bash theme - partly inspired by https://github.com/ohmyzsh/ohmyzsh/blob/master/themes/robbyrussell.zsh-theme
 __bash_prompt() {
     local userpart='`export XIT=$? \
```
samruddhikhandale, line 3: Can we bump the minor version? Thanks

https://github.com/devcontainers/features/blob/38481091ba5f070d51c51053d908197926f50824/src/common-utils/devcontainer-feature.json#L3

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -23,3 +22,20 @@ __bash_prompt() {
...
+# Function to reset the terminal title to the shell type after the command is executed
+precmd() {
+    echo -ne "\033]0;${USER}@${HOSTNAME}: ${SHELL}\007"
+}
+
+# Trap DEBUG signal to call preexec before each command
+trap 'preexec' DEBUG
+
+# Set the PROMPT_COMMAND to call precmd before displaying the prompt
+PROMPT_COMMAND='precmd'
```
samruddhikhandale, line 41: Can we append to the PROMPT_COMMAND instead of overwriting? because if the user has other commands set in PROMPT_COMMAND, they will be lost.

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -23,3 +22,20 @@ __bash_prompt() {
...
+    echo -ne "\033]0;${USER}@${HOSTNAME}: ${cmd}\007"
+}
+
+# Function to reset the terminal title to the shell type after the command is executed
+precmd() {
+    echo -ne "\033]0;${USER}@${HOSTNAME}: ${SHELL}\007"
+}
+
+# Trap DEBUG signal to call preexec before each command
+trap 'preexec' DEBUG
```
samruddhikhandale, line 38: Using the `DEBUG` trap can introduce performance overhead because it triggers the `preexec` function before every command.

Can we wrap these new changes within the condition `"${TERM_PROGRAM}" = "codespaces"` ? If we could have a new TERM_PROGRAM for `workspace` then it's even better.

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -23,3 +22,20 @@ __bash_prompt() {
 }
 __bash_prompt
 export PROMPT_DIRTRIM=4
+
+# Function to set the terminal title to the command being executed
+preexec() {
+    local cmd=$(history 1 | sed 's/^[ ]*[0-9]*[ ]*//')
```
samruddhikhandale, line 28: Should we check if `history` command works?  The `history 1` command might not work as expected in all contexts, especially if the history is not enabled or if the shell history settings are different.

src/common-utils/scripts/devcontainers.zsh-theme
```diff
@@ -24,3 +24,19 @@ __zsh_prompt() {
     unset -f __zsh_prompt
 }
 __zsh_prompt
+
+# Function to set the terminal title to the current command
```
samruddhikhandale, line 28: Same comment here, can we condition this to use `TERM_PROGRAM`?

src/common-utils/scripts/devcontainers.zsh-theme
```diff
@@ -24,3 +24,19 @@ __zsh_prompt() {
...
+preexec() {
+    local cmd=${1}
+    echo -ne "\033]0;${USER}@${HOST}: ${cmd}\007"
```
samruddhikhandale, line 31: Should this be `HOSTNAME` ?

src/common-utils/scripts/devcontainers.zsh-theme
```diff
@@ -24,3 +24,19 @@ __zsh_prompt() {
...
+}
+
+# Function to reset the terminal title to the shell type after the command is executed
+precmd() {
+    echo -ne "\033]0;${USER}@${HOST}: ${SHELL}\007"
```
samruddhikhandale, line 36: Same here?

src/common-utils/scripts/rc_snippet.sh
```diff
@@ -24,3 +23,20 @@ if [ -z "$(git config --get core.editor)" ] && [ -z "${GIT_EDITOR}" ]; then
...
+# Function to reset the terminal title to the shell type after the command is executed
+precmd() {
+    echo -ne "\033]0;${USER}@${HOSTNAME}: ${SHELL}\007"
+}
+
+# Trap DEBUG signal to call preexec before each command
+trap 'preexec' DEBUG
+
+# Set the PROMPT_COMMAND to call precmd before displaying the prompt
+PROMPT_COMMAND='precmd'
```
samruddhikhandale, line 42: Similar comments as mentioned for bash_theme_snippet.sh.

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -23,3 +22,20 @@ __bash_prompt() {
 }
 __bash_prompt
 export PROMPT_DIRTRIM=4
```
samruddhikhandale, line 24: Can you help fix the test failures? thanks!

samruddhikhandale (changes requested):

src/common-utils/scripts/bash_theme_snippet.sh
osortega: Changed it to 2.4.8

src/common-utils/scripts/bash_theme_snippet.sh
osortega: Added a check for `xterm` which is what we use for workspace

src/common-utils/scripts/bash_theme_snippet.sh
osortega: Changed it so we use `BASH_COMMAND` instead

src/common-utils/scripts/devcontainers.zsh-theme
osortega: Checking $TERM now

src/common-utils/scripts/devcontainers.zsh-theme
osortega: Changed to HOSTNAME

src/common-utils/scripts/rc_snippet.sh
osortega: Changed it

src/common-utils/scripts/devcontainers.zsh-theme
osortega: Changed it

src/common-utils/scripts/bash_theme_snippet.sh
osortega: Appending it now

samruddhikhandale (changes requested):

src/common-utils/scripts/bash_theme_snippet.sh
samruddhikhandale: @osortega Can you check why https://github.com/osortega/features/blob/0556f86d4db352c75888c04cd303200bc5cc44f3/test/common-utils/configure_zsh_as_default_shell.sh#L12 check started failing with your PR changes? 

Now the test is green but I think that's because of the `TERM` condition.

test/common-utils/scenarios.json
```diff
@@ -259,5 +259,22 @@
...
+            "TERM": "xterm"
+        }
+    },
+    "no-terminal-title-by-default": {
+        "image": "node",
+        "features": {
+            "common-utils": {}
+        },
+        "containerEnv": {
+        }
```
samruddhikhandale, line 278:
```suggestion
        }
```

Let's remove as we are not adding any env variables?

src/common-utils/scripts/bash_theme_snippet.sh
```diff
@@ -23,19 +23,23 @@ __bash_prompt() {
...
 
-# Trap DEBUG signal to call preexec before each command
-trap 'preexec' DEBUG
+    # Trap DEBUG signal to call preexec before each command
+    trap 'preexec' DEBUG
 
-# Set the PROMPT_COMMAND to call precmd before displaying the prompt
-PROMPT_COMMAND='precmd'
+    # Append precmd to PROMPT_COMMAND if it is already set
+    # Append to PROMPT_COMMAND to call precmd before displaying the prompt
```
samruddhikhandale, line 43:
```suggestion
```

duplicate

samruddhikhandale (approved):
Looks great, thank you! 

Let's wait for the tests to pass and let's merge this on Monday

src/common-utils/scripts/bash_theme_snippet.sh
osortega: This is fixed

samruddhikhandale (dismissed):

test/common-utils/terminal-title-on-xterm.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+# Optional: Import test library
+source dev-container-features-test-lib
+
```
samruddhikhandale, line 7: Can you fix the failing test? Thanks!

osortega: Fixed it :)

samruddhikhandale (approved):

PR #1100, opened 2024-08-20
Ensure `common-utils` rc snippets end with newlines

trxcllnt:
Adds line endings to common-utils' rc snippets after https://github.com/devcontainers/features/pull/1095.

This ensures any features appending to these files after installing `common-utils` don't produce mangled files.

what happens now:
```shell
$ echo "export foo=bar" >> ~/.bashrc
$ tail -n1 ~/.bashrc
fiexport foo=bar
```

what used to/should happen:
```shell
$ echo "export foo=bar" >> ~/.bashrc
$ tail -n1 ~/.bashrc
export foo=bar
```

Maybe also a good idea to add the following vscode settings to the repo's `devcontainer.json` (and/or encourage contributors to use these settings?): 
```json
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
```


samruddhikhandale (approved):
Thanks for the quick fix!

PR #1101, opened 2024-08-20
[docker-in-docker] - method 2 - daemon.json creation - #1023

gauravsaini04:
**Feature Name**
* Docker-in-Docker

**Description**

* Provides an alternate solution to #1023 
* An attempt for creation of `/etc/docker/daemon.json` file before starting the docker daemon
* This is a workaround for the docker-in-docker feature breaking with Codespaces+Kind and docker v27+ (docker's ipv6 breaking changes)

**_changelog_**

* Added a new method to create the `/etc/docker/daemon.json` file before starting the docker daemon


src/docker-in-docker/install.sh
```diff
@@ -20,6 +20,7 @@ INSTALL_DOCKER_COMPOSE_SWITCH="${INSTALLDOCKERCOMPOSESWITCH:-"true"}"
 MICROSOFT_GPG_KEYS_URI="https://packages.microsoft.com/keys/microsoft.asc"
 DOCKER_MOBY_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal jammy noble"
 DOCKER_LICENSED_ARCHIVE_VERSION_CODENAMES="bookworm buster bullseye bionic focal hirsute impish jammy noble"
+IP6_TABLES="${IP6TABLES:-""}"
```
samruddhikhandale, line 23: Can we publicize this Feature option as mentioned in https://github.com/devcontainers/features/issues/1023#issuecomment-2221238973?

samruddhikhandale, line 23: Can we look into fixing test failures?

src/docker-in-docker/install.sh
```diff
@@ -561,6 +563,20 @@ dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAU
         DEFAULT_ADDRESS_POOL="--default-address-pool $DOCKER_DEFAULT_ADDRESS_POOL"
     fi
 
+    if [ -z "$DOCKER_DEFAULT_IP6_TABLES" ]
+    then
+        DEFAULT_IP6_TABLES=true
+    else
+        DEFAULT_IP6_TABLES=$DOCKER_DEFAULT_IP6_TABLES
+    fi
```
samruddhikhandale, line 571: Why don't we set `DOCKER_DEFAULT_IP6_TABLES` to `true` if the value is not set in the code above? It will avoid this.

src/docker-in-docker/install.sh
```diff
@@ -561,6 +563,20 @@ dockerd_start="AZURE_DNS_AUTO_DETECTION=${AZURE_DNS_AUTO_DETECTION} DOCKER_DEFAU
...
+
+    # Create the daemon.json file with the provided value
+    cat <<JSON_EOF > /etc/docker/daemon.json
+{
+    "ip6tables": $DEFAULT_IP6_TABLES
```
samruddhikhandale, line 576: Instead of this, have you tried using `--ip6tables=<value>` same like DEFAULT_ADDRESS_POOL ?

samruddhikhandale (changes requested):

src/docker-in-docker/install.sh
gauravsaini04: Pls check for the above mentioned method as a solution [here](https://github.com/devcontainers/features/pull/1068)

src/docker-in-docker/install.sh
gauravsaini04: done 👍

src/docker-in-docker/install.sh
gauravsaini04: when I test for `ip6tables -L` command inside built container, getting the following output: 
```
root@a139636130dc:/# ip6tables -L
ip6tables v1.8.4 (legacy): can't initialize ip6tables table `filter': Table does not exist (do you need to insmod?)
Perhaps ip6tables or your kernel needs to be upgraded.
```

gauravsaini04: done

src/docker-in-docker/install.sh
samruddhikhandale: If we have https://github.com/devcontainers/features/pull/1068 open, why do we need this PR which attempts to do the same thing?

gauravsaini04:
[#1068](https://github.com/devcontainers/features/pull/1068) is an alternate method as a solution for the same issue.

Issue #1084, opened 2024-08-10
[Java] Add JDK 22 support


alanondra:
JDK 22 was [released in March](https://openjdk.org/projects/jdk/22/) but the feature `ghcr.io/devcontainers/features/java:1` with `version` set to `latest` only goes up to 21.

samruddhikhandale:
@prathameshzarkar9 Can you help take a look at this?  Thanks!

prathameshzarkar9:
Hi @samruddhikhandale ,

Created a PR https://github.com/devcontainers/features/pull/1102

I have removed the get_jdk_distro function from install.sh as it was not being used correctly in the sequence of installing distro specific version until JDK_DISTRO is not passed as an argument with any specific value like: ms,tem etc.

prathameshzarkar9:
Hi @samruddhikhandale ,
kindly review the changes done in this PR to install the latest jdk 22

PR #1102, opened 2024-08-20
[Java] Add JDK 22 support

prathameshzarkar9:
[Issue 1084](https://github.com/devcontainers/features/issues/1084)
[Java] Add JDK 22 support

**Feature Name**
- java

**Changelog**
- Changes to install.sh for features issue 1084

**Checklist**
- [x] if pass version as latest, it will install java 22 which was released in march

### DIFF
--- a/src/java/install.sh
+++ b/src/java/install.sh
@@ -201,15 +201,16 @@ find_version_list() {
     install_type=$3
     ifLts="$4"
     version_list=$5
+    java_ver=$6
     
     check_packages jq
     all_lts_versions=$(curl -s https://api.adoptium.net/v3/info/available_releases)
     if [ "${ifLts}" = "true" ]; then 
         major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_lts')
-    elif [ "${VERSION}" = "latest" ]; then
+    elif [ "${java_ver}" = "latest" ]; then
         major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_feature_release') 
     else 
-        major_version=$(echo "${JAVA_VERSION}" | cut -d '.' -f 1)
+        major_version=$(echo "$java_ver" | cut -d '.' -f 1)
     fi
     
     if [ "${JDK_DISTRO}" = "ms" ]; then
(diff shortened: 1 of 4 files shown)


src/java/install.sh
```diff
@@ -195,29 +195,29 @@ check_packages() {
     esac
 }
 
-# Use Microsoft JDK for everything but JDK 8 and 18 (unless specified differently with jdkDistro option)
-get_jdk_distro() {
-    VERSION="$1"
```
samruddhikhandale, line 200: I think we still need it, right? Look at test failures https://github.com/devcontainers/features/actions/runs/10472188406/job/29001078551?pr=1102#step:4:670

samruddhikhandale (changes requested):

src/java/install.sh
```diff
@@ -59,6 +59,8 @@ if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
     sed -i s/mirror.centos.org/vault.centos.org/g /etc/yum.repos.d/*.repo
     sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo
     sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo
+    yum update -y
+    yum install epel-release -y
```
samruddhikhandale, line 63: Can we instead use `check_packages` function ? Would need to move around the installtion.

src/java/install.sh
```diff
@@ -59,6 +59,8 @@ if [ "${ADJUSTED_ID}" = "rhel" ] && [ "${VERSION_CODENAME-}" = "centos7" ]; then
...
 fi
```
samruddhikhandale, line 65: Can you write an explicit test to validate JDK 22 works fine?

src/java/install.sh
```diff
@@ -195,28 +196,33 @@ check_packages() {
...
+    
+    check_packages jq
+    all_lts_versions=$(curl -s https://api.adoptium.net/v3/info/available_releases)
     if [ "${ifLts}" = "true" ]; then 
-        all_lts_versions=$(curl -s https://api.adoptium.net/v3/info/available_releases)
         major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_lts')
-        regex="${prefix}\\K${major_version}\\.?[0-9]*\\.?[0-9]*${suffix}"
-    else 
+    elif [ "${java_ver}" = "latest" ]; then
+        major_version=$(echo "$all_lts_versions" | jq -r '.most_recent_feature_release')
```
samruddhikhandale, line 212: How is this latest related to LTS? 

Java could release 22 as latest and not make it LTS for a while, then this would fail, right?

samruddhikhandale (changes requested):

src/java/install.sh
samruddhikhandale: @prathameshzarkar9 Can you help respond to this comment?

samruddhikhandale (changes requested):

samruddhikhandale (approved):

PR #1104, opened 2024-08-21
[Go] Update proposals for the "version" option

sho-hata:
Go 1.23 was released (https://go.dev/doc/devel/release#go1.23.0).
Updated the version specified in the proposal following the release of Go 1.23.


samruddhikhandale (approved):
Thanks!

Issue #1078, opened 2024-08-05
failed to build with error: bash: "/usr/local/share/nvm/nvm.sh: No such file or directory"


SouthWest7:
When I try to open the folder in devcontainer, feature "Node.js (via nvm), yarn and pnpm" (ghcr.io/devcontainers/features/node) failed to install!
logs follows:
```
[+] Building 442.3s (17/18)                                                                    docker:desktop-linux
 => [internal] load build definition from Dockerfile.extended                                                  0.0s
 => => transferring dockerfile: 4.80kB                                                                         0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1.4                                    2.6s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io                                               0.0s
 => CACHED docker-image://docker.io/docker/dockerfile:1.4@sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c  0.0s
 => [internal] load .dockerignore                                                                              0.0s
 => => transferring context: 2B                                                                                0.0s
 => [context dev_containers_feature_content_source] load .dockerignore                                         0.0s
 => => transferring dev_containers_feature_content_source: 2B                                                  0.0s
 => [internal] load metadata for mcr.microsoft.com/vscode/devcontainers/base:ubuntu                            0.0s
 => [context dev_containers_feature_content_source] load from client                                           0.1s
 => => transferring dev_containers_feature_content_source: 170.74kB                                            0.0s
 => CACHED [dev_containers_feature_content_normalize 1/3] FROM mcr.microsoft.com/vscode/devcontainers/base:ub  0.0s
 => [dev_containers_feature_content_normalize 2/3] COPY --from=dev_containers_feature_content_source devconta  0.0s
 => [dev_containers_feature_content_normalize 3/3] RUN chmod -R 0755 /tmp/build-features/                      0.3s
 => CACHED [dev_containers_target_stage 2/8] RUN mkdir -p /tmp/dev-container-features                          0.0s
 => CACHED [dev_containers_target_stage 3/8] COPY --from=dev_containers_feature_content_normalize /tmp/build-  0.0s
 => CACHED [dev_containers_target_stage 4/8] RUN echo "_CONTAINER_USER_HOME=$( (command -v getent >/dev/null   0.0s
 => [dev_containers_target_stage 5/8] RUN --mount=type=bind,from=dev_containers_feature_content_source,sour  298.0s
[...]

PR #1114, opened 2024-09-05
Fixed Nvm Source Issue(#1078)

gvatsal60:
- Fixed nvm source issue

See the issue at #1078


src/node/install.sh
```diff
@@ -227,7 +227,7 @@ install_yarn() {
                 # Yum/DNF want to install nodejs dependencies, we'll use NPM to install yarn
                 su ${USERNAME} -c "umask 0002 && . '${NVM_DIR}/nvm.sh' && nvm use ${_ver} && npm install --global yarn"
```
samruddhikhandale, line 228: Can you help bump the patch version of this Feature? 

https://github.com/devcontainers/features/blob/a9247d8cdb3388d8849cf13ef5b0f5da862a84e7/src/node/devcontainer-feature.json#L3

samruddhikhandale (changes requested):
Apologies for the delay, this PR slipped through the cracks.

src/node/install.sh
gvatsal60: Updated the version to `1.6.1`

src/node/install.sh
```diff
@@ -309,7 +309,7 @@ curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/v${NVM_VERSION}/install.
     curl -so- "https://raw.githubusercontent.com/nvm-sh/nvm/\${PREV_NVM_VERSION}/install.sh" | bash
     NVM_VERSION="\${PREV_NVM_VERSION}"
 }
-source "${NVM_DIR}/nvm.sh"
+[ -s "\$NVM_DIR/nvm.sh" ] && . "\$NVM_DIR/nvm.sh"
```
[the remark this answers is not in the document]
gvatsal60: @samruddhikhandale

> @gvatsal60 Curious, why do we want to escape the $ sign?
  => As NVM_DIR is already set, it can be escaped here. However, to maintain consistency, I followed the suggestion.

> Can we continue using source instead of .?
    => Yes, that’s a great choice! Since this is a Bash script and not POSIX compliant, using `source` works well.

samruddhikhandale (approved):
Thank you, let's merge once the tests are green! ⚡

Thompson1985 (approved):

PR #1119, opened 2024-09-12
cuda support for 12.3 .. 12.5

prathameshzarkar9:


src/nvidia-cuda/install.sh
```diff
@@ -69,7 +71,15 @@ apt-get install -yq "$cuda_pkg"
 
 if [ "$INSTALL_CUDNN" = "true" ]; then
     # Ensure that the requested version of cuDNN is available AND compatible
-    cudnn_pkg_version="libcudnn8=${CUDNN_VERSION}-1+cuda${CUDA_VERSION}"
+    #if major cudnn version is 9, then we need to install libcudnn9-cuda-<major_version> package
+    #else we need to install libcudnn8-cuda-<major_version> package
+    if [[ $major_cudnn_version = "9" ]]
```
samruddhikhandale, line 76: Can we check for 9 and greater? So that we don't need more changes when v10 is released

src/nvidia-cuda/install.sh
```diff
@@ -81,7 +91,13 @@ fi
 
 if [ "$INSTALL_CUDNNDEV" = "true" ]; then
     # Ensure that the requested version of cuDNN development package is available AND compatible
-    cudnn_dev_pkg_version="libcudnn8-dev=${CUDNN_VERSION}-1+cuda${CUDA_VERSION}"
+    if [[ $major_cudnn_version = "9" ]]
```
samruddhikhandale, line 94: Same comment here.

samruddhikhandale (changes requested):
I have left minor comments, besides that looks wonderful.

src/nvidia-cuda/install.sh
prathameshzarkar9: done

src/nvidia-cuda/install.sh
prathameshzarkar9: Done

samruddhikhandale (approved):
Wonderful, thanks!

Issue #1072, opened 2024-08-01
bug(git-lfs): Can't check signature: No public key


CaffeineDaemon:
When building our devcontainer on our Gitlab-CI build pipeline (to use it in ci-jobs) with the following command

`devcontainer build . --image-name $CI_REGISTRY_IMAGE/devcontainer:$CI_COMMIT_SHA --workspace-folder .  --cache-from="type=local,src=.buildx/cache" --cache-to="type=local,dest=.buildx/cache" --cache-to="type=registry,ref=reg.goibykus.de/stp/devcontainer:cache,mode=max,image-manifest=true"`

the git-lfs feature fails installation with the following log:
```
#18 0.145 ===========================================================================
#18 0.145 Feature       : Git Large File Support (LFS)
#18 0.145 Description   : Installs Git Large File Support (Git LFS) along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like git and curl.
#18 0.145 Id            : ghcr.io/devcontainers/features/git-lfs
[2024-08-01T07:30:38.548Z] 
#18 0.145 Version       : 1.2.2
#18 0.145 Documentation : https://github.com/devcontainers/features/tree/main/src/git-lfs
#18 0.145 Options       :
#18 0.145     VERSION="latest"
#18 0.145     AUTOPULL="true"
#18 0.145     INSTALLDIRECTLYFROMGITHUBRELEASE="false"
#18 0.145 ===========================================================================
[2024-08-01T07:30:38.741Z] #18 0.184 Installing Git LFS...
#18 0.188 (*) No apt package for bookworm amd64. Installing manually.
[2024-08-01T07:30:38.961Z] #18 0.558 GIT_LFS_VERSION=3.5.1
#18 0.558 Looking for release artfact: git-lfs-linux-amd64-v3.5.1.tar.gz
[2024-08-01T07:30:45.146Z] #18 6.743 (*) Keyserver hkp://keyserver.ubuntu.com is not reachable.
[2024-08-01T07:30:50.164Z] #18 11.76 (*) Keyserver hkp://keyserver.pgp.com is not reachable.
[2024-08-01T07:30:51.203Z] #18 12.80 (*) Downloading GPG key...
[2024-08-01T07:30:51.364Z] #18 12.81 gpg: keybox '/tmp/tmp-gnupg/pubring.kbx' created
[...]

PR #1124, opened 2024-09-19
[Git-Lfs] - Solution to #1072 issue - Can't download public key

gauravsaini04:
Ref: [#1072](https://github.com/devcontainers/features/issues/1072)

**Feature Name**
* Git - Lfs

**Description**
* As a solution to Features issue #1072 
* Similar changes as in #1056 

**_Changelog_**
* Added similar changes as in #1056 to `install.sh` file's `get_gpg_key_servers` function

**Checklist**

### DIFF
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -62,13 +62,6 @@ find_version_from_git_tags() {
 
 # Get the list of GPG key servers that are reachable
 get_gpg_key_servers() {
-    declare -A keyservers_curl_map=(
-        ["hkp://keyserver.ubuntu.com"]="http://keyserver.ubuntu.com:11371"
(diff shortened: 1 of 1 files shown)


src/git-lfs/install.sh
```diff
@@ -62,29 +62,33 @@ find_version_from_git_tags() {
 
 # Get the list of GPG key servers that are reachable
 get_gpg_key_servers() {
```
samruddhikhandale, line 64: Can we bump patch version for the Feature?

samruddhikhandale (changes requested):

src/git-lfs/install.sh
gauravsaini04: made changes

samruddhikhandale (approved):

PR #1125, opened 2024-09-19
Align how http proxies are handled when fetching gpg keys for download verification

rbambrough-intel:
I noticed while trying to run a dev-container setup behind a corporate proxy that git-lfs was getting stuck waiting for gpg keys.

It downloads the collateral file from, mostly, github and then checks to make sure the hashes match up with what is expected. Due to how the proxy was being handled through the install scripts it would either failure to download the collateral (http proxy not correctly setup in container environment) or would fail to verify the collateral (http proxy setup correctly, but gpg not being passed proxy values).

To solve this I went through and aligned all of the features which use gpg checking with how the python feature does it.

I also removed the httpProxy parameter from the features which had it as anyone using it would have already had to have had the proxy correctly setup in order to download the collateral initially. I was one the fence about this change and would remove it if it is deemed no backwards compatible. I didn't see any documentation on how this type of removal should be handled.

Affected features:

git-lfs
git
github-cli
kubectl-helm-minikube
python
ruby
terraform

### DIFF
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -70,16 +70,11 @@ get_gpg_key_servers() {
         ["hkp://keyserver.pgp.com"]="http://keyserver.pgp.com:11371"
     )
 
-    local curl_args=""
     local keyserver_reachable=false  # Flag to indicate if any keyserver is reachable
-
-    if [ ! -z "${KEYSERVER_PROXY}" ]; then
-        curl_args="--proxy ${KEYSERVER_PROXY}"
-    fi
 
     for keyserver in "${!keyservers_curl_map[@]}"; do
         local keyserver_curl_url="${keyservers_curl_map[${keyserver}]}"
-        if curl -s ${curl_args} --max-time 5 ${keyserver_curl_url} > /dev/null; then
+        if curl -s --max-time 5 ${keyserver_curl_url} > /dev/null; then
             echo "keyserver ${keyserver}"
             keyserver_reachable=true
         else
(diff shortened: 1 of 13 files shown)


src/git-lfs/install.sh
```diff
@@ -15,6 +15,7 @@ GIT_LFS_ARCHIVE_GPG_KEY_URI="https://packagecloud.io/github/git-lfs/gpgkey"
 GIT_LFS_ARCHIVE_ARCHITECTURES="amd64 arm64"
 GIT_LFS_ARCHIVE_VERSION_CODENAMES="stretch buster bullseye bionic focal jammy"
 GIT_LFS_CHECKSUM_GPG_KEYS="0x88ace9b29196305ba9947552f1ba225c0223b187 0x86cd3297749375bcf8206715f54fe648088335a9 0xaa3b3450295830d2de6db90caba67be5a5795889"
+KEYSERVER_PROXY="${HTTP_PROXY:-""}"
```
samruddhikhandale, line 18: @rbambrough-intel Thank you so much for taking the time to contribute this PR.

Instead of these changes, how about we making similar changes https://github.com/devcontainers/features/pull/1119 ?

samruddhikhandale (changes requested):

Jarsop:
Thanks for your PR, I came to the same conclusion and it avoids duplicate work.

src/git-lfs/install.sh
Jarsop: It seems the link is not pointing to the right PR and I don't understand your statement as it seems to be the common way to define a default variable value in this project (e.g.: https://github.com/devcontainers/features/blob/0be79410981d50c2b5fa35e97f4af8a1a4f4727d/src/ruby/install.sh#L18, https://github.com/devcontainers/features/blob/302fecab063c63a4f3b6de1bbba408f4b896f7cd/src/java/install.sh#L29, https://github.com/devcontainers/features/blob/302fecab063c63a4f3b6de1bbba408f4b896f7cd/src/powershell/install.sh#L16). However I think this following will be enough:

```suggestion
KEYSERVER_PROXY="${HTTP_PROXY}"
 ```

 But maybe I missed something.

Comment (samruddhikhandale) [reply src/git-lfs/install.sh:18]: Apologies, I pasted the incorrect link.

I was recommending https://github.com/devcontainers/features/pull/1124 which tests the servers before using them to download the keys.

> KEYSERVER_PROXY="${HTTP_PROXY:-""}"

This simply assigns the `KEYSERVER_PROXY` to "" if `HTTP_PROXY` is not defined.

Comment (rbambrough-intel) [reply src/git-lfs/install.sh:18]: @Jarsop If we are fine with removing the httpProxy setting from some of the features then it can be simplified into:

```bash
KEYSERVER_PROXY="$HTTP_PROXY"
```

rbambrough-intel: @samruddhikhandale It does do curl checking of the key servers, just not the same as the one you linked (map instead of multiple function calls). I wasn't sure touching the function one so I adjusted all the others that looked closee

Issue #718, opened 2023-10-09
 Version '8.9.5.29-1+cuda12.1' for 'libcudnn8' was not found


FlatMapIO:
It looks like the cudnn install script needs to be changed to accommodate different version combinations.  

Here is the discussion: https://forums.developer.nvidia.com/t/e-version-8-3-1-22-1-cuda10-2-for-libcudnn8-was-not-found/200801/30 

config file:

```json
{
  "$schema": "https://raw.githubusercontent.com/devcontainers/spec/main/schemas/devContainer.base.schema.json",
  "name": "pytorch-devcontainer",
  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
  "features": {
    "ghcr.io/devcontainers/features/nvidia-cuda:1": {
      "cudaVersion": "12.1",


      "installCudnn": true,
      // https://developer.nvidia.com/rdp/cudnn-download
      "cudnnVersion": "8.9.5.29"
    },
}
```

FlatMapIO:
```
DISTRO=ubuntu2204
ARCH=x86_64

CUDA_VERSION=${CUDAVERSION:-12.1}

CUDNN_VERSION=$(apt-cache policy libcudnn8 | grep "$CUDA_VERSION" | grep -Eo '^[^-1+]*' | sort -V | tail -n1)
```

samruddhikhandale:
Hi 👋 

Currently, the nvidia-cuda Feature expects `cudaVersion` and `cudnnVersion` to install tools correctly (ensure compatabolity). However, I agree that we should update it to automatically find a compatible version. 

@FlatMapIO would you be interested in contributing a PR with the enhancement changes? We'd appreciate that, thanks!

prathameshzarkar9:
Hi @samruddhikhandale ,
Created PR https://github.com/devcontainers/features/pull/1126

PR #1126, opened 2024-09-23
auto find cudnn version

prathameshzarkar9:
[Issue 718](https://github.com/devcontainers/features/issues/718)
Version '8.9.5.29-1+cuda12.1' for 'libcudnn8' was not found

**Feature Name**
- nvidia-cuda

**Changelog**
- Changes to install.sh for features issue 718 to autofind the cudnn version if cuda version is below 12.3
- Changes to scenario.json to test the older version of cuda is installing a latest cudnn version automatically

**Checklist**
- [x] if pass version as 12.2 and below then it is finding the cudnn version


src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -45,32 +45,16 @@
...
+        "9.1.1.17",
+        "9.2.0.82",
+        "9.2.1.18",
+        "9.3.0.75",
+        "9.4.0.58"
       ],
-      "default": "8.6.0.163",
-      "description": "Version of cuDNN to install"
+      "default": "9.4.0.58",
+      "description": "Version of cuDNN to install if Cuda Version is greater than or equal to 12.3"
```
samruddhikhandale, line 57: This is little confusing. Does this mean it will work only if Cuda version is less than 12.3? 

> auto searching of latest cudnn version and package works only for libcudnn major version 8. If we try to search the libcudnn9 then it is not found hence we have to pass the cudnn version if cuda version is greater than or equal to 12.3

As folks prefer to use the latest of every tooling, having the automated script work only for an older version won't really be helpful.  

Can you explain why we cannot have automations for libcudnn9?

samruddhikhandale (changes requested):

src/nvidia-cuda/devcontainer-feature.json
prathameshzarkar9: yeah, you are right. I found something, few packages are not able to being located using apt-cache. So I tried to update the apt packages using `apt-get update` after installing `cuda-libraries`. Now we are able to filter all the packages available at url : `https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/`
Accordingly deprecated `cudnnVersion` and auto finding compatible cudnn version for all the cuda versions. Kindly find all the test scenarios for new versions 12.3, 12.4, 12.5

src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -44,43 +44,6 @@
       ],
       "default": "11.8",
       "description": "Version of CUDA to install"
-    },
-    "cudnnVersion": {
```
samruddhikhandale, line 48: We should not delete the Feature option, we should still let user pin to a specific version if they'd like.

Instead, we could update `default` to `automatic`

src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "nvidia-cuda",
-  "version": "1.1.3",
+  "version": "1.1.4",
```
samruddhikhandale, line 3:
```suggestion
  "version": "1.2.0",
```

samruddhikhandale (changes requested):

src/nvidia-cuda/devcontainer-feature.json
prathameshzarkar9: Done

src/nvidia-cuda/devcontainer-feature.json
prathameshzarkar9: done

src/nvidia-cuda/devcontainer-feature.json
```diff
@@ -77,9 +77,10 @@
         "9.2.0.82",
         "9.2.1.18",
         "9.3.0.75",
-        "9.4.0.58"
+        "9.4.0.58",
+        "automatic"
```
samruddhikhandale, line 81: Let's move this value at the top of the list

samruddhikhandale (approved):
Looks great, left a minor comment

src/nvidia-cuda/devcontainer-feature.json
prathameshzarkar9: Done

test/nvidia-cuda/scenarios.json
```diff
@@ -16,47 +16,40 @@
...
+    "install_cuda_12_5_version": {
         "image": "debian",
         "features": {
             "nvidia-cuda": {
                 "installCudnn": true,
-                "installNvtx": true,
                 "installCudnnDev": true,
-                "installToolkit": true,
-                "cudaVersion": "12.5",
-                "cudnnVersion": "9.4.0.58"
```
samruddhikhandale, line 59: Can we keep few scenarios to validate `cudnnVersion` where it is NOT automatic?

samruddhikhandale (changes requested):

test/nvidia-cuda/scenarios.json
prathameshzarkar9: Done

eljog (approved):

PR #1127, opened 2024-09-23
feat(nix): Enable persistent volume for shared nix-store

### LINKED ISSUE #284: Nix feature doesn't cache /nix
Environment setup after container restarts can take a really long time if all of nixos.org has to be downloaded every time. I guess you might want the option to clear the cache or switch it off, but IMO it would be better if /nix was a volume mount by default.

max06:
Closes #284

Enables a shared, persistent docker volume for /nix/store. 
Garbage collection is no issue since nix simply refetches everything that's missing. On the other hand this volume will probably save a lot of storage and bandwidth due to deduplication.

My use case: I enabled the nix feature per default for every devcontainer, to be used with my dotfiles-repo (and home-manager). 

Alternative: Duplicate feature to have both options, one with cache, one without.

-----

**Update**

The nix-store volumes are no longer shared with all containers using this feature. Nix (or it's store) does not seem to be ready for shared access. Instead, every devcontainer using this feature gets its own persistent store volume. This reduces the rebuild-times and data transfer, but not as much as a the shared store. Imo, still a useful change.

The diff looks bad, but it's just because I removed one if-else-condition spanning 55 lines. 

(Wrong statement about renamed package removed)

### DIFF
--- a/src/nix/devcontainer-feature.json
+++ b/src/nix/devcontainer-feature.json
@@ -46,7 +46,7 @@
     "mounts": [
         { 
             "source": "devcontainer-nix-store", 
-            "target": "/nix/store", 
+            "target": "/nix", 
             "type": "volume" 
         }
     ],
--- a/src/nix/devcontainer-feature.json
+++ b/src/nix/devcontainer-feature.json
@@ -43,5 +43,12 @@
     "containerEnv": {
         "PATH": "/nix/var/nix/profiles/default/bin:/nix/var/nix/profiles/default/sbin:${PATH}"
     },
+    "mounts": [
+        { 
+            "source": "devcontainer-nix-store", 
+            "target": "/nix/store", 
+            "type": "volume" 
+        }
+    ],
     "entrypoint": "/usr/local/share/nix-entrypoint.sh"
 }
--- a/src/nix/devcontainer-feature.json
+++ b/src/nix/devcontainer-feature.json
@@ -7,7 +7,10 @@
     "options": {
         "version": {
             "type": "string",
-            "proposals": ["latest", "2.11"],
+            "proposals": [
+                "latest",
+                "2.11"
+            ],
             "default": "latest",
             "description": "Version of Nix to install."
         },
@@ -44,10 +47,10 @@
         "PATH": "/nix/var/nix/profiles/default/bin:/nix/var/nix/profiles/default/sbin:${PATH}"
     },
     "mounts": [
-        { 
-            "source": "devcontainer-nix-store", 
-            "target": "/nix", 
-            "type": "volume" 
+        {
+            "source": "${devcontainerId}-nix-store",
+            "target": "/nix",
+            "type": "volume"
         }
     ],
     "entrypoint": "/usr/local/share/nix-entrypoint.sh"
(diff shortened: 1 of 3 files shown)


src/nix/devcontainer-feature.json
```diff
@@ -43,5 +43,12 @@
     "containerEnv": {
         "PATH": "/nix/var/nix/profiles/default/bin:/nix/var/nix/profiles/default/sbin:${PATH}"
     },
+    "mounts": [
```
samruddhikhandale, line 46: Looks like this is causing multiple test failures, https://github.com/devcontainers/features/actions/runs/10992584155/job/30536518797?pr=1127 

Can you take a look? thanks!

samruddhikhandale (changes requested):

src/nix/devcontainer-feature.json
max06: I think I mounted too deep, `/nix/store` instead of `/nix`. Let's try that again.

samruddhikhandale: Not helping 😞 see https://github.com/devcontainers/features/actions/runs/11014623966/job/30596929100?pr=1127

max06: Alright - sorry for keeping you busy. I'll dig deeper locally.

max06: New approach: Persistent nix-stores per devcontainer, not shared. It looks like installing nix as single-user isn't very happy about existing single-user installations and breaks somewhere™️. To avoid potential bigger issues I'm no longer sharing that volume with every container using the feature. 

My local tests are still running, but it looks promising so far.

max06:
All tests green - on my machine. Let's hope CI gets the same result 😄

src/nix/devcontainer-feature.json
```diff
@@ -7,7 +7,10 @@
     "options": {
         "version": {
```
samruddhikhandale, line 8: Can we bump the minor version? Thanks! 

https://github.com/devcontainers/features/blob/260aeae1c8b1aa0a6da18aa96b924b92d6de7743/src/nix/devcontainer-feature.json#L3

src/nix/devcontainer-feature.json
samruddhikhandale: Looks like it worked 🎉

src/nix/devcontainer-feature.json
```diff
@@ -43,5 +46,12 @@
     "containerEnv": {
         "PATH": "/nix/var/nix/profiles/default/bin:/nix/var/nix/profiles/default/sbin:${PATH}"
     },
+    "mounts": [
+        {
+            "source": "${devcontainerId}-nix-store",
```
samruddhikhandale, line 51:
```suggestion
            "source": "nix-store-${devcontainerId}",
```

samruddhikhandale (changes requested):

src/nix/devcontainer-feature.json
max06: As you wish 😺

samruddhikhandale (approved):
Looks great, thank you! ⚡

PR #1144, opened 2024-10-07
[git] Git subtree install option

tsuga:
This PR enhances the existing git feature to install git/contrib/subtree by specifying to do so in the options.
```
    "options": {
        "version": {
            "type": "string",
            "proposals": [
                "latest",
                "system",
                "os-provided"
            ],
            "default": "os-provided",
            "description": "Select or enter a Git version."
        },
        "ppa": {
            "type": "boolean",
            "default": true,
            "description": "Install from PPA if available (only supported for Ubuntu distributions)"
        },
        "installSubtree": {
            "type": "boolean",
            "default": false,
            "description": "Install git/contrib/subtree"
        }
    },
```

### DIFF
--- a/src/git/devcontainer-feature.json
+++ b/src/git/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "git",
-    "version": "1.3.2",
+    "version": "1.3.3",
     "name": "Git (from source)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/git",
     "description": "Install an up-to-date version of Git, built from source as needed. Useful for when you want the latest and greatest features. Auto-detects latest stable version and installs needed dependencies.",
--- a/src/git/devcontainer-feature.json
+++ b/src/git/devcontainer-feature.json
@@ -19,6 +19,11 @@
             "type": "boolean",
             "default": true,
             "description": "Install from PPA if available (only supported for Ubuntu distributions)"
+        },
+        "installSubtree": {
+            "type": "boolean",
+            "default": false,
+            "description": "Install git/contrib/subtree"
         }
     },
     "installsAfter": [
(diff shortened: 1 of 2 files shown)


tsuga:

samruddhikhandale:
Hi @tsuga 👋 

Thank you for taking the time to contribute this PR. Can you help me understand how you utilize `git/contrib/subtree` and how impactful this change will be for the community?

I understand that the `git-subtree` command is an alternative to `git-submodule` for managing project dependencies. However, we'd like more feedback before we add this feature option. Hope that makes sense to you. Thanks!

// cc @bamurtaugh

tsuga:
Hi @samruddhikhandale 
There are various pros and cons over subtree vs submodule.
You may find good explanation on the Internet.

Karmaibringkaosanddrama:
> This PR enhances the existing git feature to install git/contrib/subtree by specifying to do so in the options.
> ```
>     "options": {
>         "version": {
>             "type": "string",
>             "proposals": [
>                 "latest",
>                 "system",
>                 "os-provided"
>             ],
>             "default": "os-provided",
>             "description": "Select or enter a Git version."
>         },
>         "ppa": {
>             "type": "boolean",
>             "default": true,
>             "description": "Install from PPA if available (only supported for Ubuntu distributions)"
>         },
>         "installSubtree": {
>             "type": "boolean",
>             "default": false,
>             "description": "Install git/contrib/subtree"
>         }
>     },
> ```

@

Kaniska244:
Hello @tsuga ,

Thank you for the PR. Would you kindly add some tests to verify the git subtree feature when opted in by the user using the flag. A simple "git --version" like command isn't available for git subtree as I can see.

With Regards,
Kaniska

Kaniska244:
> @Kaniska244 Thank you for your attention to this PR! Indeed, `git subtree --version` does not exist. However, if subtree is installed, `git subtree --help` should return the help text. If not installed, `git subtree --help` will raise an error `git: 'subtree' is not a git command.`
> 
> Do you think `git subtree --help` is enough? If you can agree, I will add the line as a test.

Hello @tsuga ,

Thank you for the response. Sure if git subtree -h or git subtree --help works, it could be used in the test scripts.

With Regards,

PR #1146, opened 2024-10-08
Add support Enterprise Linux for feature kubectl-helm-minikube

xrow:
I was not able to test for debian based repos. I hope it works out of the box.

### DIFF
--- a/src/kubectl-helm-minikube/install.sh
+++ b/src/kubectl-helm-minikube/install.sh
@@ -125,30 +125,99 @@ find_prev_version_from_git_tags() {
     set -e
 }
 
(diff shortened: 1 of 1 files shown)


xrow:

Kaniska244:
Hi @xrow ,

Thanks for the PR. Would it be possible for you to add tests all of these linux distros (such as alpine, fedora, rhel, mariner & debian) to which support for installation of kubetnetes-hulm-minikube feature is supposed to be extended as part of this PR.

Kindly let me know in case of any concern/ clarifications.

With Regards,
Kaniska

PR #1151, opened 2024-10-16
Create azure-container-webapp.yml

ghost:


bamurtaugh:
Thanks for the PR! Could you please provide some more info on the scenario it'd support?

Issue #1026, opened 2024-07-01
Common-utils feature should include shell history


pickfire:
```json
    "ghcr.io/stuartleeks/dev-container-features/shell-history:0": {}
```

I need to rely on this to get shell history or had to do manual intervention, I expect this from `common-utils`.

PR #1157, opened 2024-10-22
[common-utils] - append shell history functionality - #1026

gauravsaini04:
Ref: #1026 

**Feature**
* Common-Utils

**Description**
* In response to #1026 have appended shell history functionality to `common-utils` feature as a configurable option

**_changelog_**
* Have added changes to `devcontainer-feature.json` file, added boolean option `allowShellHistory` and mounts as a configurable option
* Have added changes to `main.sh` in `common-utils` features src 
* Have added to `scenarios` to test this new feature
* Have added test cases in `allow_shell_history.sh` file

**Checklist**
- [x] checked that applied changes work as expected

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -64,6 +64,18 @@
             "type": "boolean",
             "default": false,
             "description": "Add packages from non-free Debian repository? (Debian only)"
+        },
+        "allowShellHistory": {
+            "type": "boolean",
+            "default": false,
+            "description": "Preserve shell history across dev container instances? (Currently supports bash, zsh, and fish)"
         }
-    }
+    },
+    "mounts": [
+        {
+            "source": "${devcontainerId}-shellhistory",
+            "target": "/dc/shellhistory",
+            "type": "volume"
+        }
+    ]
 }
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -18,6 +18,7 @@ USERNAME="${USERNAME:-"automatic"}"
 USER_UID="${USERUID:-"automatic"}"
 USER_GID="${USERGID:-"automatic"}"
 ADD_NON_FREE_PACKAGES="${NONFREEPACKAGES:-"false"}"
+ALLOW_SHELL_HISTORY="${ALLOWSHELLHISTORY:-"false"}"
 
 MARKER_FILE="/usr/local/etc/vscode-dev-containers/common"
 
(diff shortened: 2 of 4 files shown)


eitsupi (commented):
Some questions:

- Wouldn't adding more functionality to common-utils here go against this discussion? #67
- Why not use lifecycle-scripts? See stuartleeks/dev-container-features#16

src/common-utils/devcontainer-feature.json
```diff
@@ -64,6 +64,18 @@
...
+        },
+        "allowShellHistory": {
+            "type": "boolean",
+            "default": false,
+            "description": "Preserve shell history across dev container instances? (Currently supports bash, zsh, and fish)"
+        }
+    },
+    "mounts": [
+        {
+            "source": "${devcontainerId}-shellhistory",
```
chrmarti, line 76: This will add a volume for each container which will make volumes harder to manage for the user. Maybe you could use a single volume `devcontainers` at `/devcontainers` and in that create a folder `/devcontainers/shellHistory/${devcontainerId}` for each dev container.

src/common-utils/main.sh
```diff
@@ -567,6 +568,55 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
+# On shells where this isn't possible, we symlink as a fallback.
+#
+
+if [ "${ALLOW_SHELL_HISTORY}" = "true" ]; then
+# Set HISTFILE for bash
+cat << EOF >> "${user_home}/.bashrc"
+if [[ -z "\$HISTFILE_OLD" ]]; then
+    export HISTFILE_OLD=\$HISTFILE
+fi
+export HISTFILE=/dc/shellhistory/.bash_history
```
chrmarti, line 587: Would there be value in keeping the default path `~/.bash_history` and instead make that a symbolic link to the actual file in the volume? (I haven't tried using a symbolic link, best to check first if bash/zsh work correctly with that.)

chrmarti (changes requested):

src/common-utils/devcontainer-feature.json
gauravsaini04: have added logic such that uuidgen utility is used for generating random devcontainer id for a container, intent is to embed it inside the folder structure like this :
/devcontainers/$(uuidgen)/shellHistory for a symlink to the ~/.bash_history being kept here. Logic is working for retention of shell history, keeping it as draft so you can check for bash shell at the moment, similarly can be implemented for zsh as well.

eitsupi (changes requested):
Why add this functionality to common-utils instead of relying on another Feature?
Wouldn't common-utils be too heavy a dependency for users who only want to use history?

And I am also convinced that the lifecycle script approach is far superior when implementing such a functionality.
Please check  stuartleeks/dev-container-features#16

src/common-utils/main.sh
gauravsaini04: done like this. thanks.

gauravsaini04:
> And I am also convinced that the lifecycle script approach is far superior when implementing such a functionality. 

Thanks for valuable inputs @eitsupi, waiting for the logic to be approved by @chrmarti, will be converting to a lifecycle script after this

src/common-utils/main.sh
```diff
@@ -567,6 +568,53 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
+    if [[ ! -f "${VOLUME_HISTORY_FILE}" ]]; then
+        # Create an empty history file if it doesn’t already exist
+        sudo touch "${VOLUME_HISTORY_FILE}"
+        sudo chown -R "${USERNAME}" "${VOLUME_HISTORY_FILE}"
+        sudo chmod -R u+rwx "${VOLUME_HISTORY_FILE}"
+    fi
+
+    # Create or update the user’s .bash_history to append to the volume’s history
+    if [[ ! -f "${USER_HISTORY_FILE}" ]]; then
+        sudo touch "${USER_HISTORY_FILE}"
```
chrmarti, line 604: This will be overwritten by the symbolic link below. I think this can be removed.

chrmarti (changes requested):

chrmarti:
> Why add this functionality to common-utils instead of relying on another Feature? Wouldn't common-utils be too heavy a dependency for users who only want to use history?

I agree, I'm also not sure if adding a `devcontainers` volume for all users of our base images will be ideal. Though having persistence for shell history built-in would be a great feature.

> And I am also convinced that the lifecycle script approach is far superior when implementing such a functionality. Please check [stuartleeks/dev-container-features#16](https://github.com/stuartleeks/dev-container-features/pull/16)

This seems to require password-less `sudo` access. The feature would not work without that.

src/common-utils/main.sh
```diff
@@ -567,6 +568,53 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
+if [ "${ALLOW_SHELL_HISTORY}" = "true" ]; then
+    echo "Activating feature 'shell-history'"
+    echo "User: ${USERNAME}     User home: ${user_home}"
+
+    if ! command -v uuidgen &> /dev/null; then
+        sudo apt-get update
+        sudo apt-get install -y uuid-runtime
+    fi
+    # Create the shell history directory in the mounted volume
+    DEVCONTAINER_ID=$(uuidgen)
```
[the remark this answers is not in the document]
gauravsaini04: made changes as suggested.

src/common-utils/main.sh
```diff
@@ -567,6 +568,53 @@ if [ "${INSTALL_ZSH}" = "true" ]; then
...
+
+    # Create the history directory in the volume, if it doesn’t already exist
+    sudo mkdir -p "${HISTORY_DIR}"
+    sudo chown -R "${USERNAME}" "${HISTORY_DIR}"
+    sudo chmod -R u+rwx "${HISTORY_DIR}"
+
+    # Ensure the volume's history file exists and set permissions
+    if [[ ! -f "${VOLUME_HISTORY_FILE}" ]]; then
+        # Create an empty history file if it doesn’t already exist
+        sudo touch "${VOLUME_HISTORY_FILE}"
```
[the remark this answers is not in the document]
gauravsaini04: removed the check

src/common-utils/main.sh
gauravsaini04: ok

eitsupi:
> This seems to require password-less `sudo` access. The feature would not work without that.

Isn't that irrelevant whether or not lifecycle scripts are used? We don't know the permissions of a mounted volume until it is mounted, do we?

chrmarti:
One issue with the change is that it unconditionally adds mount points independent of whether preserving the shell history is enabled or not. Given that the `common-utils` are in all our base images, this would affect a lot of dev containers. Possible approaches to avoid this are:
- Add support for conditional mount points for features in the spec and CLI.
- Move the functionality to a separate feature. Users will have to add it explicitly.
- Make support for preserving the shell history part of the CLI. (This could allow us to reuse the `vscode` volume because the Dev Containers extension could tell the CLI to use that. It might also allow us to enable it by default with a setting in the Dev Containers extension.)

I suggest we put this on hold for now.

PR #1159, opened 2024-10-25
[common-utils] Fix group creation if a group with the given id already it exists

x241c297f:
On my computer I have user id 501 and group id 20 which I want to preserve within the container.

However, that group id often already exists in a lot of containers.
Container creation fails when using common-utils and setting the `userGid` to the id of an already existing group, because the current implementation just tries to add the group right away.

This PR adds functionality to detect whether a group with the given id already exists and handles it accordingly.


x241c297f:

PR #1160, opened 2024-10-25
[common-utils] Fixes issue installing libssl3 on debian unstable

pxslip:
Debian unstable has `libssl3t64` installed, which conflicts with `libssl3` and causes image creation to fail.

Added a test to pick libssl3t64 if available, falls back to libssl3 if not.

Some thoughts:

- Nesting `if`s inside of the `^libssl3` test is unnecessary, but felt good for grouping the selection logic together.
- The `elif` fallback to `libssl3` is likewise overly verbose, but I included it for readability. Changing to `else` should have the same outcome.

### DIFF
--- a/src/common-utils/main.sh
+++ b/src/common-utils/main.sh
@@ -81,8 +81,13 @@ install_debian_packages() {
         fi
 
         # Include libssl3 if available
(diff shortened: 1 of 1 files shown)


pxslip:

PR #1178, opened 2024-11-12
Update devcontainer-feature.json

haliliceylan:
Since a new Node version has been released, LTS is now automatically installing version 22. This is causing the dev containers in my work environment to crash (project support only 20). We should support selecting version 20 as the default. In fact, I think it should be added in the latest version (22) to avoid such problems in the future.


itssimon:
I'm hitting this issue and would like to see this merged

Issue #885, opened 2024-02-27
Oryx: Unpin .NET version 8.0.101


samruddhikhandale:
https://github.com/devcontainers/features/pull/882 pins the .NET version to unblock oryx build failures. 

Once the oryx repo updates their dependencies, start installing `latest` version of .NET using `apt` install.

prathameshzarkar9:
Hi @samruddhikhandale ,
Created a PR for this change here: https://github.com/devcontainers/features/pull/991
As suggested, the PR is in draft mode and will be addressed once all the dependent packages are up-to-date for this dotnet migration

samruddhikhandale:
@prathameshzarkar9 Any update here?

Mathiyarasy:
Unpinning the version caused issue. 
Pinned dotnet version is 8.0.202. 
PR: https://github.com/devcontainers/features/pull/1181.

PR #1181, opened 2024-11-14
Oryx Issue

Mathiyarasy:
Universal Image Latest Version: Caused issue in a Pipeline 
[Azure DevOps Pipeline generate-kitchensink-automated Failed](https://github.com/github/codespaces/issues/19823)

This PR aims to do things:
1.  The pinned dotnet version now is 8.0.202 : [Oryx: Unpin .NET version 8.0.101](https://github.com/devcontainers/features/issues/885 )
2. Publish oryx app with **--self-contained true** tag . Without this tag oryx fails to launch stating compatible .net runtime version not found

### DIFF
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -186,7 +186,7 @@ cd $GIT_ORYX
 ${DOTNET_BINARY} build "$SOLUTION_FILE_NAME" -c Debug
 
 ${DOTNET_BINARY} publish -property:ValidateExecutableReferencesMatchSelfContained=false -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj --self-contained true
-${DOTNET_BINARY} publish -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildServer/BuildServer.csproj
+${DOTNET_BINARY} publish -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildServer/BuildServer.csproj --self-contained true
 
 chmod a+x ${BUILD_SCRIPT_GENERATOR}/GenerateBuildScript
 
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -235,7 +235,11 @@ if [[ "${PINNED_SDK_VERSION}" != "" ]]; then
     rm -f ${GIT_ORYX}/global.json
     rm -rf /usr/share/dotnet/sdk/$PINNED_SDK_VERSION
     NEW_RUNTIME_VERSIONS=$(dotnet --list-runtimes | awk '{print $2}' | sort | uniq)
-    SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS" | grep -vxFf <(echo "$RUNTIME_VERSIONS"))
+    if [ -n "${RUNTIME_VERSIONS:-}" ]; then
+        SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS" | grep -vxFf <(echo "$RUNTIME_VERSIONS"))
+    else
+        SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS")
+    fi
     rm -rf /usr/share/dotnet/shared/Microsoft.NETCore.App/$SDK_INSTALLED_RUNTIME
     rm -rf /usr/share/dotnet/shared/Microsoft.AspNetCore.App/$SDK_INSTALLED_RUNTIME
     rm -rf /usr/share/dotnet/templates/$SDK_INSTALLED_RUNTIME
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -158,19 +158,11 @@ PINNED_SDK_VERSION=""
 # Oryx needs to be built with .NET 8
 if [[ "${DOTNET_BINARY}" = "" ]] || [[ $MAJOR_VERSION_ID != "8" ]] ; then
     echo "'dotnet 8' was not detected. Attempting to install .NET 8 to build oryx."
-
-    # The oryx build fails with .Net 8.0.201, see https://github.com/devcontainers/images/issues/974
-    # Pinning it to a working version until the upstream Oryx repo updates the dependency
     install_dotnet_using_apt
-    # PINNED_SDK_VERSION="8.0.101"
-    # install_dotnet_with_script ${PINNED_SDK_VERSION}
-
     if ! dotnet --version > /dev/null ; then
         echo "(!) Please install Dotnet before installing Oryx"
         exit 1
     fi
-
-    DOTNET_BINARY="/usr/bin/dotnet"
 fi
 
 BUILD_SCRIPT_GENERATOR=/usr/local/buildscriptgen
--- a/src/oryx/install.sh
+++ b/src/oryx/install.sh
@@ -156,20 +156,21 @@ PATCH_VERSION_ID=$(echo $(dotnet --version) | cut -d . -f 3)
 
 PINNED_SDK_VERSION=""
 # Oryx needs to be built with .NET 8
-if [[ "${DOTNET_BINARY}" = "" ]] || [[ $MAJOR_VERSION_ID != "8" ]] || [[ $MAJOR_VERSION_ID = "8" && ${PATCH_VERSION_ID} -ge "101" ]] ; then
+if [[ "${DOTNET_BINARY}" = "" ]] || [[ $MAJOR_VERSION_ID != "8" ]] ; then
     echo "'dotnet 8' was not detected. Attempting to install .NET 8 to build oryx."
 
     # The oryx build fails with .Net 8.0.201, see https://github.com/devcontainers/images/issues/974
     # Pinning it to a working version until the upstream Oryx repo updates the dependency
-    # install_dotnet_using_apt
-    PINNED_SDK_VERSION="8.0.101"
-    install_dotnet_with_script ${PINNED_SDK_VERSION}
+    install_dotnet_using_apt
+    # PINNED_SDK_VERSION="8.0.101"
+    # install_dotnet_with_script ${PINNED_SDK_VERSION}
 
     if ! dotnet --version > /dev/null ; then
         echo "(!) Please install Dotnet before installing Oryx"
         exit 1
     fi
 
+    DOTNET_BINARY="/usr/bin/dotnet"
 fi
 
 BUILD_SCRIPT_GENERATOR=/usr/local/buildscriptgen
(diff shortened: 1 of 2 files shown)


src/oryx/install.sh
```diff
@@ -156,20 +156,21 @@ PATCH_VERSION_ID=$(echo $(dotnet --version) | cut -d . -f 3)
...
     echo "'dotnet 8' was not detected. Attempting to install .NET 8 to build oryx."
 
     # The oryx build fails with .Net 8.0.201, see https://github.com/devcontainers/images/issues/974
     # Pinning it to a working version until the upstream Oryx repo updates the dependency
-    # install_dotnet_using_apt
-    PINNED_SDK_VERSION="8.0.101"
-    install_dotnet_with_script ${PINNED_SDK_VERSION}
+    install_dotnet_using_apt
+    # PINNED_SDK_VERSION="8.0.101"
+    # install_dotnet_with_script ${PINNED_SDK_VERSION}
```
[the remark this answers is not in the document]
Mathiyarasy: Oryx build is not failing in latest dotnet versions. Removed the commented code.

src/oryx/install.sh
```diff
@@ -156,20 +156,21 @@ PATCH_VERSION_ID=$(echo $(dotnet --version) | cut -d . -f 3)
...
 
     if ! dotnet --version > /dev/null ; then
         echo "(!) Please install Dotnet before installing Oryx"
         exit 1
     fi
 
+    DOTNET_BINARY="/usr/bin/dotnet"
```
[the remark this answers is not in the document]
Mathiyarasy: Removed this too as this is set in  install_dotnet_using_apt function

src/oryx/install.sh
```diff
@@ -192,7 +193,7 @@ echo "Building solution '$SOLUTION_FILE_NAME'..."
 cd $GIT_ORYX
 ${DOTNET_BINARY} build "$SOLUTION_FILE_NAME" -c Debug
 
-${DOTNET_BINARY} publish -property:ValidateExecutableReferencesMatchSelfContained=false -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
+${DOTNET_BINARY} publish -property:ValidateExecutableReferencesMatchSelfContained=false -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj --self-contained true
```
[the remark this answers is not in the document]
Mathiyarasy: No oryx documentation to explain how to build dotnet project. 
In https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-publish documentation it is recommended to use this tag when using -r tag.
I could see similar code in Oryx:
https://github.com/microsoft/Oryx/blob/ca568d2aa65efd119ae0939ccc4352e2f92b13fb/images/build/build_buildimages.sh#L40 .

src/oryx/install.sh
```diff
@@ -192,8 +185,8 @@ echo "Building solution '$SOLUTION_FILE_NAME'..."
 cd $GIT_ORYX
 ${DOTNET_BINARY} build "$SOLUTION_FILE_NAME" -c Debug
 
-${DOTNET_BINARY} publish -property:ValidateExecutableReferencesMatchSelfContained=false -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
-${DOTNET_BINARY} publish -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildServer/BuildServer.csproj
+${DOTNET_BINARY} publish -property:ValidateExecutableReferencesMatchSelfContained=false -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj --self-contained true
+${DOTNET_BINARY} publish -r linux-x64 -o ${BUILD_SCRIPT_GENERATOR} -c Release $GIT_ORYX/src/BuildServer/BuildServer.csproj --self-contained true
```
Mathiyarasy, line 189: Added --self-contained tag in reference to Oryx: https://github.com/microsoft/Oryx/blob/ca568d2aa65efd119ae0939ccc4352e2f92b13fb/images/build/build_buildimages.sh#L41

src/oryx/install.sh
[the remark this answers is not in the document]
Mathiyarasy: Images were failing with oryx feature when installing using apt.
 base:noble image apt Installed dotnet version : 8.0.110 (Oryx build is failing with this version)
Modified the code to install dotnet using script: PINNED_SDK_VERSION="8.0.202". All test cases are passing

src/oryx/install.sh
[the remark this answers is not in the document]
Mathiyarasy: Yes There is a increase in feature size by 1 MB.
Before this change we had 103M now it is 104M

src/oryx/install.sh
```diff
@@ -156,20 +156,17 @@ PATCH_VERSION_ID=$(echo $(dotnet --version) | cut -d . -f 3)
 
 PINNED_SDK_VERSION=""
 # Oryx needs to be built with .NET 8
-if [[ "${DOTNET_BINARY}" = "" ]] || [[ $MAJOR_VERSION_ID != "8" ]] || [[ $MAJOR_VERSION_ID = "8" && ${PATCH_VERSION_ID} -ge "101" ]] ; then
+if [[ "${DOTNET_BINARY}" = "" ]] || [[ $MAJOR_VERSION_ID != "8" ]] || [[ $MAJOR_VERSION_ID = "8" && ${PATCH_VERSION_ID} -ne "202" ]] ; then
```
eljog, line 159: What about patch below and above 202?

src/oryx/install.sh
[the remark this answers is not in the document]
eljog: That's fair

eljog (dismissed):

src/oryx/install.sh
Mathiyarasy: **If Dotnet SDK PATCH version is below 202 Oryx build fails**: 
mcr.microsoft.com/devcontainers/base:noble image do not have dotnet installed.
When installed dotnet using apt it will install **dotnet 8.0.110** and Oryx build fails for this version.
**If Dotnet SDK PATCH version is 202 Oryx build works**: 
So to be consistent we are installing .net sdk 8.0.202 if this version is not in image
In existing code this installed sdk for Oryx is removed after oryx build completes. So it would not affect any other dotnet behavior of the image

Mathiyarasy:
Looked into the failing tests
Oryx do not support Jammy. Unsupported distribution version 'jammy'. is a valid failure scenario. The following 2 failures are valid ad expected
1. test (oryx, mcr.microsoft.com/devcontainers/base:ubuntu) 
2. test (ubuntu:jammy)

In test scenarios oryx **install_dotnet_and_oryx** fails at test case '**two versions of dotnet runtimes are present**'. 
Issue is because existing code is not removing dotnet runtime installed with SDK.
SDK: 8.0.202 installs Runtime: 8.0.3
But existing code tried to remove 8.0.2
Modified the code to identify the runtime installed with sdk in oryx . Test case is passing now.

src/oryx/install.sh
```diff
@@ -236,16 +234,19 @@ fi
...
-    # Extract the major, minor version and the first digit of the patch version
-    MAJOR_MINOR_PATCH1_VERSION=${PINNED_SDK_VERSION%??}
-    rm -rf /usr/share/dotnet/shared/Microsoft.NETCore.App/$MAJOR_MINOR_PATCH1_VERSION
-    rm -rf /usr/share/dotnet/shared/Microsoft.AspNetCore.App/$MAJOR_MINOR_PATCH1_VERSION
-    rm -rf /usr/share/dotnet/templates/$MAJOR_MINOR_PATCH1_VERSION
+    NEW_RUNTIME_VERSIONS=$(dotnet --list-runtimes | awk '{print $2}' | sort | uniq)
+    if [ -n "${RUNTIME_VERSIONS:-}" ]; then
+        SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS" | grep -vxFf <(echo "$RUNTIME_VERSIONS"))
+    else
+        SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS")
```
eljog, line 241: > SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS")

Why does the assignment require `echo`, not sure if I correctly follow the logic here

src/oryx/install.sh
```diff
@@ -236,16 +234,19 @@ fi
...
     rm -f ${GIT_ORYX}/global.json
     rm -rf /usr/share/dotnet/sdk/$PINNED_SDK_VERSION
-
-    # Extract the major, minor version and the first digit of the patch version
-    MAJOR_MINOR_PATCH1_VERSION=${PINNED_SDK_VERSION%??}
-    rm -rf /usr/share/dotnet/shared/Microsoft.NETCore.App/$MAJOR_MINOR_PATCH1_VERSION
-    rm -rf /usr/share/dotnet/shared/Microsoft.AspNetCore.App/$MAJOR_MINOR_PATCH1_VERSION
-    rm -rf /usr/share/dotnet/templates/$MAJOR_MINOR_PATCH1_VERSION
+    NEW_RUNTIME_VERSIONS=$(dotnet --list-runtimes | awk '{print $2}' | sort | uniq)
+    if [ -n "${RUNTIME_VERSIONS:-}" ]; then
```
Mathiyarasy, line 238: This RUNTIME_VERSIONS variable will be set only for images with dotnet even before oryx feature installation
So for images without DOTNET this variable will not be set .
This throws issue stating unbound variable.
"${RUNTIME_VERSIONS:-} This syntax will check RUNTIME_VERSIONS is unset or null if unset or null it will use an empty string as the value.

src/oryx/install.sh
Mathiyarasy: Removed echo from here. Explained the logic in the next comment

eljog (dismissed):

eljog:
We should update the feature version as well, right? @Mathiyarasy

src/oryx/install.sh
```diff
@@ -236,16 +234,19 @@ fi
...
+        SDK_INSTALLED_RUNTIME=$(echo "$NEW_RUNTIME_VERSIONS" | grep -vxFf <(echo "$RUNTIME_VERSIONS"))
+    else
+        SDK_INSTALLED_RUNTIME="$NEW_RUNTIME_VERSIONS"
```
eljog, line 241: Are there cases where more than one sdk is installed? I see plural `NEW_RUNTIME_VERSIONS`
In that case `SDK_INSTALLED_RUNTIME` may get assigned a multiline string? Wouldn't that case issues with deletions on the lines below?

Mathiyarasy: No we install only one sdk as a part of oryx and this is installing only one runtime
It is in plural because for images with dotnet even before oryx they sometime might have multiple runtimes from those runtimes we pick only the runtime installed from oryx feature.

eljog: In that case should the naming be updated to avoid the confusion?
I will go ahead and checkin this PR, you may followup with a PR if needed

PR #887, opened 2024-02-28
Adds /home/USER/.local/bin/ to PATH in /etc/sudoers.d/vscode,  

prabhakk-mw:
fixes devcontainers/features#870

Issue #870, opened 2024-02-20
Codespace unable to start in Jupyter using python feature with option "installJupyterLab": true 


prabhakk-mw:
Listed below is the `devcontainer.json` file that I am using to start a codespace on GitHub.

```JSON
{
  "name": "Only Jupyter",
  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
  "hostRequirements": {
    "cpus": 4
  },
  "features": {
    "ghcr.io/devcontainers/features/python": {
      "version": "3.10",
      "installJupyterlab": true,
      "configureJupyterlabAllowOrigin": "*"
    }
  },
  "containerUser": "vscode"
}
```

When attempting to open in `Jupyter`  as shown in the image below:
![image](https://github.com/devcontainers/features/assets/64955767/32d5d7aa-4526-4dc7-95de-6c5013fb8492)

I get the following error:
![image](https://github.com/devcontainers/features/assets/64955767/45c721b1-2c5d-4b21-86ad-ab3e5259773a)

The documentation for [opening-your-codespace-in-jupyterlab ](https://docs.github.com/en/codespaces/developing-in-a-codespace/getting-started-with-github-codespaces-for-machine-learning#opening-your-codespace-in-jupyterlab) seems to suggest that this should be all I need to do.

Restarting the codespace as the option suggests has no effect.

I can however, open in VSCode, and start `jupyter lab` from the terminal and access it in a separate browser tab.

Any help would be much appreciated!

prabhakk-mw:
Tagging for visibility: @jungaretti  @samruddhikhandale

samruddhikhandale:
Hi 👋 

Can you try with the following `.devcontainer.json`?

```json
{
  "name": "Only Jupyter",
  "image": "mcr.microsoft.com/devcontainers/universal:2",
  "hostRequirements": {
    "cpus": 4
  }
}
```

 I vaguely remember there needs another configuration change (when using Python Feature) which doesn't seem to be documented.

prabhakk-mw:
Thank you @samruddhikhandale,

Yes, the "mcr.microsoft.com/devcontainers/universal:2" image opens with the Jupyterlab editor. Probably because it already has Jupyter installed within it.

This image works with JupyterLab, with or **without** the python feature.

However, my goal is to **use features** to introduce the Python & JupyterLab into an image that **didn't** already have it; and then use that image to start in Codespace's Jupyter Editor.

[...]

PR #1199, opened 2024-11-25
Codespace unable to start in Jupyter using python feature with option "installJupyterLab": true 

Mathiyarasy:
Issue https://github.com/devcontainers/features/issues/870
- For image mcr.microsoft.com/devcontainers/base:ubuntu

Issue https://github.com/github/codespaces/issues/19082 
- For image mcr.microsoft.com/devcontainers/go:1-1.23-bullseye

JupyterLab is not opening from codespace due to PATH variable difference for non interactive shell when jupyterlab is installed for non root user

**Feature Name**
 - Python

**Changelog**
- Created a symlink for installed path of jupyter and added it to devcontainer-features.json containerEnv PATH variable.
- Remove changes added in [PR - 887](https://github.com/devcontainers/features/pull/887) which earlier aimed to fixed the issue

**Checklist**
- [x] Check if jupyter command  is recognised from non interactive shell

### DIFF
--- a/src/python/devcontainer-feature.json
+++ b/src/python/devcontainer-feature.json
@@ -67,7 +67,7 @@
     "PYTHON_PATH": "/usr/local/python/current",
     "PIPX_HOME": "/usr/local/py-utils",
     "PIPX_BIN_DIR": "/usr/local/py-utils/bin",
-    "PATH": "/usr/local/python/current/bin:/usr/local/py-utils/bin:${PATH}"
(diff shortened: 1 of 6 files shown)


eljog (approved):

Thompson1985 (approved):

Thompson1985 (approved):

SSmart95 (commented):
### `__`

PR #1210, opened 2024-12-04
aws-cli: add option to surpress unzip output

jasondamour:


jasondamour:

Kaniska244:
Hi @jasondamour ,

Thanks for the PR. Would you kindly help me understand the utility of having the unzip command execute in quite mode other than reducing log file size. Also would it be possible to add a test for this with the new flag introduced.

Please let me know in case of any concern/ clarification.

With Regards,
Kaniska

jasondamour:
The log output is so long that it can't be displayed in Github Actions. So the use case is just being able to see ANY other logs from my image build.

Kaniska244 (approved):

jasondamour:
@devcontainers/maintainers, Please review

v-Kaniska244:
Related to #1600

Issue #1202, opened 2024-11-27
[docker-in-docker] `version` is not respected when moby is set to `false`


hellt:
Hi,
I am trying to build a d-in-d devcontainer in a multiarch setting (linux/amd64 and linux/arm64).

When I added the arm64 platform to the action the build process failed with an error that moby package does not exist in the package repository (I pinned the moby version to 26.1.4).

I then set `moby:false` in my devcontainer inputs, but then the version is not respected at all:

```
#37 /bin/sh -c cp -ar /tmp/build-features-src/docker-in-docker_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/docker-in-docker_0  && cd /tmp/dev-container-features/docker-in-docker_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/docker-in-docker_0
  #37 0.062 ===========================================================================
  #37 0.062 Feature       : Docker (Docker-in-Docker)
  #37 0.062 Description   : Create child containers *inside* a container, independent from the host's docker instance. Installs Docker extension in the container along with needed CLIs.
  #37 0.062 Id            : ghcr.io/devcontainers/features/docker-in-docker
  #37 0.062 Version       : 2.12.0
  #37 0.062 Documentation : https://github.com/devcontainers/features/tree/main/src/docker-in-docker
  #37 0.062 Options       :
  #37 0.062     VERSION="26.1.4"
  #37 0.062     MOBY="false"
  #37 0.062     MOBYBUILDXVERSION="latest"
  #37 0.062     DOCKERDASHCOMPOSEVERSION="none"
  #37 0.062     AZUREDNSAUTODETECTION="true"
  #37 0.062     DOCKERDEFAULTADDRESSPOOL=""
  #37 0.062     INSTALLDOCKERBUILDX="true"
  #37 0.062     INSTALLDOCKERCOMPOSESWITCH="true"
  #37 0.062     DISABLEIP6TABLES="false"
  #37 0.062 ===========================================================================
  #37 0.074 Distro codename  'bookworm'  matched filter  'bookworm buster bullseye bionic focal hirsute impish jammy noble'

  #37 6.771 The following NEW packages will be installed:
  #37 6.771   containerd.io docker-ce docker-ce-cli
```

[...]

PR #1215, opened 2024-12-13
[docker-in-docker] - issue #1202 solution - moby=false, holds configured docker-ce and docker-ce-cli version

gauravsaini04:
Ref: #1202 

**Feature Name**
* Docker-in-docker

**Description**
* As a solution to issue #1202 , whenever moby was set to false in configured option while building the feature, it led to implicitly installing docker-compose-plugin, reinstalling the docker-ce and docker-ce-cli to an undesired version.

**_changelog_**
* Have pinned the docker-ce and docker-ce-cli to the configured versions for the feature

**Checklist**


thediveo:
Can this please be moved forward, as I'm also blocked in a slightly different use case compared to https://github.com/devcontainers/features/issues/1202 where I want to install multiple Docker versions into a dev container, activating only exactly one after another as needed. We want to use this to work on and test upgrade use cases. Currently, the unwanted upgrade makes installing a precise Docker version impossible as far as I can tell from my logs.

ddoyle2017 (approved):
LGTM. Thanks for adding this!

PR #1222, opened 2024-12-17
Avoid using Oryx to install Python by default

Scttmrx123:


Kaniska244:
Hello @Scttmrx123 ,

Thank you for the PR. Could you please give us more more context on why this change is required. Also please merge the upstream on your dev branch.

With Regards,
Kaniska

PR #1225, opened 2024-12-18
[aws-cli] install less dependency

### LINKED ISSUE #1224: [aws-cli] aws-cli feature does not contain `less` dependency
The https://ghcr.io/devcontainers/features/aws-cli does not install `less` which is a requirement for `aws-cli`.

```shellsession
$ aws sts get-caller-identity

Unable to redirect output to pager. Received the following error when opening pager:
[Errno 2] No such file or directory: 'less'

Learn more about configuring the output pager by running "aws help config-vars".
```

Setting the envvar overrides the default, allowing the commands to work.

```shellsession
$ export AWS_PAGER=""
$ aws sts get-caller-identity
... output ...
```

So in my mind, either the `less` dependency should be installed by the feature, or the envvar should be set to allow operation.
Yes, I realize this can be accomplished by adding `less` to a Dockerfile, but that is not the solution one should have to arrive to after the feature fails to work.

The "correct" one ought to be to install `less` instead of mucking with the user's env.

miketheman:
Fixes #1224

### DIFF
--- a/src/aws-cli/devcontainer-feature.json
+++ b/src/aws-cli/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "aws-cli",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "name": "AWS CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/aws-cli",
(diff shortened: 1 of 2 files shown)


miketheman:

miketheman:
An alternative solution proposed in #1226 - one of these should resolve the issue.

PR #1230, opened 2024-12-28
Use new CDN for dotnet builds

sliekens:
The `azureedge.net` domain might stop working as detailed in https://github.com/dotnet/core/issues/9671

Changes in this PR:

- **dotnet**: update the logic to determine the latest version
- **dotnet and oryx**: copy latest dotnet-install script from
  https://github.com/dotnet/install-scripts/blob/main/src/dotnet-install.sh
  (Normally this happens automatically by a GitHub Action, but there is a sense of urgency here, so I did it manually)

Additional context
- https://devblogs.microsoft.com/dotnet/critical-dotnet-install-links-are-changing/
- https://build5nines.com/retirement-of-azureedge-net-dns-edg-io-business-closure-and-what-you-need-to-know/

### DIFF
--- a/src/dotnet/scripts/vendor/dotnet-install.sh
+++ b/src/dotnet/scripts/vendor/dotnet-install.sh
@@ -423,11 +423,17 @@ get_normalized_architecture_for_specific_sdk_version() {
 # args:
 # version or channel - $1
 is_arm64_supported() {
-    #any channel or version that starts with the specified versions
-    case "$1" in
-        ( "1"* | "2"* | "3"*  | "4"* | "5"*) 
-            echo false
-            return 0
+    # Extract the major version by splitting on the dot
+    major_version="${1%%.*}"
+
+    # Check if the major version is a valid number and less than 6
+    case "$major_version" in
+        [0-9]*)  
+            if [ "$major_version" -lt 6 ]; then
+                echo false
+                return 0
+            fi
+            ;;
     esac
 
     echo true
@@ -951,6 +957,37 @@ get_absolute_path() {
 }
 
 # args:
+# override - $1 (boolean, true or false)
+get_cp_options() {
+    eval $invocation
+
+    local override="$1"
+    local override_switch=""
+
+    if [ "$override" = false ]; then
+        override_switch="-n"
+
+        # create temporary files to check if 'cp -u' is supported
+        tmp_dir="$(mktemp -d)"
+        tmp_file="$tmp_dir/testfile"
+        tmp_file2="$tmp_dir/testfile2"
+
+        touch "$tmp_file"
+
+        # use -u instead of -n if it's available
+        if cp -u "$tmp_file" "$tmp_file2" 2>/dev/null; then
+            override_switch="-u"
+        fi
+
+        # clean up
+        rm -f "$tmp_file" "$tmp_file2"
+        rm -rf "$tmp_dir"
+    fi
+
+    echo "$override_switch"
+}
+
+# args:
 # input_files - stdin
 # root_path - $1
 # out_path - $2
@@ -961,15 +998,7 @@ copy_files_or_dirs_from_list() {
     local root_path="$(remove_trailing_slash "$1")"
     local out_path="$(remove_trailing_slash "$2")"
     local override="$3"
-    local osname="$(get_current_os_name)"
-    local override_switch=$(
-        if [ "$override" = false ]; then
-            if [ "$osname" = "linux-musl" ]; then
-                printf -- "-u";
-            else
-                printf -- "-n";
-            fi
-        fi)
+    local override_switch="$(get_cp_options "$override")"
 
     cat | uniq | while read -r file_path; do
         local path="$(remove_beginning_slash "${file_path#$root_path}")"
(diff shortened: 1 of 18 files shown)


sliekens:
Cc @samruddhikhandale @Chuxel

src/dotnet/scripts/vendor/dotnet-install.sh
```diff
@@ -1295,6 +1379,11 @@ get_download_link_from_aka_ms() {
             return 1
         fi
 
+        sanitized_redirect_url=$(sanitize_redirect_url "$aka_ms_download_link")
```
chrmarti, line 1382: Do we know when and how the aka.ms redirects for the dotnet downloads will be updated? @bamurtaugh Do you have a contact to check with?

@sliekens Maybe the script could check if the redirect still points to `azureedge.net` and only in that case run the sanitize step? That might be more robust towards changes of the redirects.

chrmarti (changes requested):
Great, thanks for the PR! Left a few questions.

sliekens:
@chrmarti the dotnet-install.sh script is copied from https://github.com/dotnet/install-scripts/blob/main/src/dotnet-install.sh so I'm not qualified to answer those questions but I think @mmitche can help.

src/dotnet/scripts/vendor/dotnet-install.sh
chrmarti: https://github.com/dotnet/core/issues/9671 says about the aka.ms redirects: `We expect these links to be changed in early January.` So this should keep working without modification.

chrmarti: Seeing that you copied https://github.com/dotnet/install-scripts/blob/main/src/dotnet-install.sh, I guess that might see another update?

mmitche:
Yeah, the install script is going to get another update soon, removing the use of the azureedge domains, at least outside of sanitization. So the script would only try to access the new domains. We're not quite sure yet when we will update old aka.ms links. Once they are updated, then we would remove the sanitization altogether.

chrmarti:
@mmitche Great, this sounds like we should wait for that imminent update of the install script?

Updating the aka.ms redirects would smooth things out I guess, is there a reason to hold back on that?

chrmarti:
> Don't forget that script is updated here automatically every Sunday, e.g. see #1236
> 
> I just realized I forgot to bump the version 🤔. Not sure about the semantics of this change either. Probably not a major bump, the behavior didn't change that much and usage remains the same. On the other hand, this change could break users behind corporate firewalls where the CDN is not allowed.

Minor (or patch) might work best since that will be picked up when the devcontainer.json only specifies the major version. There is also an optional lockfile devcontainer-lock.json that pins the exact feature version and checksum and would hold the update back.

I guess we could just run the workflow you mention then (catching up on how we do this): https://github.com/devcontainers/features/actions/workflows/update-dotnet-install-script.yml

sliekens:
A new minor version makes the most sense to me, I'll do that. I can revert the changes to `dotnet-install.sh` vendor script in this PR if desired, but the changes in the `dotnet-helpers.sh` script are still needed, I wrote those scripts specifically for the devcontainer feature to find the latest version no matter the STS or LTS status.

sliekens:
I guess I should fix the failing test while I'm at it 🤔.

sliekens:
@chrmarti I think this work is complete now.

Final changes:
- Fixed failing scenarios which installed the latest dotnet version (now .NET 9) but then tried to run a .NET 8 project
- Bumped dotnet minor version
- Bumped oryx minor version

chrmarti:
@sliekens Looks great, thanks! We can update the script again when the original changes later this/next week. Adding @gauravsaini04 as a second reviewer.

chrmarti (approved):

gauravsaini04 (approved):
All the changes done, LGTM 👍🏻
Thanks @sliekens !

eljog:
> We can update the script again when the original changes later this/next week

SGTM, @gauravsaini04 let's make a note of that followup

src/dotnet/scripts/vendor/dotnet-install.sh
```diff
@@ -1306,7 +1395,9 @@ get_download_link_from_aka_ms() {
 get_feeds_to_use()
 {
     feeds=(
+    "https://builds.dotnet.microsoft.com/dotnet"
     "https://dotnetcli.azureedge.net/dotnet"
```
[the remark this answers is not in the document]
richlander: Yes, please remove the old URLs.

src/dotnet/scripts/vendor/dotnet-install.sh
richlander: Yes.

- We will update the file again in the first half of January. Please grab copies.
- We will be providing users with guidance on how to do freshness checks on when to grab new copies.

Related: https://github.com/dotnet/install-scripts/issues/566

src/dotnet/scripts/vendor/dotnet-install.sh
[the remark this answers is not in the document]
chrmarti: Seeing there is a PR getting ready on the original, we can just copy that once it is merged. Thanks.

cetinbug:
For our Azure Functions, we currently use the URL **functionscdn.azureedge.net** to retrieve extension bundles. However, after reviewing the issue, I noticed that there is no specific mention of the subdomain **_functionscdn_** in the migration details. Could you please clarify if this subdomain will also be migrated and, if so, to which domain?

For example, we are getting extension bundles from a url like: _"https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle.Preview/$EXTENSION_BUNDLE_VERSION"_

PR #1252, opened 2025-01-27
[docker-outside-of-docker] - Correction in fetching previous version from github api url

gauravsaini04:
**Feature Name**
* `Docker-Outside-Of-Docker`

**Description of Changes**
* Correcting the fallback method for `Docker-outside-of-Docker` Feature as previously it was fetching the `latest` version in a plight to fetch the `previous` version from the `GitHub API URL`, which was misleading and therefore now fetching the second element's `tag_name` property's value in `releases` array json for correcting the above

**_Changelog_**
* Made changes to `install.sh` file
* Made changes to `docker_build_compose_fallback.sh` file in test case scenarios implementations of docker-outside-of-docker

**Checklist**

### DIFF
--- a/src/docker-outside-of-docker/devcontainer-feature.json
+++ b/src/docker-outside-of-docker/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "docker-outside-of-docker",
-    "version": "1.6.0",
+    "version": "1.6.1",
     "name": "Docker (docker-outside-of-docker)",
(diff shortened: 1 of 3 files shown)


cguess:
This same concern happened again today with a new release. Any chance this could be reviewed and merged?

Mathiyarasy (changes requested):
If user passes the version as v2 this condition will not satisfy for latest tag and fallback mechanism will fail

Mathiyarasy (approved):

ddoyle2017 (approved):

Issue #1270, opened 2025-02-11
CuDnn installation script regex for version is buggy


Superskyyy:
```bash
apt-cache policy libcudnn9-cuda-12 | grep "Candidate" | awk '{print $2}' | grep -Eo '^[^-1+]*'
9.7.

```
@prathameshzarkar9 FYI

Kaniska244:
Hi @Superskyyy ,

The PR is merged. Kindly let me know in case of further concerns with this.

With Regards,
Kaniska

PR #1275, opened 2025-02-14
[nvidia-cuda] - Cudnn install script issue, unable to fetch latest version of libcudnn9 library

Kaniska244:
**Ref#** https://github.com/devcontainers/features/issues/1270 

**Description:** Unable to fetch the latest libcudnn9 library version for cuda 12.3 or greater versions.

**Root Cause:** The regular expression used for fetching latest libcudnn9 library version was incorrect as it wasn't expecting digit 1 to be present in the version number.

**Changelog:** Changes done in install.sh script of nvidia-cuda to correct the regular expression & also changes done in the test scripts to test the presence of latest libcudnn9 library instead of specific version when the libcudnn version isn't specified.   

**Checklist:**

- Checked that applied changes work as expected.


ddoyle2017 (approved):
As we discussed in stand-up, this approach makes sense! LGTM⚡

PR #1277, opened 2025-02-19
Update install.sh

andreaske76:
If the baseimage contains a preinstalled hugo version, the change checks the version and if it's differs from $VERSION then it installs hugo with the $VERSION


Kaniska244:
Hi @andreaske76 ,

Thank you for the PR! Would you kindly provide some details on the issue to proceed with this PR further.

With Regards,
Kaniska

Kaniska244:
Hi @andreaske76 ,

Thank you for the PR. Would it be possible to add a test for this specific scenario(installing the correct version if not present) you have fixed as part of this PR. 

With Regards,
Kaniska

lanrat:
Looks like this PR would fix an issue I've been having with the devcontainer using an outdated version of hugo.

Could this be updated to also check $EXTENDED matches in addition to $VERSION?

PR #1291, opened 2025-03-07
rremoved shell env which defaults to the shell /bin/bash

sireeshajonnalagadda:
Modified Dockerfile to remove the shell env which defaults to the shell /bin/bash.


sireeshajonnalagadda:
provided a workaround to the user.
Not required to change the dockerfile of universal image.

PR #1298, opened 2025-03-17
Add `apt clean` to reduce layer size

DavidS-ovm:
The downloaded packages do not need to be kept in the layer. Running `apt clean` will remove them from the cache.

To quote from apt's man page:
> clean clears out the local repository of retrieved package files. It removes everything but the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -27,6 +27,7 @@ set -e
 
 # Clean up
 rm -rf /var/lib/apt/lists/*
+apt clean
 
 # Setup STDERR.
 err() {
@@ -427,6 +428,7 @@ if [ -f "/usr/local/share/docker-init.sh" ]; then
     echo "/usr/local/share/docker-init.sh already exists, so exiting."
     # Clean up
     rm -rf /var/lib/apt/lists/*
+    apt clean
     exit 0
 fi
 echo "docker-init doesn't exist, adding..."
@@ -640,5 +642,6 @@ chown ${USERNAME}:root /usr/local/share/docker-init.sh
 
 # Clean up
 rm -rf /var/lib/apt/lists/*
+apt clean
 
 echo 'docker-in-docker-debian script has completed!'


bamurtaugh:
cc @Kaniska244 I'm linking https://github.com/orgs/devcontainers/discussions/206 as well, would love your input when you get a moment!

src/docker-in-docker/install.sh
```diff
@@ -27,6 +27,7 @@ set -e
 
 # Clean up
 rm -rf /var/lib/apt/lists/*
+apt clean
```
eitsupi, line 30: Could this not make sense as it only removes the installation stuff in the previous layer?

DavidS-ovm: Yeah, I don't _quite_ understand why this script cleans at the start and then again at the end :shrug: 

As a start I see an existing clean up, I add my change. If someone with a better overview of the project points me in a different direction to solve this at a higher level so that the apt cache is never stored in the image (e.g. using `RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt/lists` everywhere), I'd be happy to also do that.

eitsupi: I believe the removal of `/var/lib/apt/lists/*` is done because subsequent installations may fail if the cache remains in the lower layers.
The size of the image is irrelevant.

DavidS-ovm: Can you elaborate on "The size of the image is irrelevant"? My team is paying for image storage by the byte and devcontainer download time is a regular annoyance, so image size is very much of concern for us.

DavidS-ovm: I'm happy to add this to all the places touched in #210, if that's what you're implying.

eitsupi: What I'm trying to say is that I'm wondering if it's worth adding this at the beginning of the script.
(Because it doesn't affect the layer size)

Kaniska244: Hello @DavidS-ovm ,

Thank you for the PR. I was trying test docker-in-docker feature with this PR change. In fact I see that with `apt clean` right at the beginning of the `install.sh`, package installation failing. It doesn't remove the directory structure or the lock file, so later on the following condition in `apt_get_update()` function body doesn't satisfy & `apt-get update -y` doesn't execute before installing the additional packages for docker-in-docker feature.
```
apt_get_update()
{
    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
        apt-get update -y
    fi
}
``` 
It would be best to have the clean-up done once & done right at the end of the `install.sh` execution & that should serve the purpose of reducing the layer size in the resultant image. Kindly let me know in case of further concern.

With Regards,
Kaniska

DavidS-ovm: Thanks for the analysis! This is indeed more complex than I initially expected. Meanwhile I've done some local optimizations for other containers in my work and I'm now using

```
RUN --mount=type=cache,target=/var/cache/apt \
    --mount=type=cache,target=/var/lib/apt/lists \
    [...]
```

any time apt is getting called. This ensures that 
* apt caches and metadata are never baked into any layer
* individual steps can re-use previous steps' already downloaded data, and
* if buildx is appropriately configured, the cache can be used across multiple builds, further improving efficiency if requested
* individual steps don't need to care about refreshing and cleaning apt metadata
 
Poking around in the devcontainer CLI, this seems to be the appropriate location to add that:

https://github.com/devcontainers/cli/blob/5c9623c78d17ee4f2d75899c1d9c59a9b266611e/src/spec-configuration/containerFeaturesConfiguration.ts#L292-L357

additionally there's `getContainerFeaturesBaseDockerFile` in the same file that looks like a good place to `apt update` once for the entire build.

Does that sound sensible?

Kaniska244: Hi @DavidS-ovm ,

Thank you for the contribution & the detailed analysis on this issue. I will check this further & update you.

PR #1302, opened 2025-03-19
[docker-in-docker] - Correction in fetching previous version from github api url 

Mathiyarasy:
**Feature Name**

  - Docker-In-Docker
   
**Description of Changes**

- Correcting the fallback method for Docker-in-Docker Feature as previously it was fetching the latest version to fetch the previous version from the GitHub API URL. This was misleading and therefore now fetching the second element's tag_name property's value in releases array json for correcting the above

- Docker-in-docker dockerDashComposeVersion is confusing. As per [Issue](https://github.com/devcontainers/features/issues/745) dockerDashComposeVersion should default to latest i.e.)v2
But current implementation shows latest and v2 as two different options
- Changed the default to v2 as similar to docker-outside-of-docker issue
-When we use v2 code is implemented to fetch the latest tag from docker compose
- If latest tag is created but code is not released will fallback to the previous release tag
- If latest is used instead of v2 still the latest tag is fetched from v2 version so there shouldnt be any change in functionality

**Changelog**

- Made changes to install.sh file
- Made changes to docker_build_compose_fallback.sh file in test case scenarios implementations of docker-outside-of-docker
- Changed the dockerDashComposeVersion default value to v2 instead of latest in devcontainer.features.json 

**Checklist**

 - [X] Checked that applied changes work as expected


Kaniska244 (approved):
Changes are fine.

Issue #1292, opened 2025-03-09
powershell feature but for all distros


elibroftw:
how can we make this work on all platforms? I couldn't install powershell in my dockerfile (almalinux) even after installing dependencies:

```
nothing provides krb5 needed by powershell
```

Then I see this powershell feature which is debian-only

This is impeding adoption of powershell in my opinion.

My install script FYI:

```bash
#!/usr/bin/env bash

dnf install -y wget
wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq && chmod +x /usr/bin/yq
dnf install -y krb5-libs libicu openssl-libs zlib
# Problem: conflicting requests
#   - nothing provides krb5 needed by powershell-7.5.0-1.cm.aarch64 from @commandline
latest_release=$(curl -s https://api.github.com/repos/PowerShell/PowerShell/releases/latest) && \
    rpmlink=$(echo "$latest_release" | arch=$(arch) yq -r '[.assets[] | select(.name == ("*" + strenv(arch) + ".rpm")) | .browser_download_url'][0]) && \
    dnf install -y $rpmlink
```

Seems like `rpm -Uvh --nodeps  powershell-7.5.0-1.cm.x86_64.rpm $rpmlink` does the trick

sireeshajonnalagadda:
Hi @elibroftw, 
Thank you for raising the issue! I will check and get back to you. 

Thanks, 
Sireesha

sireeshajonnalagadda:
Hi, 
Can you please help me with the exact distro you are trying to install powershell for?

elibroftw:
Alma Linux 9

Mathiyarasy:
Fixed in :https://github.com/devcontainers/features/pull/1303

PR #1303, opened 2025-03-21
Extending support of powershell feature for AlmaLinux

sireeshajonnalagadda:
Feature Name

- Powershell

Description of Changes

- Extending support of powershell feature for other linux distributions (alma linux)

- As per the user request https://github.com/devcontainers/features/issues/1292

Changelog

- Modified the install.sh to work the expected installation for AlmaLinux 

- Modified the test cases scenarios to meet the required checks for the powershell.

Checklist

 

- [ ] Checked that applied changes work as expected


src/powershell/install.sh
```diff
@@ -72,19 +140,42 @@ apt_get_update()
...
+                apt-get update
+                apt-get install -y --no-install-recommends "$package"
+            else
+                echo "Package $package is already installed (APT)."
+            fi
+        done
+    elif command -v rpm > /dev/null 2>&1 && command -v dnf > /dev/null 2>&1; then
+        # If rpm and dnf exist, assume DNF-based system (RHEL/AlmaLinux)
+        for package in "$@"; do
+            if ! rpm -q "$package" > /dev/null 2>&1; then
```
Kaniska244, line 158: Hi @sireeshajonnalagadda ,

Why do we need to check with rpm for the pakcge existence, can't we use dnf directly in this case?

src/powershell/install.sh
```diff
@@ -17,18 +17,86 @@ POWERSHELL_MODULES="${MODULES:-""}"
...
 GPG_KEY_SERVERS="keyserver hkp://keyserver.ubuntu.com
 keyserver hkp://keyserver.ubuntu.com:80
 keyserver hkps://keys.openpgp.org
-keyserver hkp://keyserver.pgp.com"
+keyserver hkp://keyserver.pgp.com
+keyserver hkp://keyserver.fedoraproject.org
+keyserver hkps://keys.openpgp.org
+keyserver hkp://pgp.mit.edu
+keyserver hkp://keyserver.redhat.com"
+
```
Kaniska244, line 32: Hi @sireeshajonnalagadda ,

Do we need all the above keyservers for almalinux support? If so, kindly add suitable comments.

src/powershell/install.sh
```diff
@@ -202,7 +293,12 @@ install_using_github() {
         check_packages git
     fi
     if [ "${architecture}" = "amd64" ]; then
+        architecture="amd64"
+        elif [ "${architecture}" = "x86_64" ]; then 
+        architecture="x86_64"
```
Kaniska244, line 298: Hi @sireeshajonnalagadda ,

The amd64 or x86_64 are basically the same CPU architecture, so should work with same installation package. Why do we need multiple conditions here?

src/powershell/install.sh
```diff
@@ -212,7 +308,12 @@ install_using_github() {
     fi
 
     # Ugly - but only way to get sha256 is to parse release HTML. Remove newlines and tags, then look for filename followed by 64 hex characters.
-    curl -sSL -o "release.html" "https://github.com/PowerShell/PowerShell/releases/tag/v${POWERSHELL_VERSION}"
+    #curl -sSL -o "release.html" "https://github.com/PowerShell/PowerShell/releases/tag/v${POWERSHELL_VERSION}" 
+    wget https://github.com/PowerShell/PowerShell/releases/download/v${POWERSHELL_VERSION}/${powershell_filename}
```
Kaniska244, line 312: Hi @sireeshajonnalagadda ,

Kindly explain the reason behind switching from curl to wget in this scenario. Please add suitable comment for the same.

Kaniska244 (changes requested):
Hi @sireeshajonnalagadda ,

Left a few comments. Kindly check & let me know in case of further concerns.

src/powershell/install.sh
Kaniska244: Looks fine with the latest check-in.

src/powershell/install.sh
Kaniska244: This is also fine with code comments added specifically for this in the latest check-in

Kaniska244 (approved):

src/powershell/install.sh
```diff
@@ -201,18 +296,24 @@ install_using_github() {
     if ! type git > /dev/null 2>&1; then
         check_packages git
     fi
-    if [ "${architecture}" = "amd64" ]; then
-        architecture="x64"
+    if [ "${architecture}" = "x86_64" ]; then
```
ddoyle2017, line 299: What was the reason for switching the logic around setting the architecture? I could use some context here 😅

sireeshajonnalagadda:
Hi @ddoyle2017 , 
Changes are made related to amd64, removal of x86_64  working as expected, please review the changes and let me know in case of any further queries!

ddoyle2017 (approved):

PR #1312, opened 2025-03-28
[Ruby] - 3.1 EOL and 3.4 addition

Kaniska244:
**Ref:** [Issue](https://github.com/devcontainers/images/issues/90)

**Description:** Ruby 3.1 EOL on March 31st 2025 & Adding Ruby 3.4 available since Dec 2024 to the ruby devcontainer feature.

**Changelog:** 

- Change in devcontainer-feature.json to remove ruby 3.1 & add ruby 3.4
- Change in Scenarios.json to correct the Ruby test criteria based on latest versions.
- Change in install_additonal_ruby.sh & ruby_fallback_test.sh test scripts correct the expected ruby versions.

**Checklist:**

- All changes work as expected.

### DIFF
--- a/src/ruby/devcontainer-feature.json
+++ b/src/ruby/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "ruby",
-    "version": "1.2.3",
+    "version": "1.3.0",
     "name": "Ruby (via rvm)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/ruby",
     "description": "Installs Ruby, rvm, rbenv, common Ruby utilities, and needed dependencies.",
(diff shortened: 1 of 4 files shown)


ddoyle2017 (approved):
Changes look good! However it might probably be better to split these into separate PRs next time: 
- one for the removal of 3.1
- one for adding 3.4

Kaniska244:
> Changes look good! However it might probably be better to split these into separate PRs next time:
> 
> * one for the removal of 3.1
> * one for adding 3.4

Hi @ddoyle2017 ,

That did cross my mind while doing this PR. I decided otherwise because of the fact that for ruby feature both the higher & lower versions were outdated already which prompted me to update both, while doing so thought that it could be better I do it at once in the same PR so that in case of any issue with both the versions changing, it would be easy to isolate the problem.
Next time onwards I will come up separate PR's for things like this. Kindly let me know if its fine to go ahead with this PR for the time being or should I come up with separate PRs.

With Regards,
Kaniska

PR #1314, opened 2025-04-01
Custom instructions for features

chrmarti:
https://github.com/microsoft/vscode-copilot/issues/14749


bamurtaugh (approved):
Looks great to me, thank you!

eitsupi:
Would adding something like this be recommended for other features?
Might be worth updating the SPEC and adding a field for description for LLM?

PR #1324, opened 2025-04-08
feat: add UV to python feature

FabrizioCafolla:
Hi, I opened a pull request for adding UV to Python feature


FabrizioCafolla:

karthiknadig (approved):

chrmarti:
Opened https://github.com/devcontainers/features/issues/1327 for the test failure which seems to be triggered by some update upstream.

karthiknadig:
The test failure seems to be caused by attempting to install `uv` using python 3.6. python 3.6 is deprecated and should be updated for 3.9.

karthiknadig:
@FabrizioCafolla i think all we need to do is only install it when version of Python is greater than 3.9.

@chrmarti can you provide any guidance for this?

chrmarti:
Maybe add a version check for Python when the util is `uv` here: https://github.com/devcontainers/features/blob/6654579de4c31cd9f9f9e19e873521f502403929/src/python/install.sh#L871

There already is a version check a few lines below that.

/cc @gauravsaini04 for insights.

FabrizioCafolla:
@chrmarti I added a check in the code where you indicated.

FabrizioCafolla:
@chrmarti @karthiknadig any news about this pr?

chrmarti:
Apologies for the delay. It looks like the tests fail because they need the same version check as the install script.

Issue #1235, opened 2025-01-03
[docker-in-docker] Not working on fedora kernel > 6.11.7-300.fc41.x86_64


michalholis:
Feature https://github.com/devcontainers/features/tree/main/src/docker-in-docker for some reason doesn't work for us on Fedora 41 on kernel versions 6.11.7-300.fc41.x86_64. Kernel 6.11.7-300.fc41.x86_64 is the last one working without issues. On any newer kernel any docker command returns `Cannot connect to the Docker daemon at unix:/var/run/docker.sock. Is the docker daemon running?`.

Kaniska244:
Hello @michalholis ,

Thank you for reporting the issue. I will check and get back on this.

With Regards,
Kaniska

David-337:
Same issue here on Bluefin which is a distro based on Fedora

HerrNaN:
Any updates on this?

matus-sabo:
Some information and workaround (debian bookworm)

`sudo dockerd --debug`

```bash
failed to start daemon: Error initializing network controller: error obtaining controller instance: failed to register "bridge" driver: failed to create NAT chain DOCKER: iptables failed: iptables -t nat -N DOCKER: modprobe: FATAL: Module ip_tables not found in directory /lib/modules/6.12.11-200.fc41.x86_64
iptables v1.8.9 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
Perhaps iptables or your kernel needs to be upgraded.
 (exit status 3)
```

after mount volume /lib/modules/:/lib/modules

`sudo dockerd --debug`
```bash
failed to start daemon: Error initializing network controller: error obtaining controller instance: failed to register "bridge" driver: failed to create NAT chain DOCKER: iptables failed: iptables -t nat -N DOCKER: iptables v1.8.9 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
Perhaps iptables or your kernel needs to be upgraded.
 (exit status 3)
```

after adding this to Dockerfile

as root user
```Dockerfile
RUN apt install \
   kmod \
   -y
```

or as sudo user
```Dockerfile
RUN sudo apt install \
   kmod \
   -y
```

`docker info` finally working 🚀

spietras:
In my case, I am using `mcr.microsoft.com/devcontainers/base:1.2.3-ubuntu-24.04` image as a base and getting the same thing as @matus-sabo.

[...]

PR #1328, opened 2025-04-11
Docker-in-docker feature support  for fedora

sireeshajonnalagadda:
Feature Name

- Docker-in-docker

Description of Changes

- Extending support of Docker-in-docker feature for other linux distributions (Fedora)

- As per the user request https://github.com/devcontainers/features/issues/1235

Changelog

- Modified the install.sh to work the expected installation for Fedora

- Wrote test case scenarios to meet the required checks for the Docker-in-docker

Checklist

- [ ☑️ ]  Checked that applied changes work as expected

### DIFF
--- a/src/docker-in-docker/install.sh
+++ b/src/docker-in-docker/install.sh
@@ -11,7 +11,7 @@
 DOCKER_VERSION="${VERSION:-"latest"}" # The Docker/Moby Engine + CLI should match in version
 USE_MOBY="${MOBY:-"true"}"
 MOBY_BUILDX_VERSION="${MOBYBUILDXVERSION:-"latest"}"
-DOCKER_DASH_COMPOSE_VERSION="${DOCKERDASHCOMPOSEVERSION:-"v2"}" #v1, v2 or none
+DOCKER_DASH_COMPOSE_VERSION="${DOCKERDASHCOMPOSEVERSION:-"latest"}" #v1, v2 or none
 AZURE_DNS_AUTO_DETECTION="${AZUREDNSAUTODETECTION:-"true"}"
 DOCKER_DEFAULT_ADDRESS_POOL="${DOCKERDEFAULTADDRESSPOOL:-""}"
 USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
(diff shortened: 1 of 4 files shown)


src/docker-in-docker/install.sh
```diff
@@ -247,7 +279,7 @@ fi
 
 # Refresh apt lists
 apt-get update
-
+fi
```
Kaniska244, line 282: Please correct the indentation for better readability.

src/docker-in-docker/install.sh
```diff
@@ -62,19 +62,37 @@ fi
...
     fi
+elif [ "$ID" = "fedora" ] || [ "$ID" = "centos" ] || [ "$ID_LIKE" == "rhel" ]; then
+    if ! dnf list installed "$@" > /dev/null 2>&1; then
+        dnf -y install "$@"
+    fi
+fi
+
 }
+# Install dependencies for fedora
+if ! command -v git wget which&> /dev/null; then
```
Kaniska244, line 92: Just out of curiosity, how is this condition only passing for fedora & not for debian or ubuntu?

src/docker-in-docker/install.sh
```diff
@@ -62,19 +62,37 @@ fi
...
 {
-    if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
+     . /etc/os-release
+    if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
+       if [ "$(find /var/lib/apt/lists/* | wc -l)" = "0" ]; then
         echo "Running apt-get update..."
         apt-get update -y
+        apt-get install -y gnupg curl
+        apt-get -y install --no-install-recommends "$@"
+       fi
```
Kaniska244, line 72: These installations should not be done in this `apt_get_update()` function. Would it possible to move them below as part of code in this [code](https://github.com/devcontainers/features/pull/1328/files#diff-a72a4e49c44e54f09feb84b7767d355d68d34fbee561d8f96a99d12d2488b9c5L215), if the package isn't included there already.

sireeshajonnalagadda:
Hi @kaniska, 
Made the changes according to the review comments, please review and let me know in case of any further concerns/clarity.

src/docker-in-docker/install.sh
```diff
@@ -293,33 +327,93 @@ if [ "${USE_MOBY}" = "true" ]; then
...
+        echo "baseurl=https://packages.microsoft.com/yumrepos/microsoft-fedora$RELEASE_VER" >> /etc/yum.repos.d/microsoft.repo
+        echo "enabled=1" >> /etc/yum.repos.d/microsoft.repo
+        echo "gpgcheck=1" >> /etc/yum.repos.d/microsoft.repo
+        echo "gpgkey=https://packages.microsoft.com/keys/microsoft.asc" >> /etc/yum.repos.d/microsoft.repo
+        
+        # Install Moby Engine and CLI
+        echo "Installing Moby packages..."
+        sudo dnf install -y moby-engine moby-cli moby-buildx --skip-unavailable
+        
+    else
```
Kaniska244, line 354: Hi @sireeshajonnalagadda ,

When does the code inside this `else` condition get executed ? I see [here](https://github.com/devcontainers/features/pull/1328/files#diff-a72a4e49c44e54f09feb84b7767d355d68d34fbee561d8f96a99d12d2488b9c5R404) this `install_docker_or_moby` function called only when `USE_MOBY` flag is set to `true`. Please let me know if I am missing something.

Kaniska244 (changes requested):

test/docker-in-docker/scenarios.json
```diff
@@ -166,5 +166,16 @@
         },
         "remoteUser": "vscode",
         "onCreateCommand": "docker ps && sleep 5s && docker ps"
+    },
+    "fedora_kernel": {
+        "image": "fedora:41",
+        "features": {
+            "docker-in-docker": {
+                "iptables": "true",
+                "moby": "true",
```
Kaniska244, line 175: Hi @sireeshajonnalagadda ,

We should have another test case also with `"moby": "false"` for fedora

sireeshajonnalagadda:
HI @Kaniska244, 
Addressed the review comments and made few changes that works for both ubuntu and fedora, please review the latest changes and let me know in case of any Queries!

PR #1335, opened 2025-04-14
[dotnet] - Fixing the dotnet smoke test issue with changes in test script.

Kaniska244:
**Ref#** [Issue](https://github.com/devcontainers/internal/issues/260)

**Description:** This change is done to fetch the dotnet sdk `lts` version number in the same manner how the version is picked up while downloading the package in the dotnet_install.sh vendor script. As the current way of getting the version # for dotnet sdk `lts` via specific CDN url is giving incorrect version. See [here](https://github.com/dotnet/sdk/issues/47865)

**Changelog:** Change for this is done only in the following test scripts. & that too for lts scenario alone.
- test/dotnet/dotnet_helpers.sh
- test/dotnet/install_dotnet_lts.sh

**Checklist:**

- All changes work as expected.

### DIFF
--- a/src/dotnet/devcontainer-feature.json
+++ b/src/dotnet/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "dotnet",
-    "version": "2.2.1",
+    "version": "2.2.2",
     "name": "Dotnet CLI",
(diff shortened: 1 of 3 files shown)


ddoyle2017 (approved):
Nice catch!

Issue #1192, opened 2024-11-18
Conda devcontainer not support on Apple Silicon: `Architecture aarch64 unsupported`


cedricvidal:
Error logs

```
 > [dev_containers_target_stage 5/5] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=conda_0,target=/tmp/build-features-src/conda_0     cp -ar /tmp/build-features-src/conda_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/conda_0  && cd /tmp/dev-container-features/conda_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/conda_0:
0.112 ===========================================================================
0.112 Feature       : Conda
0.112 Description   : A cross-platform, language-agnostic binary package manager
0.112 Id            : ghcr.io/devcontainers/features/conda
0.112 Version       : 1.0.9
0.112 Documentation : https://github.com/devcontainers/
[2024-11-18T20:44:41.077Z] features/tree/main/src/conda
0.112 Options       :
0.112     VERSION="3.10"
0.112     ADDCONDAFORGE="false"
0.112 ===========================================================================
0.115 + export DEBIAN_FRONTEND=noninteractive
0.115 + DEBIAN_FRONTEND=noninteractive
0.116 + rm -rf '/var/lib/apt/lists/*'
0.117 ++ id -u
0.118 + '[' 0 -ne 0 ']'
0.118 + rm -f /etc/profile.d/00-restore-env.sh
0.119 ++ sh -lc 'echo $PATH'
0.122 + echo 'export PATH=/opt/conda/bin:/usr/local/python/current/bin:/usr/local/py-utils/bin:/usr/local/share/nvm/current/bin:/usr/local/bin:$PATH'
0.122 + chmod +x /etc/profile.d/00-restore-env.sh
0.123 + '[' vscode = auto ']'
0.123 + '[' vscode = automatic ']'
0.123 + '[' vscode = none ']'
0.123 + id -u vscode
0.123 ++ uname -m
0.124 + architecture=aarch64
0.124 + '[' aarch64 '!=' x86_64 ']'
0.124 + echo '(!) Architecture aarch64 unsupported'
0.124 (!) Architecture aarch64 unsupported
0.124 + exit 1
0.125 ERROR: Feature "Conda" (ghcr.io/devcontainers/features/co
[...]

PR #1342, opened 2025-04-21
[anaconda] - feature installation support for arm64/ aarch64.

Kaniska244:
**Ref#** https://github.com/devcontainers/features/issues/1192

**Description:** The change is mainly done to to extend installation  support of anaconda feature to arm64 architecture.

**Changelog:**  The following files files are changed.

- src/anaconda/install.sh - Change is done to extend support for arm64 architecture, upgrade the anaconda installation package version, fixing permissions & upgrading python packages for which older, vulnerable packages were installed. Also further changes made to extend this feature to support RHEL based linux distributions.
- src/anaconda/devcontainer-feature.json - Version bump.
- test/anaconda/install_anaconda_bookworm.sh - Added test script for anaconda installation on debian bookworm.
- test/anaconda/install_anaconda_bullseye.sh - Added test script for anaconda installation on debian bullseye.
- test/anaconda/install_anaconda_jammy.sh - Added test script for anaconda installation on ubuntu jammy.
- test/anaconda/install_anaconda_noble.sh - Added test script for anaconda installation on ubuntu noble.
- test/anaconda/install_anaconda_noble_without_user.sh - Added test script for anaconda installation on ubuntu noble without user in the input.
- test/anaconda/scenarios.json - Added for all test cases.

**Checklist:**

- Checked that applied changes work as expected.

### DIFF
--- a/src/terraform/devcontainer-feature.json
+++ b/src/terraform/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "terraform",
-    "version": "1.4.0",
+    "version": "1.4.1",
     "name": "Terraform, tflint, and TFGrunt",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/terraform",
     "description": "Installs the Terraform CLI and optionally TFLint and Terragrunt. Auto-detects latest version and installs needed dependencies.",
--- a/src/terraform/install.sh
+++ b/src/terraform/install.sh
@@ -50,6 +50,15 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+# Detect Ubuntu Noble and use new repo setup, else use legacy GPG logic
+IS_NOBLE=0
+if grep -qi 'ubuntu' /etc/os-release; then
+    . /etc/os-release
+    if [[ "$VERSION_CODENAME" == "noble" ]]; then
+        IS_NOBLE=1
+    fi
+fi
+
 # Get the list of GPG key servers that are reachable
 get_gpg_key_servers() {
     declare -A keyservers_curl_map=(
@@ -89,7 +98,7 @@ receive_gpg_keys() {
         keyring_args="--no-default-keyring --keyring $2"
     fi
     if [ ! -z "${KEYSERVER_PROXY}" ]; then
-	keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
+        keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
     fi
 
     # Install curl
@@ -101,6 +110,21 @@ receive_gpg_keys() {
     export GNUPGHOME="/tmp/tmp-gnupg"
     mkdir -p ${GNUPGHOME}
     chmod 700 ${GNUPGHOME}
+    
+    # Special handling for HashiCorp GPG key on Ubuntu Noble
+    if [ "$IS_NOBLE" -eq 1 ] && [ "$keys" = "$TERRAFORM_GPG_KEY" ]; then
+        echo "(*) Ubuntu Noble detected, using Keybase for HashiCorp GPG key import...."
+        curl -fsSL https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
+        if ! gpg --list-keys "${TERRAFORM_GPG_KEY}" > /dev/null 2>&1; then
+            gpg --list-keys
+            echo "(*) Warning: HashiCorp GPG key not found in keyring after import."
+            echo "    Continuing installation without GPG verification on Ubuntu Noble."
+            echo "    This is expected behavior for Ubuntu Noble due to keyserver issues."
+            return 1  # Return failure to indicate GPG verification should be skipped
+        fi
+        return 0
+    fi
+
     echo -e "disable-ipv6\n$(get_gpg_key_servers)" > ${GNUPGHOME}/dirmngr.conf
     # GPG key download sometimes fails for some reason and retrying fixes it.
     local retry_count=0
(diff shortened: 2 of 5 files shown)


Mathiyarasy (approved):

src/anaconda/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "anaconda",
-    "version": "1.0.13",
+    "version": "1.0.14",
```
AlvaroRausell: Can you bump minor version, i.e. `1.1.0`

src/anaconda/install.sh
```diff
@@ -13,11 +13,63 @@ USERNAME="${USERNAME:-"${_REMOTE_USER:-"automatic"}"}"
...
+        echo "No supported package manager found (apt-get, apk, dnf, microdnf, tdnf, yum)."
+        exit 1
+    fi
+}
+
+detect_package_manager
+
 # Clean up
-rm -rf /var/lib/apt/lists/*
+rm -rf "$PKG_LISTS"
```
AlvaroRausell: Can we run `PKG_CLEAN` instead? Many distros have cleanup mechanisms that I would rather use than removing the files ourselves if we can help it

src/anaconda/install.sh
```diff
@@ -66,15 +123,101 @@ updaterc() {
...
+                echo "Running $PKG_UPDATE..."
+                eval "$PKG_UPDATE"
+                eval "$PKG_INSTALL $pkg"
+            fi
+        else
+            if ! rpm -q "$pkg" > /dev/null 2>&1; then
+                echo "Running $PKG_UPDATE..."
+                eval "$PKG_UPDATE"
+                eval "$PKG_INSTALL $pkg"
+            fi
```
AlvaroRausell: We could simplify this by setting a variable like `$PKG_CHECK` that gets set to `apk info -e`, `rpm -q "$pkg"` etc based on the package manager. That way we avoid having an if-else chain again

src/anaconda/install.sh
```diff
@@ -66,15 +123,101 @@ updaterc() {
...
+    local dir="$1"
+    for item in "$dir"/*; do
+        if [ -e "$item" ]; then
+            if [ -d "$item" ]; then
+                chmod g+s "$item"
+                set_directory_permissions "$item"
+            fi
+        fi
+    done
+}
```
AlvaroRausell: Why not just use `chmod -R`?

src/anaconda/install.sh
```diff
@@ -83,30 +226,66 @@ if ! conda --version &> /dev/null ; then
...
+    # Set setgid bit on all directories - use find+xargs if available, fallback to recursive function
+    if command -v find > /dev/null && command -v xargs > /dev/null; then
+        find "${CONDA_DIR}" -type d -print0 | xargs -n 1 -0 chmod g+s
+    else
+        # Fallback for systems without find or xargs
+        if [ -d "${CONDA_DIR}" ]; then
+            chmod g+s "${CONDA_DIR}"
+            set_directory_permissions "${CONDA_DIR}"
+        fi
+    fi
```
AlvaroRausell: Why not simplify to:
```suggestion
chmod -R <perms> ${CONDA_DIR}
```

src/anaconda/install.sh
```diff
@@ -83,30 +226,66 @@ if ! conda --version &> /dev/null ; then
...
+
+    # Temporary fixes
+    # Due to https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23491
+    install_user_package certifi
+    # Due to https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-0286 and https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-23931
+    install_user_package pyopenssl
+    install_user_package cryptography
+    # Due to https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-40897
+    install_user_package setuptools
+    install_user_package tornado
```
AlvaroRausell: What is the issue, do old versions get installed by default?

src/anaconda/install.sh
```diff
@@ -135,7 +314,8 @@ if [ -f "/etc/bash.bashrc" ]; then
     echo "${notice_script}" | tee -a /etc/bash.bashrc
 fi
 
-# Clean up
-rm -rf /var/lib/apt/lists/*
+# Final clean up
+eval "$PKG_CLEAN"
+rm -rf "$PKG_LISTS"
```
AlvaroRausell: No need to delete if you are running clean

src/anaconda/README.md
```diff
@@ -35,7 +35,8 @@ conda install python=3.7
 ## OS Support
 
 This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
-
+Also RHEL based linux distributions such as almalinux, rockylinux, fedora are supported now.
+Please do note that Alpine and cbl-mariner aren't supported due system level restrictions with the anaconda installer in alpine linux and `groupadd`, `usermod`, `awk` commands not being supported in mariner.
```
AlvaroRausell: I would turn this into a bullet-point list, just add a note about alpine and mariner, but not going into too much detail

src/anaconda/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

I have changed this now.

src/anaconda/devcontainer-feature.json
v-Kaniska244: Done.

src/anaconda/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

This is corrected now. Kindly let me know in case of further concerns.

src/anaconda/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

This doesn’t appeared to be required anymore. The permissions were already provided upfront to the `$CONDA_DIR` and it doesn't appear to be necessary to set the setgid bit on all directories recursively inside `$CONDA_DIR` as the tests work fine without that as its able to launch conda environment seamlessly  with the `remoteUser` as `vscode`. My mistake, not removing this earlier.

src/anaconda/install.sh
v-Kaniska244: Same as the previous comment. This is actually not required. I have removed it.

src/anaconda/install.sh
v-Kaniska244: Done.

src/anaconda/README.md
v-Kaniska244: I have made some changes and updated `Notes.md` as well. Kindly let me know in case of further corrections required.

src/anaconda/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

The default anaconda package has older versions of these tools. In fact in the conda feature which is a minimal version of anaconda, actually have a similar implementation in the [install.sh](https://github.com/devcontainers/features/blob/main/src/conda/install.sh#L117-L123) script. Kindly let me know if change is required.

v-Kaniska244:
Hello @AlvaroRausell ,

Thank you for the detailed review. I have done the corrections. Would you kindly have another look and let me know in case of any further corrections.

With Regards,
Kaniska

src/anaconda/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

To add on this further, we actually don't need these to be manually installed any longer. The tools are present and up to date in the anaconda package itself. So i have removed them.

PR #1360, opened 2025-05-15
[aws-cli] Change dependency from gnupg2 to gpg

eszense:
Currently the aws-cli feature installs `gnupg2` for verifying aws-cli signature
However, `gnupg2` package install a few other utilities besides`gpg`. Only `gpg` package is required for signature verification, the other components are reductant.

Using `gpg` as the dependency improves install speed, save disk space for automated workflow, and more importantly reduces the attack surface of minimal systems.

### DIFF
--- a/src/aws-cli/devcontainer-feature.json
+++ b/src/aws-cli/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "aws-cli",
-    "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


Mathiyarasy:
Could you please update the feature version?

eszense:

AlvaroRausell (approved):

PR #1374, opened 2025-06-02
Check for a .nvmrc or .node-version file to determine the Node.js version

Mathiyarasy:
**Feature Name**

- node
  
**Description of Changes**

  - Fixes Issue: https://github.com/devcontainers/internal/issues/268  

**Changelog**

- Added a new version option, 'project-file', this checks for a .nvmrc or .node-version file to determine the Node.js version.
- Updated the install.sh script to retrieve the Node.js version from these files if they exist and are not empty.
- If 'project-file' is specified but neither file exists or both are empty, the script installs the Node.js 'lts' version by default.
- Added new test cases
- Replaced the hardcoded 'lts' Node.js version in the existing test case with logic to fetch the version from the Node.js URL


test/node/scenarios.json
```diff
@@ -199,5 +199,35 @@
...
+    },
+    "test_node_project_nvm_nodev": {
+        "build": {
+            "dockerfile": "DockerFile"
+        },
+        "features": {
+            "node": {
+                "version": "project-file"
+            }
+        }
```
v-Kaniska244, line 231: Any specific reason why we are adding these test cases such that the container is built from Dockerfile only & not using the image tag directly?

v-Mathiyarasy: Yes @Kaniska244 
Using Docker File I can include the .nvmrc (or .node-version) file in the image during the build process which is then gets used in feature.

src/node/install.sh
```diff
@@ -287,6 +286,34 @@ if ! type git > /dev/null 2>&1; then
     check_packages git
 fi
 
+# Determine the Node.js version using the .nvmrc or .node-version file if present.
+if [[ "${NODE_VERSION}" == "project-file" ]]; then
+    echo "Finding Node version from .nvmrc or .node-version file..."
+    NODE_VERSION_PATH=$(find . -type f -name ".node-version" | head -n 1)
```
AlvaroRausell: Not sure `find` is the best option. Can we not assume the files need to be in the root directory? What if the project is huge, `find` would take a long time. if there are multiple of these files, `find` might order them differently based on where it is running, so something like:
```
  for version_file in ".node-version" ".nvmrc"; do
        if [[ -f "$version_file" && -s "$version_file" ]]; then
            file_version=$(tr -d '[:space:]' < "$version_file")
            if [[ -n "$file_version" ]]; then
                echo "Using Node version from $version_file: $file_version"
                NODE_VERSION="$file_version"
                break
            else
                echo "$version_file exists but contains only whitespace. Continuing search..."
            fi
        fi
    done
```
might work more predictably

AlvaroRausell (changes requested):

Issue #1359, opened 2025-05-08
Rust feature does not work on non-Debian system


elibroftw:
0.527 ls: cannot access '/var/lib/apt/lists/': No such file or directory
0.683 find: '/var/lib/apt/lists/*': No such file or directory
0.685 Running apt-get update...
0.690 ./install.sh: line 123: apt-get: command not found
0.691 ERROR: Feature "Rust" (ghcr.io/devcontainers/features/rust) failed to inst
all! Look at the documentation at https://github.com/devcontainers/features/tree
/main/src/rust for help troubleshooting this error.

Kaniska244:
Hello @elibroftw ,

Thank you for reporting this issue. As of the now rust feature is only supported with debian/ ubuntu OS with apt package manager, [ref](https://github.com/devcontainers/features/tree/main/src/rust#os-support). Would you kindly let me know which platform you are using?

With Regards,
Kaniska

elibroftw:
AlmaLinux. I created my own rust feature fyi.

v-Kaniska244:
PR has been merged. Closing this issue, kindly report in case of further issues.

PR #1376, opened 2025-06-05
[rust] - Extending the feature support for non-debian platforms.

Kaniska244:
**Ref:** #1359 

**Description:** This PR aims to solve https://github.com/devcontainers/features/issues/1359 to extend rust installation support to non-debian OS systems.

- The installation script is changed to extend support for different package managers other than `apt`.
- Test scripts for different non-debian base images.
- The readme file modifed with the added information.
- Version bump.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/rust/README.md
+++ b/src/rust/README.md
@@ -31,7 +31,10 @@ Installs Rust, common Rust utilities, and their required dependencies
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
+This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, Alma, RockyLinux 
+and Mariner distributions with the `apt`, `yum`, `dnf`, `microdnf` and `tdnf` package manager installed.
+Alpine is not supported because the rustup-init binary requires glibc to run, but Alpine Linux does not include `glibc` 
+by default. Instead, it uses musl libc, which is not binary-compatible with glibc.
 
 `bash` is required to execute the `install.sh` script.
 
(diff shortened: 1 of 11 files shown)


Mathiyarasy (approved):

src/rust/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.3.3",
+    "version": "1.3.4",
```
AlvaroRausell: Can you bump the minor version instead? (i.e. to 1.4.0)

src/rust/README.md
```diff
@@ -31,7 +31,10 @@ Installs Rust, common Rust utilities, and their required dependencies
 
 ## OS Support
 
-This Feature should work on recent versions of Debian/Ubuntu-based distributions with the `apt` package manager installed.
+This Feature should work on recent versions of Debian/Ubuntu, RedHat Enterprise Linux, Fedora, Alma, RockyLinux 
+and Mariner distributions with the `apt`, `yum`, `dnf`, `microdnf` and `tdnf` package manager installed.
+Alpine is not supported because the rustup-init binary requires glibc to run, but Alpine Linux does not include `glibc` 
+by default. Instead, it uses musl libc, which is not binary-compatible with glibc.
```
AlvaroRausell:
```suggestion


**Note:** Alpine is not supported because the rustup-init binary requires glibc to run, but Alpine Linux does not include `glibc` 
by default. Instead, it uses musl libc, which is not binary-compatible with glibc.
```

src/rust/devcontainer-feature.json
v-Kaniska244: Hello @AlvaroRausell ,

I have updated it now.

src/rust/README.md
v-Kaniska244: Hello @AlvaroRausell ,

I have updated this now.

AlvaroRausell (approved):

Issue #1379, opened 2025-06-09
`git-lfs` not actually running `git lfs install`


andreilgeorgescu:
The `git-lfs` feature successfully downloads onto my codespace, however it is not actually running `git lfs install` as is expected. This also makes the `autoPull` option non-functioning.

I've tried many permutations but here is an example of my image/feature combination.
```json
{
  "image": "mcr.microsoft.com/devcontainers/base:2-ubuntu-24.04",
  "features": {
    "ghcr.io/devcontainers/features/git-lfs:1": {
      "autoPull": false // Doesn't change anything even if I make it `true` because `git-lfs` isn't installed
    }
  }
}
```

Running `git lfs pull` manually in the terminal results in the following:

```text
@andreilgeorgescu ➜ /workspaces/base (main) $ git lfs pull
Skipping object checkout, Git LFS is not installed for this repository.
Consider installing it with 'git lfs install'.
```

The workaround I have found is to add the following commands to my codespace:

```json
  "onCreateCommand": "git lfs install",
  "postCreateCommand": "git lfs pull",
```

However, it would be much better if the extension could handle this itself.

PR #1381, opened 2025-06-11
Fix git-lfs not actually running git lfs install

Mathiyarasy:
**Feature Name**

- git-lfs
  
**Description of Changes**

  - Fixes Issue: https://github.com/devcontainers/features/issues/1379
  
**Changelog**

- git lfs install command is added

### DIFF
--- a/src/git-lfs/install.sh
+++ b/src/git-lfs/install.sh
@@ -270,7 +270,7 @@ if ! git lfs ls-files > /dev/null 2>&1; then
     echo "(!) Skipping automatic 'git lfs pull' because no git lfs files were detected"
     exit 0
 fi
-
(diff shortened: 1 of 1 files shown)


AlvaroRausell (changes requested):
Can we add a test for this feature? It would be good to ensure this is now fully working

v-Mathiyarasy:
> Can we add a test for this feature? It would be good to ensure this is now fully working



> Can we add a test for this feature? It would be good to ensure this is now fully working

There is an existing test case scenario, '**autoPullEnabled**', that validates this change.

This test case was failing before this update and only passes with the current modifications

AlvaroRausell (approved):

Issue #1393, opened 2025-06-21
[desktop-lite] broken for nonvc versions 1.3.0 and above


vangeest:
## Short problem description
desktop-lite doesn't work with noVNC versions above 1.2.0

## Observed behaviour for novnc 1.2.0

snippet from devcontainer.json that I use
```
	"image": "mcr.microsoft.com/devcontainers/universal:latest",
	"features": {
		"ghcr.io/devcontainers/features/desktop-lite:latest": {
			"version": "latest",
			"noVncVersion": "1.2.0", 
			"password": "secret",
			"webPort": "6080",
			"vncPort": "5901"
		}
```
1. When the container has booted, you see a port opened in vs-code.
2. Clicking on the link from port 6080 will open novnc and user can connect to the vnc-server

## Observed behaviour for novnc 1.3.0 and later

change 1.2.0 in the above devcontainer configuration to anything else (i tried 1.4.0 and 1.6.0)

1. When the container has booted, you see no port opened in vs-code.

## Expected behaviour
versions 1.3.0 and above are expected to work (start novnc) just like 1.2.0

## Probable cause

The desktop-lite install.sh depends on novnc*/utils/launch.sh

see https://github.com/devcontainers/features/blob/main/src/desktop-lite/install.sh

launch.sh is not available anymore as of novnc versions 1.3.0

see 
https://github.com/novnc/noVNC/releases
`The novnc_proxy command (earlier launch.sh) is now able to interface better towards Websockify and now has it's own man page.`

vangeest:
This issue is related to #910 (closed)

vangeest:
Fixed by #1395 
Thanx @Kaniska244 !

PR #1395, opened 2025-06-23
[desktop-lite] - Adding support for noVnc latest version in the feature

Kaniska244:
**Ref:** #1393 

**Description:** This PR aims to solve https://github.com/devcontainers/features/issues/1393 by adding support for latest `noVnc` version. All the `noVnc` versions post `1.2.0` are currently not working with the `desktop-lite` feature due to unavailability of `launch.sh` script which is responsible for launching the `noVnc` client. In these higher versions of `noVnc` client post `1.2.0`, the client is required to be started using `novnc_proxy` utility. Changes are done for adding that support & also ensured that support for the older `noVnc` clients(<=1.2.0) remain as-is. 

**Changlog:** The following changes are done.

- Changed the code in install.sh to add support for `noVnc` client versions higher that `1.2.0`
- Updated the `devcontainer-feature.json` to do version bump & pining the `noVncVersion` property to the latest `1.6.0` version.
- Added a test for the older versions(<=1.2.0) separately.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -394,9 +394,8 @@ fi
 
 # Spin up noVNC if installed and not running.
 if [ -d "/usr/local/novnc" ]; then
-    NOVNC_VERSION_INSTALLED=\$(ls /usr/local/novnc | grep -oP 'noVNC-\K[0-9.]+' | head -n1)
     if [ "\$(ps -ef | grep /usr/local/novnc/noVNC*/utils/launch.sh | grep -v grep)" = "" ] && [ "\$(ps -ef | grep /usr/local/novnc/noVNC*/utils/novnc_proxy | grep -v grep)" = "" ]; then
-        if version_gt "\$NOVNC_VERSION_INSTALLED" "1.2.0"; then
+        if version_gt "${NOVNC_VERSION}" "1.2.0"; then
             keepRunningInBackground "noVNC" sudoIf "/usr/local/novnc/noVNC*/utils/novnc_proxy --listen ${NOVNC_PORT} --vnc localhost:${VNC_PORT}"
             log "noVNC started with novnc_proxy."
         else
@@ -404,7 +403,7 @@ if [ -d "/usr/local/novnc" ]; then
             log "noVNC started with launch.sh."
         fi
     else
-        log "noVNC is already running or not installed."
+        log "noVNC is already."
     fi
 else
     log "noVNC is not installed."
--- a/src/desktop-lite/install.sh
+++ b/src/desktop-lite/install.sh
@@ -403,7 +403,7 @@ if [ -d "/usr/local/novnc" ]; then
             log "noVNC started with launch.sh."
         fi
     else
-        log "noVNC is already."
+        log "noVNC is already running."
     fi
 else
     log "noVNC is not installed."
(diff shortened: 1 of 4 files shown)


src/desktop-lite/install.sh
```diff
@@ -386,11 +393,21 @@ else
 fi
 
 # Spin up noVNC if installed and not running.
-if [ -d "/usr/local/novnc" ] && [ "\$(ps -ef | grep /usr/local/novnc/noVNC*/utils/launch.sh | grep -v grep)" = "" ]; then
-    keepRunningInBackground "noVNC" sudoIf "/usr/local/novnc/noVNC*/utils/launch.sh --listen ${NOVNC_PORT} --vnc localhost:${VNC_PORT}"
-    log "noVNC started."
+if [ -d "/usr/local/novnc" ]; then
+    NOVNC_VERSION_INSTALLED=\$(ls /usr/local/novnc | grep -oP 'noVNC-\K[0-9.]+' | head -n1)
+    if [ "\$(ps -ef | grep /usr/local/novnc/noVNC*/utils/launch.sh | grep -v grep)" = "" ] && [ "\$(ps -ef | grep /usr/local/novnc/noVNC*/utils/novnc_proxy | grep -v grep)" = "" ]; then
+        if version_gt "\$NOVNC_VERSION_INSTALLED" "1.2.0"; then
```
AlvaroRausell: Why is `$NOVNC_VERSION_INSTALLED` escaped in this case? Shouldn't we cant to expand the variable?

src/desktop-lite/install.sh
```diff
@@ -386,11 +393,21 @@ else
...
+            keepRunningInBackground "noVNC" sudoIf "/usr/local/novnc/noVNC*/utils/novnc_proxy --listen ${NOVNC_PORT} --vnc localhost:${VNC_PORT}"
+            log "noVNC started with novnc_proxy."
+        else
+            keepRunningInBackground "noVNC" sudoIf "/usr/local/novnc/noVNC*/utils/launch.sh --listen ${NOVNC_PORT} --vnc localhost:${VNC_PORT}"
+            log "noVNC started with launch.sh."
+        fi
+    else
+        log "noVNC is already running or not installed."
```
AlvaroRausell: In this case `noVNC` is already running, as opposed to not installed

AlvaroRausell (changes requested):

src/desktop-lite/install.sh
v-Kaniska244: Hi @AlvaroRausell ,

This part of the code is basically writing into a desktop-init.sh & thats's why need to escape that to have `$NOVNC_VERSION_INSTALLED` in the new script. Kindly let me know if any change needed specifically for this one..

v-Kaniska244: Hi @AlvaroRausell ,

Sorry for misunderstanding the review comment in the first place & thank you for pointing this out. I have corrected it now. Would you kindly review this.

src/desktop-lite/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

Thank you for pointing this out. I have corrected the same now.

Issue #1273, opened 2025-02-13
rust: install components


mattlafayette:
Similar to #1032 , it would be useful to allow Rust to install additional components, for instance installing coverage tools for fuzzers.

For each component, do:

`rustup  component add <component>`

v-Kaniska244:
Closing this issue as PR has been merged.

PR #1404, opened 2025-07-02
[rust] - Introducing new options to install components

Kaniska244:
**Ref:** #1273 

**Description:** This PR aims to solve https://github.com/devcontainers/features/issues/1273 to introduce new options `customComponents` and `components` to install rust components based on user input

- The installation script is changed to introduce new options `customComponents` and `components` (comma separated value for the components to installed ) to install rust components based on user input.
- Change in devcontainer-feature.json by adding the new options `customComponents` and `components` to install rust components based on user input.
- Test scripts for different components in different combinations
- The readme file modified with the added information.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/rust/devcontainer-feature.json
+++ b/src/rust/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.4.0",
+    "version": "1.4.1",
     "name": "Rust",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/rust",
     "description": "Installs Rust, common Rust utilities, and their required dependencies",
@@ -57,7 +57,23 @@
                 "armv7-unknown-linux-gnueabihf",
                 "x86_64-unknown-redox,x86_64-unknown-uefi"
             ]
-        }
+        },
+        "customComponents": {
+            "type": "boolean",
+            "default": false,
+            "description": "Use custom components list instead of default components (rust-analyzer, rust-src, rustfmt, clippy)."
+        },
+        "components": {
+            "type": "string",
+            "default": "",
+            "description": "Optional comma separated list of Rust components to install when customComponents is true.",
+            "proposals": [
+                "rust-analyzer,rust-src,rustfmt,clippy",
+                "rust-analyzer,rust-src",
+                "rustfmt,clippy,rust-docs",
+                "llvm-tools-preview,rust-src,rustfmt"
+            ]
+        }        
     },
     "customizations": {
         "vscode": {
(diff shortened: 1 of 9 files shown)


src/rust/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "rust",
-    "version": "1.4.0",
+    "version": "1.4.1",
```
AlvaroRausell: This will need a minor version update, to 1.5.0

AlvaroRausell (changes requested):
I am not the biggest fan of having two flags for one action (installing custom components). I would prefer if we did the following:
`create a components option, the default being `rust-analyzer,rust-src,rustfmt,clippy` (what currently gets installed), but allowing it to be overridden. This will then allow a user to modify this list if required, without changing default behaviour.
So, as a summary:
- Remove `customComponents`
- Bump minor version.

Once that is done, I will go through the changes again to give a more detailed review

src/rust/devcontainer-feature.json
v-Kaniska244: Hello @AlvaroRausell ,

Thank you for pointing this out. I have bumped it to `1.5.0` now.

v-Kaniska244:
> I am not the biggest fan of having two flags for one action (installing custom components). I would prefer if we did the following: `create a components option, the default being `rust-analyzer,rust-src,rustfmt,clippy` (what currently gets installed), but allowing it to be overridden. This will then allow a user to modify this list if required, without changing default behaviour. So, as a summary:
> 
> * Remove `customComponents`
> * Bump minor version.
> 
> Once that is done, I will go through the changes again to give a more detailed review

Hello @AlvaroRausell ,

Thank you for the review. I have done changes as per the comments provided. Would you kindly review the PR once again.

src/rust/devcontainer-feature.json
```diff
@@ -57,7 +57,18 @@
                 "armv7-unknown-linux-gnueabihf",
                 "x86_64-unknown-redox,x86_64-unknown-uefi"
             ]
-        }
+        },
+        "components": {
+            "type": "string",
+            "default": "",
+            "description": "Optional comma separated list of Rust components to install when customComponents is true.",
```
AlvaroRausell: Dont forget to update the default and remove the mention of `customComponents`!

AlvaroRausell (changes requested):
Good work, just one thing you missed

src/rust/devcontainer-feature.json
v-Kaniska244: Hello @AlvaroRausell ,

Thank you for pointing this out. I have corrected the same. Would you kindly check once.

PR #1414, opened 2025-07-18
Provide error response when failure occurs fetching Git tags

eimhin-rover:
For the past 2 days we have observed the git feature failing to install on random occasions with the following error:

```
10.36 Invalid git version: latest
10.36 ERROR: Feature "Git (from source)" (ghcr.io/devcontainers/features/git) failed to install! Look at the documentation at ********/devcontainers/features/tree/main/src/git for help troubleshooting this error.
```

From looking into the source of the git feature I was able to identify where this error is stemming from and found that the curl used to fetch git tags does not have proper error handling.

As such, I have updated the install.sh script to improve the error handling here so that we fail earlier and can see the actual response from the curl to understand what the issue is. 

My hunch is that there could be some kind of rate limiting happening so I'm curious also if anyone else has been seeing this issue.


AlvaroRausell (changes requested):
Hello! Thanks for the contribution. Could you bump the patch version [here](https://github.com/eimhin-rover/features/blob/809a7f1f76a2d6d25bd5c024a4ea3a87f8c6ea8f/src/git/devcontainer-feature.json#L3) please to 1.3.5?

eimhin-rover:
Hey @AlvaroRausell! I've updated the version and also changed my approach. It seems that `--fail-with-body` is not available on older versions of curl, so I've opted for a different approach to ensure backwards compatibility.

eimhin-rover:
Any updates on this @AlvaroRausell ?

eimhin-rover:
Thanks for the review @AlvaroRausell ! Do I need to take any further action to have this merged or will it be done at some stage by one of the admins?

AlvaroRausell:
@Kaniska244 Free free to bump this up to Sam/Abdu, I don't remember exactly the context as to why I did not merge this, but should be fine (also, hi again!). 
@eimhin-rover Sorry for the delay! I'm not at GitHub anymore, so I can't merge, sadly.

eimhin-rover:
Oh hey I completely forgot about this 😓 No problem thanks for taking care of this!

Issue #1418, opened 2025-07-22
No valid OpenPGP data found


cabbirolo:
Hello,

I've seen similar issues related to OpenPGP, and this is also happening using the terraform feature (ghcr.io/devcontainers/features/terraform:1) with the image mcr.microsoft.com/vscode/devcontainers/base:ubuntu

[2025-07-22T20:52:23.575Z] Processing triggers for man-db (2.12.0-4build2) ...
[2025-07-22T20:52:24.013Z] TERRAFORM_VERSION=1.12.2
[2025-07-22T20:52:24.331Z] TFLINT_VERSION=0.58.1
[2025-07-22T20:52:24.670Z] TERRAGRUNT_VERSION=0.83.2
[2025-07-22T20:52:24.672Z] Downloading terraform...
[2025-07-22T20:52:26.830Z] (*) Downloading GPG key...
[2025-07-22T20:52:27.690Z] gpg: no valid OpenPGP data found.
[2025-07-22T20:52:27.691Z] (*) Failed getting key, retrying in 10s...
[2025-07-22T20:52:37.695Z] (*) Downloading GPG key...
[2025-07-22T20:52:38.480Z] gpg: no valid OpenPGP data found.
[2025-07-22T20:52:38.481Z] (*) Failed getting key, retrying in 10s...
[2025-07-22T20:52:48.483Z] (*) Downloading GPG key...
[2025-07-22T20:52:49.275Z] gpg: no valid OpenPGP data found.
[2025-07-22T20:52:49.277Z] (*) Failed getting key, retrying in 10s...
[2025-07-22T20:52:59.279Z] (*) Downloading GPG key...
[2025-07-22T20:53:00.074Z] gpg: no valid OpenPGP data found.
[2025-07-22T20:53:00.075Z] (*) Failed getting key, retrying in 10s...
[2025-07-22T20:53:10.078Z] (*) Downloading GPG key...
[2025-07-22T20:53:10.867Z] gpg: no valid OpenPGP data found.
[2025-07-22T20:53:10.869Z] (*) Failed getting key, retrying in 10s...
[2025-07-22T20:53:20.872Z] (*) Resolving GPG keyserver IP address...
[2025-07-22T20:53:20.903Z] (*) GPG keyserver IP address 185.125.188.27
(*) Downloading GPG key...
[2025-07-22T20:53:20.926Z] gpg: keyserver receive failed: No keyserver available
[...]

PR #1421, opened 2025-07-24
[terraform] - Fix terraform installation in ubuntu noble.

Kaniska244:
**Ref:** #1418 

**Description:** This PR aims to solve https://github.com/devcontainers/features/issues/1418 to solve broken terraform installation in ubuntu noble (24.04). The issue is happening in ubuntu noble platform while installing terraform as noble has stricter GPG validation and different default keyserver behavior compared to jammy/focal due to changes in how GPG keys are handled in Ubuntu Noble (24.04) and newer versions have deprecated the use of apt-key and require keys to be in a specific format (usually .gpg or .asc) and placed in `/etc/apt/keyrings` folder. But as the given approach where in the feature the binary is downloaded from hashicorp release site verified against the GPG key, as part of this PR added alternate source for the key & imported the same for Ubuntu noble (24.04).

- The installation script is changed to add alternate source for the GPG key & imported the same for Ubuntu noble.
- Change in devcontainer-feature.json to do version bump
- Test scripts for relevant options with Ubuntu noble base image.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/terraform/install.sh
+++ b/src/terraform/install.sh
@@ -113,7 +113,7 @@ receive_gpg_keys() {
     
     # Special handling for HashiCorp GPG key on Ubuntu Noble
     if [ "$IS_NOBLE" -eq 1 ] && [ "$keys" = "$TERRAFORM_GPG_KEY" ]; then
-        echo "(*) Ubuntu Noble detected, using Keybase for HashiCorp GPG key import..."
+        echo "(*) Ubuntu Noble detected, using Keybase for HashiCorp GPG key import...."
         curl -fsSL https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
         if ! gpg --list-keys "${TERRAFORM_GPG_KEY}" > /dev/null 2>&1; then
             gpg --list-keys


src/terraform/install.sh
```diff
@@ -50,6 +50,24 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+import_hashicorp_gpg_key_noble() {
+    unset GNUPGHOME
+    curl -fsSL https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
+    if ! gpg --list-keys "${TERRAFORM_GPG_KEY}" > /dev/null 2>&1; then
+        gpg --list-keys
+        echo "(!) Warning: HashiCorp GPG key not found in keyring after import."
```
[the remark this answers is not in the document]
v-Kaniska244: I don't it would be right to exit if the curl fails here, given that I have a fallback mechanism later to call `verify_terraform_sig` if `gpg --verify` fails with this key, so this will be handled there.

src/terraform/install.sh
```diff
@@ -366,6 +384,13 @@ install_terraform() {
     curl -sSL -o ${terraform_filename} "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/${terraform_filename}"
 }
 
+verify_terraform_sig() {
+    receive_gpg_keys TERRAFORM_GPG_KEY
+    curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS" 
+    curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
+    gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS
```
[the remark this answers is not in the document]
v-Kaniska244: Here if you look closely verify_terraform_sig and verify_sentinel_sig have slight differences. So can't merge them like this.

src/terraform/install.sh
```diff
@@ -378,10 +403,18 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
...
     if [ "${TERRAFORM_SHA256}" = "automatic" ]; then
-        receive_gpg_keys TERRAFORM_GPG_KEY
-        curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS" 
-        curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
-        gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS
+        if [ "$IS_NOBLE" -eq 1 ]; then
+            import_hashicorp_gpg_key_noble
+            curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS"
+            curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
+            gpg --list-keys
```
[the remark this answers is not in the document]
v-Kaniska244: There is no variable called `DEBUG` present here. This will fail.

src/terraform/install.sh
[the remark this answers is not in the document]
v-Kaniska244: I have implemented this but made some change in the error messages to make them more relevant..

src/terraform/install.sh
```diff
@@ -378,10 +418,18 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
...
+            if ! gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS; then
+                echo "Primary GPG verification failed, attempting fallback verification..."
+                verify_terraform_sig
+            fi
+        else
+            verify_terraform_sig
+        fi
```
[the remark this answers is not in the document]
v-Kaniska244: Not needed.

src/terraform/install.sh
```diff
@@ -378,10 +418,18 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
...
     if [ "${TERRAFORM_SHA256}" = "automatic" ]; then
-        receive_gpg_keys TERRAFORM_GPG_KEY
-        curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS" 
-        curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
-        gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS
+        if [ "$IS_NOBLE" -eq 1 ]; then
+            import_hashicorp_gpg_key_noble
+            curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS"
+            curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
+            gpg --list-keys
```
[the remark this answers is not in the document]
v-Kaniska244: I would rather keep it for the moment to ensure that sufficient details available in case of gpg verification failure.

src/terraform/install.sh
```diff
@@ -477,12 +531,21 @@ if [ "${INSTALL_SENTINEL}" = "true" ]; then
...
+                    echo "Primary GPG verification failed, attempting fallback verification..."
+                    verify_sentinel_sig
+                fi
+                # Verify the SHASUM matches the archive
+                shasum -a 256 --ignore-missing -c sentinel_checksums.txt
+            else
+                verify_sentinel_sig
+                # Verify the SHASUM matches the archive
+                shasum -a 256 --ignore-missing -c sentinel_checksums.txt
+            fi
```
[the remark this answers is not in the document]
v-Kaniska244: There's no primary verification, only difference is that in the 'if part' the key the gpg key is imported differently for ubuntu noble. This comment isn't valid.

src/terraform/install.sh
```diff
@@ -378,10 +418,18 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
...
-        curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS" 
-        curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
-        gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS
+        if [ "$IS_NOBLE" -eq 1 ]; then
+            import_hashicorp_gpg_key_noble
+            curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS"
+            curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
+            gpg --list-keys
+            if ! gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS; then
+                echo "Primary GPG verification failed, attempting fallback verification..."
```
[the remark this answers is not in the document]
v-Kaniska244: This is not valid. I want to try the existing method of gpg validation in ubuntu noble as it still works in some cases in case of a failure with the new implementation.

src/terraform/install.sh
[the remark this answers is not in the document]
v-Kaniska244: As mentioned earlier, I would rather keep it as it can help in debugging in case of failure and as a matter of fact we don't have different levels of logging implemented in this script, so can't be bypassed in that manner either.

src/terraform/install.sh
```diff
@@ -378,10 +418,18 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
...
+                verify_terraform_sig
+            fi
```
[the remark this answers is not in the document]
v-Kaniska244: As mentioned above, this is not needed right now.

src/terraform/install.sh
```diff
@@ -50,6 +50,25 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+import_hashicorp_gpg_key_noble() {
```
dmnmenace, line 53: gh pr checkout 522

src/terraform/install.sh
```diff
@@ -366,6 +387,49 @@ install_terraform() {
...
+    local sha256sums_url=$2
+    local sig_url=$3
+    local sha256sums_file=$4
+    local sig_file=$5
+
+    receive_gpg_keys "$gpg_key"
+    curl -sSL -o "$sha256sums_file" "$sha256sums_url"
+    curl -sSL -o "$sig_file" "$sig_url"
+    gpg --verify "$sig_file" "$sha256sums_file"
+}
```
AlvaroRausell: Can you merge `verify_sigature` and `verify_with_fallback` and the fallback fn is an optional param at the end so that, if it fails _and_ there if a fallback fn, it calls it? That way we avoid having two fns that do virtually the same thing

src/terraform/install.sh
```diff
@@ -50,6 +50,27 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
+import_hashicorp_gpg_key_noble() {
+    unset GNUPGHOME
```
AlvaroRausell: Why did you have to unset this? Maybe add a comment to explain it

src/terraform/install.sh
```diff
@@ -50,6 +50,27 @@ if [ "$(id -u)" -ne 0 ]; then
...
+    curl -fsSL https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
```
AlvaroRausell: We seem to be importing the keys here, but also in `receive_gpg_keys`, which gets called in `verify_signature`. Can we clean up the logic so this import is done at that stage instead of the default behaviour?

src/terraform/install.sh
```diff
@@ -378,10 +442,17 @@ if grep -q "The specified key does not exist." "${terraform_filename}"; then
 fi
 if [ "${TERRAFORM_SHA256}" != "dev-mode" ]; then
     if [ "${TERRAFORM_SHA256}" = "automatic" ]; then
-        receive_gpg_keys TERRAFORM_GPG_KEY
-        curl -sSL -o terraform_SHA256SUMS "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS" 
-        curl -sSL -o terraform_SHA256SUMS.sig "${HASHICORP_RELEASES_URL}/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_SHA256SUMS.${TERRAFORM_GPG_KEY}.sig"
-        gpg --verify terraform_SHA256SUMS.sig terraform_SHA256SUMS
+        if [ "$IS_NOBLE" -eq 1 ]; then
+            verify_with_fallback \
```
AlvaroRausell: After merging the functions this wont be the case as much, but I feel like if we only run this function in noble, we can simplify the params. Lots of parameters in a bash function is confusing and hard to read

AlvaroRausell (changes requested):

AlvaroRausell:
As a side note, is there a way to bypass the key verification? If the key ever went away, we might want to allow the user to be able to get running without it

AlvaroRausell:
And lastly, would [this](https://apt.releases.hashicorp.com/gpg) be an alternative URL for the GPG key?

v-Kaniska244:
> And lastly, would [this](https://apt.releases.hashicorp.com/gpg) be an alternative URL for the GPG key?

Hello @AlvaroRausell ,

They don't appear to be the same key. so won’t work.

```
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ curl -fsSL https://keybase.io/hashicorp/pgp_keys.asc | gpg --import
gpg: key 34365D9472D7468F: public key "HashiCorp Security (hashicorp.com/security) <security@hashicorp.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ 
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ gpg --list-keys
/home/node/.gnupg/pubring.kbx
-----------------------------
pub   rsa4096 2021-04-19 [SC] [expires: 2026-04-18]
      C874011F0AB405110D02105534365D9472D7468F
uid           [ unknown] HashiCorp Security (hashicorp.com/security) <security@hashicorp.com>
sub   rsa4096 2021-04-19 [E] [expires: 2026-04-18]
sub   rsa4096 2021-04-21 [S] [expires: 2026-04-20]

@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ 
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ gpg --delete-keys C874011F0AB405110D02105534365D9472D7468F
gpg (GnuPG) 2.2.40; Copyright (C) 2022 g10 Code GmbH
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.


pub  rsa4096/34365D9472D7468F 2021-04-19 HashiCorp Security (hashicorp.com/security) <security@hashicorp.com>

Delete this key from the keyring? (y/N) y
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ 
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ gpg --list-keys
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ 
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ curl -fsSL https://apt.releases.hashicorp.com/gpg | gpg --import
gpg: key AA16FCBCA621E701: public key "HashiCorp Security (HashiCorp Package Signing) <security+packaging@hashicorp.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ gpg --list-keys
/home/node/.gnupg/pubring.kbx
-----------------------------
pub   rsa4096 2023-01-10 [SC] [expires: 2028-01-09]
      798AEC654E5C15428C8E42EEAA16FCBCA621E701
uid           [ unknown] HashiCorp Security (HashiCorp Package Signing) <security+packaging@hashicorp.com>
sub   rsa4096 2023-01-10 [S] [expires: 2028-01-09]

@Kaniska244 ➜ /workspaces/features (gpg_fix_ubuntu_noble) $ 
```

v-Kaniska244:
> As a side note, is there a way to bypass the key verification? If the key ever went away, we might want to allow the user to be able to get running without it

Hello @AlvaroRausell ,

I think it should work but I will test and confirm. 

Edit :- I have verified the same, we can proceed with installation with basic checksum validation alone. So now I have kept it as a fallback mechanism for ubuntu noble, if the key download or verification fails.

src/terraform/install.sh
v-Kaniska244: The `verify_with_fallback` function is not needed now. So removed the same.

src/terraform/install.sh
v-Kaniska244: It was not required, removed it.

src/terraform/install.sh
v-Kaniska244: Hello @AlvaroRausell ,

I have merged this logic in `import_hashicorp_gpg_key_noble` into the `receive_gpg_keys` function now.

src/terraform/install.sh
v-Kaniska244: This `verify_with_fallback` function is removed now.

v-Kaniska244:
Hello @AlvaroRausell ,

Thank you for the detailed review and the suggestions. I have corrected the implementation accordingly. Would you kindly have another look and let me know in case of any further concern.

With Regards,
Kaniska

src/terraform/install.sh
```diff
@@ -89,7 +98,7 @@ receive_gpg_keys() {
         keyring_args="--no-default-keyring --keyring $2"
     fi
     if [ ! -z "${KEYSERVER_PROXY}" ]; then
-	keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
+    keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
```
AlvaroRausell:
```suggestion
        keyring_args="${keyring_args} --keyserver-options http-proxy=${KEYSERVER_PROXY}"
```

src/terraform/install.sh
```diff
@@ -366,6 +390,36 @@ install_terraform() {
...
+    curl -sSL -o "$sha256sums_file" "$sha256sums_url"
+    curl -sSL -o "$sig_file" "$sig_url"
+
+    # Try GPG verification, but don't fail on Noble
+    gpg --verify "$sig_file" "$sha256sums_file"
+    verify_result=$?
+    if [ $verify_result -ne 0 ] && [ "$IS_NOBLE" -eq 1 ]; then
+        echo "(*) Warning: GPG verification failed on Ubuntu Noble. Continuing installation anyway."
+        echo "    This is expected behavior due to known keyserver issues on Ubuntu Noble."
+        return 1
```
AlvaroRausell: Should this not return 0?

v-Kaniska244: Hello @AlvaroRausell ,

Thank you for pointing this out. Should we really skip error in this part given the correct gpg key is available? In that case case it could imply that the download package is a vulnerable one as even with the valid gpg key the for ubuntu noble the download package verification failed. I think here we should exit straightaway even for ubuntu noble.

v-Kaniska244:
Hello @AlvaroRausell ,

I have slightly changed the error handling for the part of the code when gpg validation of the download package fails even with the correct key which you had highlighted in the last review. Would you kindly have a look and let me know in case of any further concern.

DariuszPorowski (approved):

sethbacon:
Hello @AlvaroRausell,
Any chance you can review this for @Kaniska244 ? Waiting on this PR to hit...
Thanks

src/terraform/install.sh
AlvaroRausell: In that case, I would change the message, as `Continuing installation anyway.` will imply that we are proceeding as usual.

AlvaroRausell (approved):

PR #1426, opened 2025-07-30
[rust] Update NOTES.md

olivierlemasle:
`README.md` was updated in #1376, but `NOTES.md` was left as is. The changes in `README.md` will be overriden when CI will update the documentation (e.g. with #1408).

This PR also regenerates `README.md`, which fixes the documentation for option `components`.


olivierlemasle:

PR #1430, opened 2025-08-04
If group userGid exists, use its name

### LINKED ISSUE #1429: common-utils fails for existing user with arbitrary group
## Short description

Trying to use the `common-utils` feature with a pre-existing user fails if there is **no group with user's name**.

### `Dockerfile`

```Dockerfile
FROM ubuntu:noble

RUN useradd --gid ubuntu --create-home someuser
USER someuser
```

### `devcontainer.json`

```json
{
  "build": { "dockerfile": "Dockerfile" },
  "features": {
    "ghcr.io/devcontainers/features/common-utils": {
      "username": "someuser",
      "userGid": "1000",
      "userUid": "1001"
    }
  }
}
```

Start the Dev Container.

### Expected
Dev Container starts with the given user, with zsh being the default shell.

### Actual behavior
The container build fails during the installation of the `common-utils` feature. Relevant log lines:

```
46.19 chown: invalid group: 'someuser:someuser'
46.19 ERROR: Feature "Common Utilities" (ghcr.io/devcontainers/features/common-utils) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/common-utils for help troubleshooting this error.
```

sleiner:
Fixes #1429


sleiner:
There should probably be a test for this scenario - I can add one if anybody can tell me how to to do it :-)

src/common-utils/main.sh
```diff
@@ -411,12 +411,16 @@ elif [ "${USERNAME}" = "none" ]; then
     USER_UID=0
     USER_GID=0
 fi
+
 # Create or update a non-root user to match UID/GID.
-group_name="${USERNAME}"
+group_name=$(getent group "$USER_GID" | cut -d: -f3)
```
v-Kaniska244, line 416: Hello @sleiner 

I am not sure if this would really solve the problem you had encountered as you have mentioned. It would probably fail while trying to create a new group for user `someone` as the gid will be passed on as `1000` which is already existing.

src/common-utils/main.sh
v-Kaniska244, line 1: Hello @sleiner ,

Thank you for the contribution. Please describe the fix in a detailed manner. Also add test [here](https://github.com/devcontainers/features/blob/main/test/common-utils/scenarios.json) & to test you can use `devcontainer features test` command locally in the vscode terminal. Also please do a version bump

Kaniska244 (changes requested):

PR #1441, opened 2025-08-12
[rust] Add Alpine support for rust feature

olivierlemasle:
`musl` is supported by Rustup; it is just required to download the right version of `rustup-init`.

### DIFF
--- a/test/rust/scenarios.json
+++ b/test/rust/scenarios.json
@@ -64,14 +64,14 @@
                 "components": ""
             }
         }
-    },    
+    },
     "rust_with_centos": {
         "image": "centos:centos7",
         "features": {
             "rust": {
                 "version": "latest",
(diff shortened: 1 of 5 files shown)


test/rust/scenarios.json
v-Kaniska244, line 1: Could you please also add a test case for alpine 3.22?

olivierlemasle: - `mcr.microsoft.com/devcontainers/base:alpine-3.22` has not yet been released.
- `alpine:3.22` is released but does not include `bash`, which is required for this feature.

What do you suggest, please?

olivierlemasle: As `mcr.microsoft.com/devcontainers/base:alpine-3.22` has now been released, I've updated the PR accordingly.

PR #1464, opened 2025-09-03
[sshd] Add GatewayPorts option

lzhou-lilly:
We want to optionlly enable the GatewayPorts option in sshd.

To test: `devcontainer features test --features sshd`

Background:
Our development setup runs in Codespaces with Docker Compose. To handle TLS termination and spoof internal IPs, we use a reverse tunnel via `gh cs ssh -- -R` and proxys on workstations.

Without GatewayPorts, only the main container can access the tunnel—other containers can’t. Enabling GatewayPorts allows the tunnel to bind to the container’s public interfaces too, so other containers can use it. This is required for our transparent proxy setup.

### DIFF
--- a/src/sshd/README.md
+++ b/src/sshd/README.md
@@ -15,6 +15,7 @@ Adds a SSH server into a container so that you can use an external terminal, sft
 
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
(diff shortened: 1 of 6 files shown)


lzhou-lilly:

This is a trivial work I initiated and completed on my own accord.

Kaniska244 (approved):
Hi @lzhou-lilly 

Thank you for the contribution. The change looks fine. To be further reviewed by maintainers.

Issue #1485, opened 2025-09-19
[Trixie] The repository 'https://packages.microsoft.com/repos/azure-cli trixie Release' does not have a Release file.


erwinkramer:
I'm using this devcontainer:

```jsonc
// For format details, see https://aka.ms/devcontainer.json. For config options, see the
// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
{
	"name": "Bank API dotnet10",
	"image": "mcr.microsoft.com/devcontainers/dotnet:dev-10.0-preview-trixie", // https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/dotnet/tags
	"customizations": {
		"vscode": {
			"extensions": [
				"humao.rest-client",
				"ms-dotnettools.csdevkit",
				"ms-graph.kiota",
				"stoplight.spectral",
				"streetsidesoftware.code-spell-checker",
				"github.copilot"
			],
			"settings": {
				"terminal.integrated.shell.linux": "/bin/bash",
				"dotnetAcquisitionExtension.sharedExistingDotnetPath": "/usr/share/dotnet/dotnet"
			}
		}
	},
	"containerEnv": {
		"DOTNET_LAUNCH_PROFILE": "http",
		// Zscaler proxy support when reading custom rulesets over https in Spectral
		"NODE_EXTRA_CA_CERTS": "${containerWorkspaceFolder}/.certs/ZscalerRootCA.crt"
	},
	// Features to add to the dev container. More info: https://containers.dev/features.
	"features": {
		"ghcr.io/devcontainers/features/dotnet:2": {
			"version": "none", // already pre-installed with image
			"aspNetCoreRuntimeVersions": "8.0" // for: https://github.com/microsoft/vscode-dotnettools/issues/1094
		},
		"ghcr.io/devcontainers/features/azure-cli:1": {}, // for using the host az cli credentials
		"ghcr.io/devcontainers/features/docker-in-docker:2": {
			"moby": false
		} // for .NET Aspire resource emulation
	},
	"mounts": [
		// Bind the azure (credential) folder, requires an unencrypted token cache (assuming host is Windows).
		// On your host, run `az config set core.encrypt_token_cache=false` first, login again, then start the container
		// Please read: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3033 for more information
		"source=${localEnv:HOME}${localEnv:USERPROFILE}/.azure,target=/home/vscode/.azure,type=bind",
[...]

PR #1489, opened 2025-10-03
Fix Azure-cli installation for trixie

Mathiyarasy:
**Fixes**: https://github.com/devcontainers/features/issues/1485

**Issue Summary:**
- The update for Azure CLI installation on Debian Trixie caused this issue since Microsoft does not offer packages for this distribution. Although the apt-based installation fails, Azure CLI still gets installed successfully using pip.
- However, after this change, apt-get commands fail because the broken repository file /**etc/apt/sources.list.d/azure-cli.list** is not being properly removed.
- This leftover file leads to **apt-get** errors during the installation of other features such as docker-in-docker and dotnet

**Solution:**
- If the apt-get update command fails, the script logs a message and removes the Azure CLI repository file etc/apt/sources.list.d/azure-cli.list

### DIFF
--- a/src/azure-cli/install.sh
+++ b/src/azure-cli/install.sh
@@ -98,10 +98,14 @@ install_using_apt() {
     # Import key safely (new 'signed-by' method rather than deprecated apt-key approach) and install
     curl -sSL ${MICROSOFT_GPG_KEYS_URI} | gpg --dearmor > /usr/share/keyrings/microsoft-archive-keyring.gpg
     echo "deb [arch=${architecture} signed-by=/usr/share/keyrings/microsoft-archive-keyring.gpg] https://packages.microsoft.com/repos/azure-cli/ ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/azure-cli.list
-
+    
+    # This is a workaround for the fact that Azure CLI is not yet a supported package for debian trixie. Once Azure CLI is supported we should revert to avoid script failure for non Azure CLI packages
     if ! (apt-get update); then
         echo "(!) Failed to update apt cache, removing repository file"
         rm -f /etc/apt/sources.list.d/azure-cli.list
+    else  
+        echo "WARNING: apt-get update succeeded. The workaround for broken Azure CLI install on Debian Trixie may no longer be needed."  
+        echo "TODO: Consider reverting to a simple `apt-get update` if the Azure CLI repo works reliably for debian trixie"  
     fi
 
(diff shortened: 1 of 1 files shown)


src/azure-cli/install.sh
```diff
@@ -98,7 +98,11 @@ install_using_apt() {
     # Import key safely (new 'signed-by' method rather than deprecated apt-key approach) and install
     curl -sSL ${MICROSOFT_GPG_KEYS_URI} | gpg --dearmor > /usr/share/keyrings/microsoft-archive-keyring.gpg
     echo "deb [arch=${architecture} signed-by=/usr/share/keyrings/microsoft-archive-keyring.gpg] https://packages.microsoft.com/repos/azure-cli/ ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/azure-cli.list
-    apt-get update
+
+    if ! (apt-get update); then
+        echo "(!) Failed to update apt cache, removing repository file"
+        rm -f /etc/apt/sources.list.d/azure-cli.list
+    fi
```
sam-byng, line 109: can you add an else statement here with 
```suggestion
    else
        echo "WARNING: apt-get update succeeded. The workaround for broken Azure CLI install on Debian Trixie may no longer be needed."
        echo "TODO: Fonsider reverting to a simple `apt-get update` if the Azure CLI repo works reliably for debian trixie"
    fi
```

sam-byng: otherwise we risk this script failing in future due to non azure cli package issue, after azure cli is fixed.

src/azure-cli/install.sh
```diff
@@ -98,7 +98,11 @@ install_using_apt() {
     # Import key safely (new 'signed-by' method rather than deprecated apt-key approach) and install
     curl -sSL ${MICROSOFT_GPG_KEYS_URI} | gpg --dearmor > /usr/share/keyrings/microsoft-archive-keyring.gpg
     echo "deb [arch=${architecture} signed-by=/usr/share/keyrings/microsoft-archive-keyring.gpg] https://packages.microsoft.com/repos/azure-cli/ ${VERSION_CODENAME} main" > /etc/apt/sources.list.d/azure-cli.list
-    apt-get update
+
+    if ! (apt-get update); then
```
sam-byng, line 103: Can you add a comment - "This is a workaround for the fact that Azure CLI is not yet a supported package for debian trixie. Once Azure CLI is supported we should revert to avoid script failure for non Azure CLI packages."

PR #1492, opened 2025-10-04
Use /etc/profile.d to set PATH for dotnet tools

sliekens:
Back in #628, I added `~/.dotnet/tools` to the `PATH` using a relative path in `containerEnv`, so dotnet global tools would work regardless of the container user.

While this worked for running dotnet global tools, it broke compatibility with `find -execdir`, which treats relative paths in `PATH` as unsafe.

This PR removes the relative path from `containerEnv` and instead adds a script to `/etc/profile.d` that appends `$HOME/.dotnet/tools` to `PATH` for login shells. This preserves global tool functionality without compromising security.

### DIFF
--- a/src/dotnet/devcontainer-feature.json
+++ b/src/dotnet/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "dotnet",
-    "version": "2.4.0",
+    "version": "2.4.1",
     "name": "Dotnet CLI",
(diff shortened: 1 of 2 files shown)


KyleGibsonlucky (approved):

sliekens:
@Kaniska244 who is currently the maintainer / code owner for the dotnet feature?

Issue #1473, opened 2025-09-08
Azure Linux OS support for docker-in-docker


krabinowitz1:
Just wondering if there are plans to add Azure Linux to the list of supported OS for the docker-in-docker feature? I don't see any mentions of it

v-Kaniska244:
The PR has been merged already. Kindly check and get back in case of any further issues.

PR #1506, opened 2025-10-22
Adding azure-linux OS support for docker-in-docker feature

sireeshajonnalagadda:
Feature:
Docker-in-docker

Description:

Solves [issue](https://github.com/devcontainers/features/issues/1473)
This is to add support of azureLinux OS for docker-in-docker feature

Changes:
Modified install script to work both ubuntu/debian and azureLinux 
Created scenarios to check for both moby true and false 

Checklist:
Changes worked as expected 

install.sh compatibility 
Confirmed that install.sh continues to work correctly with non–azure-linux base images, specifically Ubuntu and Debian.

Cross-distro testing 
Ran scenarios with Ubuntu, Debian, and Mariner base images to validate docker-in-docker functionality across multiple environments.

Azure Linux support 
Verified that after adding the implementation for the azure-linux distro, dev containers can be successfully built and run using docker-in-docker.

### DIFF
--- a/src/common-utils/devcontainer-feature.json
+++ b/src/common-utils/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "common-utils",
-    "version": "2.5.4",
+    "version": "2.5.5",
     "name": "Common Utilities",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/common-utils",
     "description": "Installs a set of common command line utilities, Oh My Zsh!, and sets up a non-root user.",
@@ -64,6 +64,11 @@
             "type": "boolean",
             "default": false,
             "description": "Add packages from non-free Debian repository? (Debian only)"
+        },
+        "installSsl": {
+            "type": "boolean",
+            "default": true,
+            "description": "Install SSL?"
         }
     }
 }
--- a/src/common-utils/install.sh
+++ b/src/common-utils/install.sh
@@ -18,6 +18,7 @@ USERNAME="${USERNAME:-"automatic"}"
 USER_UID="${UID:-"automatic"}"
 USER_GID="${GID:-"automatic"}"
 ADD_NON_FREE_PACKAGES="${NONFREEPACKAGES:-"false"}"
+INSTALL_SSL="${INSTALLSSL:-"true"}"
 
 MARKER_FILE="/usr/local/etc/vscode-dev-containers/common"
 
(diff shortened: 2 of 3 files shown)


src/docker-in-docker/install.sh
```diff
@@ -408,19 +744,32 @@ fallback_compose-switch() {
...
-        current_compose_path="$(which docker-compose)"
+        current_compose_path="$(command -v docker-compose)"
         target_compose_path="$(dirname "${current_compose_path}")/docker-compose-v1"
         compose_switch_version="latest"
         compose_switch_url="https://github.com/docker/compose-switch"
+        # Try to get latest version, fallback to known stable version if GitHub API fails
+        set +e
         find_version_from_git_tags compose_switch_version "$compose_switch_url"
-        curl -fsSL "https://github.com/docker/compose-switch/releases/download/v${compose_switch_version}/docker-compose-linux-${architecture}" -o /usr/local/bin/compose-switch || fallback_compose-switch "$compose_switch_url"
+        if [ $? -ne 0 ] || [ -z "${compose_switch_version}" ] || [ "${compose_switch_version}" = "latest" ]; then
```
Mathiyarasy: We already have a fallback version script :  fallback_compose-switch if this latest version fails.. 
Why do we need to hardcode version here?

src/docker-in-docker/install.sh
```diff
@@ -408,19 +744,32 @@ fallback_compose-switch() {
...
+            echo "(*) GitHub API rate limited or failed, using fallback version 1.0.5"
+            compose_switch_version="1.0.5"
```
v-Mathiyarasy: We already have a fallback version script : fallback_compose-switch if this latest version fails..
Could we use it instead of hardcoding version?

sireeshajonnalagadda: Modified to use fallback_compose-switch

Mathiyarasy (approved):

sam-byng (changes requested):
Hey @sireeshajonnalagadda - thanks for the PR :) 

In PR description: 

Could you list the testing / verification that has been performed to confirm that docker-in-docker feature is still working as expected after the addition of azure-linux OS support?

In particular:
- Is install.sh working with other base image types (not just azure-linux?)
- is azure-linux working? / can we build a dev container using it with this feature?

This seems like a big change to a fundamental technology (install script for docker-in-docker) so want to make sure we're doing the right thing here.

sireeshajonnalagadda:
Hi @sam-byng,
I have modified the description to make it clearer and more meaningful.

sam-byng:
Thanks - LGTM

Issue #1501, opened 2025-10-13
Cannot install latest python feature due to changed signature verification


Rinze-Smits:
Installing the Python feature using "version": "latest" fails on downloading the .asc signature. There is no more .asc signature.

```
13.22 (*) Building Python latest from source...
13.60 VERSION=3.14.0
13.60 Downloading Python-3.14.0.tgz...
23.29 (*) Keyserver hkp://keyserver.pgp.com is not reachable.
23.44 (*) Downloading GPG key...
24.95 Downloading Python-3.14.0.tgz.asc...
25.01 gpg: no valid OpenPGP data found.
25.01 gpg: the signature could not be verified.
25.01 Please remember that the signature file (.sig or .asc)
25.01 should be the first file given on the command line.
25.01 ERROR: Feature "Python" (ghcr.io/devcontainers/features/python) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/python for help troubleshooting this error.
```

I need to install a specific version of Python because of #1403 

The relevant info from the release notes is at https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-no-more-pgp

> ### Discontinuation of PGP signatures
> PGP (Pretty Good Privacy) signatures will not be provided for releases of Python 3.14 or future versions. To verify CPython artifacts, users must use [Sigstore verification materials](https://www.python.org/downloads/metadata/sigstore/). Releases have been signed using [Sigstore](https://www.sigstore.dev/) since Python 3.11.
> 
> This change in release process was specified in [PEP 761](https://peps.python.org/pep-0761/).

The new sigstore signatures are available for Python 3.11 and newer, so signature verification should have version detection for 3.11 and later or 3.14 and later for the new method, while still using the pgp method for older releases.

sireeshajonnalagadda:
@Rinze-Smits, 
The PR has been merged, and I hope you’re no longer experiencing this error. I’ll go ahead and close the issue as resolved, but please feel free to reopen it if you run into any further concerns.

PR #1508, opened 2025-10-27
Update Python Feature Installation: Support Sigstore Signatures for Python 3.14+ and Fix OpenSSL Build on CentOS 7/RHEL 7

Mathiyarasy:
**Fixes**: https://github.com/devcontainers/features/issues/1501

**Issue summary:**

- **PGP Signature Discontinuation:**
Python 3.14 and future releases no longer use PGP signatures [ref](https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-no-more-pgp)
CPython now adopts Sigstore for artifact verification

**Fix:**
- **Cosign Integration:** Updated the installation workflow to integrate Cosign for validating the signatures of Python versions 3.14 and newer
- **CentOS 7/RHEL 7 Compatibility issue**:
Python installation was failing on CentOS 7 due to an error encountered during the OpenSSL build process. 
Error: `"Can't locate Time/Piece.pm in @inc" ` when building OpenSSL 3.0.18+. OpenSSL 3.0.18+ requires the Perl Time::Piece module for its build system. To resolve this, the installation of the perl-Time-Piece package was added for CentOS 7 and RHEL 7,

### DIFF
--- a/src/python/install.sh
+++ b/src/python/install.sh
@@ -535,7 +535,7 @@ install_cosign() {
         echo -e "\n(!) Failed to fetch cosign v${COSIGN_VERSION}..."
         # Try previous version
         find_prev_version_from_git_tags COSIGN_VERSION "https://github.com/sigstore/cosign"
-        echo -e "\nAttempting to install ${COSIGN_VERSION}"
+        echo -e "\nAttempting to install previous cosign ${COSIGN_VERSION} version as fallback mechanism"
         
         version_for_url="${COSIGN_VERSION#v}"
         cosign_filename="/tmp/cosign_${version_for_url}_${architecture}.deb"
@@ -553,7 +553,7 @@ install_cosign() {
         rm "$cosign_filename"
         echo "Installation of cosign succeeded with ${COSIGN_VERSION}."
     else
-        echo "(!) Failed to download cosign package"
+        echo "(!) Failed to install cosign package"
         return 1
     fi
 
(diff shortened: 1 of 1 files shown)


src/python/install.sh
```diff
@@ -507,15 +654,27 @@ install_from_source() {
...
-    # Verify signature
-    if [[ ${VERSION_CODENAME} = "centos7" ]] || [[ ${VERSION_CODENAME} = "rhel7" ]]; then
-        receive_gpg_keys_centos7 PYTHON_SOURCE_GPG_KEYS
+
+    # Discontinuation of PGP signatures for releases of Python 3.14 or future versions
+    # CPython release artifacts are additionally signed with Sigstore starting with the Python 3.11.0
+    local major_version=$(echo "$VERSION" | cut -d. -f1)
+    local minor_version=$(echo "$VERSION" | cut -d. -f2)
+    echo "(*) Detected Python version: ${major_version}.${minor_version}"    
+    if [ "$major_version" -eq 3 ] && [ "$minor_version" -ge 14 ]; then
```
v-Kaniska244: What if python 4 is released in future? Perhaps not imminent but I think we should have a look into this condition.

src/python/install.sh
```diff
@@ -478,6 +499,130 @@ install_cpython() {
...
+        fi
+    fi
+    
+    # Install the package
+    if [ -f "$cosign_filename" ]; then
+        dpkg -i "$cosign_filename"
+        rm "$cosign_filename"
+        echo "Installation of cosign succeeded with ${COSIGN_VERSION}."
+    else
+        echo "(!) Failed to download cosign package"
```
v-Kaniska244: Shouldn't this error message be like "(!) Failed to install cosign package"?

src/python/install.sh
```diff
@@ -478,6 +499,130 @@ install_cpython() {
...
+    
+    echo "Downloading cosign from: ${cosign_url}"
+    
+    if curl -L -f --fail-with-body "${cosign_url}" -o "$cosign_filename" 2>/dev/null; then
+        echo "(*) Successfully downloaded cosign v${COSIGN_VERSION}"
+    else
+        echo -e "\n(!) Failed to fetch cosign v${COSIGN_VERSION}..."
+        # Try previous version
+        find_prev_version_from_git_tags COSIGN_VERSION "https://github.com/sigstore/cosign"
+        echo -e "\nAttempting to install ${COSIGN_VERSION}"
```
v-Kaniska244: Please change this error message to "\nAttempting to install previous cosign ${COSIGN_VERSION} version as fallback mechanism"

Kaniska244 (changes requested):
Hi @Mathiyarasy 

Left some small comments. Kindly let me know in case of any concern.

Issue #275, opened 2022-11-11
Nix feature doesn't work on GitHub Codespaces


NickHu:
Cross-post of existing issue report for visibility: https://github.com/NixOS/nix/issues/6680#issuecomment-1311559039

Chuxel:
There seems to be some sort of interaction with the `universal` image here that needs to be investigated.  The Dev Container Feature does work with other images (as you can see in the test battery).  Will investigate further.

CSchank:
Is there a workaround to this? Until February 14th it was working for us on base:jammy, but now it gives a similar error and falls back on the fallback image.

Chuxel:
@CSchank Hmm. I am currently able to build using `mcr.microsoft.com/devcontainers/base:jammy` and `mcr.microsoft.com/devcontainers/base:bullseye`. Its possible there was an issue with a specific version of nix.

vorburger:
> [@CSchank](https://github.com/CSchank) Hmm. I am currently able to build using `mcr.microsoft.com/devcontainers/base:jammy` and `mcr.microsoft.com/devcontainers/base:bullseye`. Its possible there was an issue with a specific version of nix.

In https://github.com/enola-dev/enola/issues/1921#issuecomment-3567587359, I can confirm this; I've raised #1517 to clarify this on docs, just because this confused the hell out of me at first, until I figured what's what; maybe this helps others in the future (until this is fully resolved).

PR #1517, opened 2025-11-23
[nix] docs: Note it's broken on universal base image (but works on base)

vorburger:
Updated OS support information and clarified requirements.

See https://github.com/devcontainers/features/issues/275.


vorburger:
@microsoft-github-policy-service agree company="Google"

Issue #1272, opened 2025-02-13
Keyserver hkp://keyserver.pgp.com is not reachable


dbeilin:
I'm using the following commands to install Helm:
```shell
curl -s https://raw.githubusercontent.com/devcontainers/features/main/src/kubectl-helm-minikube/install.sh -o /tmp/kubectl-helm-debian.sh
HELM=latest MINIKUBE=none bash /tmp/kubectl-helm-debian.sh
```

However it fails:
```shell
root@cb5da236793d:/go# HELM=latest MINIKUBE=none bash /tmp/kubectl-helm-debian.sh
Downloading kubectl...
/usr/local/bin/kubectl: OK
Downloading Helm...
HELM_VERSION=3.17.1
(*) Keyserver hkp://keyserver.pgp.com is not reachable.
Verification failed!
gpg: assuming signed data in '/tmp/helm/helm-v3.17.1-linux-arm64.tar.gz'
gpg: Signature made Wed Feb 12 20:57:29 2025 UTC
gpg:                using RSA key 748CE9B6B2EDF4D2033318CA07A0310EB5B9076F
gpg: Can't check signature: No public key
```

To test this, you can just run:

```shell
docker run -it --rm golang:latest bash
curl -s https://raw.githubusercontent.com/devcontainers/features/main/src/kubectl-helm-minikube/install.sh -o /tmp/kubectl-helm-debian.sh
HELM=latest MINIKUBE=none bash /tmp/kubectl-helm-debian.sh
```

Any help would be appreciated.

ensonic:
This happened again. The repro from the first comment still applies.

bvarner:
Happening again. @eimhin-rover Any insight?

elefantepequeno:
Happened to me today, retries do not seem to help.
Help version `4.0.0`.

KarstenB:
I have this issue today. The KEYS file in [GitHub](https://github.com/helm/helm/commits/main/KEYS) changed 4 times in the past 6 years. Maybe checking  it in to this repository and validating it offline would be sufficient.

udaya2899:
This happened again today with Helm 4.2.1 release.

PR #1529, opened 2025-12-12
Refactor Helm installation script to use local GPG keys file

KarstenB:
- Removed the download of GPG keys from GitHub.
- Eliminated the function for checking reachable GPG key servers.
- Updated the script to copy a local KEYS file instead.

This approach eliminates pgp.com server down issue like #1272 and can sometimes improve build performance. The only downside is that the KEYS file needs to be updated when upstream updates it. But a quick check of https://github.com/helm/helm/commits/main/KEYS shows that those are very infrequent.


abdurriq:
@KarstenB Thanks for raising this. I agree that there should be a solution for slowdowns caused by pgp.com issues, but I'm not sure this is the correct approach. Instead, I would rather we address the root cause (fetching being slow). If that is not possible, could we consider a hybrid (opt-in) approach to using a local key?

We also need to consider the precedent of storing all keys used in the repo locally; I'm not sure this is ideal.

abdurriq:
cc @chrmarti who may have a better idea regarding this.

PR #1530, opened 2025-12-16
feat(github-cli): add support for extensions

neilime:
If you set the `extensions` option, the feature will run `gh extension install` for each entry (comma-separated). Extensions are installed for the most appropriate non-root user (based on `USERNAME` / `_REMOTE_USER`), with a fallback to `root`.

### DIFF
--- a/src/github-cli/install.sh
+++ b/src/github-cli/install.sh
@@ -9,6 +9,7 @@
 
 CLI_VERSION=${VERSION:-"latest"}
 INSTALL_DIRECTLY_FROM_GITHUB_RELEASE=${INSTALLDIRECTLYFROMGITHUBRELEASE:-"true"}
+EXTENSIONS=${EXTENSIONS:-""}
 
 GITHUB_CLI_ARCHIVE_GPG_KEY=23F3D4EA75716059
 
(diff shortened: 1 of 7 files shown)


neilime:

test/github-cli/scenarios.json
```diff
@@ -1,11 +1,20 @@
...
     }
-}
\ No newline at end of file
+  },
+  "install_extensions": {
+    "image": "ubuntu:noble",
+    "features": {
+      "github-cli": {
+        "version": "latest",
+        "extensions": "dlvhdr/gh-dash"
```
abdurriq: Could we please test multiple extensions (at least two), just to verify that the parsing works correctly.

src/github-cli/install.sh
```diff
@@ -242,5 +243,91 @@ else
...
+                USERNAME="${CURRENT_USER}"
+                break
+            fi
+        done
+        if [ -z "${USERNAME}" ]; then
+            USERNAME=root
+        fi
+    elif [ "${USERNAME}" = "none" ] || ! id -u "${USERNAME}" > /dev/null 2>&1; then
+        USERNAME=root
+    fi
```
abdurriq, line 267: (Note to self: we need to provide this as a helper rather than duplicating across scripts).

abdurriq: devcontainers/features/issues/1551

src/github-cli/install.sh
```diff
@@ -242,5 +243,91 @@ else
...
+            fi
+        else
+            su - "${USERNAME}" -c "set -e; extensions_root=\"\${XDG_DATA_HOME:-\"\$HOME/.local/share\"}/gh/extensions\"; repo_name=\"${extension##*/}\"; mkdir -p \"\$extensions_root\"; if [ ! -d \"\$extensions_root/\$repo_name\" ]; then git clone --depth 1 \"https://github.com/${extension}.git\" \"\$extensions_root/\$repo_name\"; fi"
+        fi
+    done
+
+        # Some gh builds require auth even for `gh extension list`. If so, provide a tiny wrapper
+        # that lists locally installed extensions without network/auth.
+        if ! gh extension list >/dev/null 2>&1; then
+                cat > /usr/local/bin/gh <<'EOF'
```
abdurriq: Nit: Would be more readable if this was moved into a separate file.

src/github-cli/install.sh
```diff
@@ -242,5 +243,91 @@ else
...
+        if [ "${USERNAME}" = "root" ]; then
+            extension_home="${HOME}"
+            extensions_root="${XDG_DATA_HOME:-"${extension_home}/.local/share"}/gh/extensions"
+            repo_name="${extension##*/}"
+            mkdir -p "${extensions_root}"
+            if [ ! -d "${extensions_root}/${repo_name}" ]; then
+                git clone --depth 1 "https://github.com/${extension}.git" "${extensions_root}/${repo_name}"
+            fi
+        else
+            su - "${USERNAME}" -c "set -e; extensions_root=\"\${XDG_DATA_HOME:-\"\$HOME/.local/share\"}/gh/extensions\"; repo_name=\"${extension##*/}\"; mkdir -p \"\$extensions_root\"; if [ ! -d \"\$extensions_root/\$repo_name\" ]; then git clone --depth 1 \"https://github.com/${extension}.git\" \"\$extensions_root/\$repo_name\"; fi"
```
abdurriq: Nit: Would be nice to have this logic moved into a separate script and then called both above and here with su for readability.

abdurriq:
@neilime Thank you for this feature; it should be very useful. I have left some comments which I would appreciate if you could address, particularly the one related to testing.

neilime:
@abdurriq it's done, let me know if it's ok now.

Thanks

Kaniska244 (approved):

abdurriq:
Approved and merged, thank you for your contribution @neilime.

Issue #1536, opened 2026-01-09
Dev Containers fails with rootless Docker socket; need user-configurable host docker socket path / priority


cubicYYY:
## Description

When using Dev Containers with `docker-outside-of-docker` feature, container startup can fail if the **host Docker socket is not at a predefined path**.

For example, in rootless Docker setups, the daemon socket is typically located at:

```
/run/user/$UID/docker.sock
```

instead of the conventional:

```
/var/run/docker.sock
```

The feature/init script assumes a fixed socket position `/var/run/docker.sock`.

If the actual host socket is elsewhere, the default mount fails.
Attempting to fix this by manually mounting the rootless socket failed to override this path, causing conflicts.

## Expected behavior

* Dev Containers should work with both rootful and rootless Docker without fragile manual mounts.
* Users should be able to **explicitly specify or override the preferred host Docker socket path**.

## Actual behavior

* Startup fails when the host socket is not at an assumed location.
* Manual socket mounts can conflict with feature-managed sockets (`docker-host.sock` vs `docker.sock`).

## Environment

* Host: Linux, rootless Docker (`/run/user/$UID/docker.sock`)
* Dev Containers / docker-outside-of-docker feature
* (versions can be added)

## Suggested improvement

Add a **supported configuration option** to override host Docker socket selection, for example:

* `hostDockerSocket`
* `sourceSocket`
* or an ordered preference list (e.g. `${XDG_RUNTIME_DIR}/docker.sock` → `/var/run/docker.sock`)

This would:

* Avoid hardcoded assumptions about socket paths
* Eliminate conflicting manual mounts
* Make Dev Containers robust in rootless and nested Docker environments

## Additional note

Rootless Docker is common on shared machines and restricted environments. First-class support for non-standard socket paths would significantly improve reliability.

Clockwork-Muse:
This also affects things podman, which:
 1) Doesn't even have a socket by default
 2) Runs rootless by default
 3) Uses a different socket name, obviously

In my case, I'm running inside a toolbox on Fedora 43, using podman.  I've managed to get around things by using `podman-remote` and symlinking to that, but need this feature to work as well.

That aside, last I was aware it's not possible to expose mount configurations from extension entries in the `.devcontainer.json` - we need new spec/tooling support for that.

sireeshajonnalagadda:
Hi @cubicYYY,
[...]

PR #1549, opened 2026-01-29
Add rootless Docker support and update documentation

sireeshajonnalagadda:
This PR fixes the [issue](https://github.com/devcontainers/features/issues/1536) for rootless docker 
- Introduced socketPath option for custom Docker socket configuration in README and devcontainer-feature.json.
- Updated install script to accommodate rootless Docker socket paths.
- Added tests for rootless Docker configurations and custom socket paths.
- Created new documentation for rootless Docker setups and scenarios.
- Older Docker CE versions that bundled buildx with CLI will still respect the installDockerBuildx: false setting
Modern Docker CE versions with separate packages work as expected
- Cleanup runs after installation so it can properly detect and remove bundled buildx files
- Added explanatory comments for future maintainers

## Key Points
1. **socketPath option*: Configures where the feature expects the socket inside the container
2. **Mount source**: Must match your host's actual Docker socket location
3. **Mount target**: Must match the `socketPath` option value
4. **XDG_RUNTIME_DIR**: Usually `/run/user/{uid}` for rootless Docker
## Test Scenarios
The test scenarios demonstrate:
- `rootless_docker_socket`: Standard rootless configuration
- `root_docker_socket`: Standard root Docker (for comparison)

- [x] Checklist 
Changes works as expected for both root and rootless docker environments.

### DIFF
--- a/src/docker-outside-of-docker/install.sh
+++ b/src/docker-outside-of-docker/install.sh
@@ -329,10 +329,20 @@ else
         fi
         #install cli + buildx first
         apt-get -y install --no-install-recommends ${cli_package_name}${cli_version_suffix} "${buildx[@]}"
+        
+        # Backward compatibility: Older Docker CE versions bundled buildx with CLI
+        # Modern versions have separate packages, but this ensures consistent behavior
+        buildx_path="/usr/libexec/docker/cli-plugins/docker-buildx"
+        if [ "${INSTALL_DOCKER_BUILDX}" = "false" ] && [ -f "${buildx_path}" ]; then
+            echo "(*) Removing docker-buildx (bundled in older Docker CE) since installDockerBuildx is disabled..."
+            rm -f "${buildx_path}"
+        fi
+        
         if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "v1" ]; then
             apt-get -y install --no-install-recommends docker-compose-plugin
         fi
     fi
+    unset buildx buildx_path
 fi
 # If 'docker-compose' command is to be included
 if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "none" ]; then
--- a/test/docker-outside-of-docker/root_docker_socket.sh
+++ b/test/docker-outside-of-docker/root_docker_socket.sh
@@ -1,16 +1,11 @@
 #!/bin/bash
-# Test script to detect Docker type
+# Test script to assert root Docker socket usage
 
-if [ -S "/var/run/docker.sock" ]; then
-    echo "Root Docker detected"
-    export DOCKER_HOST="unix:///var/run/docker-host.sock"
-elif [ -S "/var/run/docker-rootless.sock" ]; then
-    echo "Rootless Docker detected"  
-    export DOCKER_HOST="unix:///var/run/docker-rootless.sock"
-else
-    echo "No Docker socket found"
+if [ ! -S "/var/run/docker-host.sock" ]; then
+    echo "ERROR: Root Docker socket not found"
     exit 1
 fi
 
+echo "Root Docker detected"
+export DOCKER_HOST="unix:///var/run/docker-host.sock"
 docker --version
-docker info --format '{{.SecurityOptions}}'


src/docker-outside-of-docker/install.sh
```diff
@@ -318,13 +319,19 @@ else
...
+        if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "v1" ]; then
+            apt-get -y install --no-install-recommends moby-compose || echo "(*) Package moby-compose (Docker Compose v2) not available for OS ${ID} ${VERSION_CODENAME} (${architecture}). Skipping."
+        fi
     else
         buildx=()
         if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
             buildx=(docker-buildx-plugin)
         fi
-        apt-get -y install --no-install-recommends ${cli_package_name}${cli_version_suffix} "${buildx[@]}" docker-compose-plugin
+        if [ "${DOCKER_DASH_COMPOSE_VERSION}" != "v1" ]; then
```
v-Kaniska244, line 341: What is the relevance of this check for rootless Docker support?

test/docker-outside-of-docker/ROOTLESS_DOCKER.md
v-Kaniska244, line 1: Why do we have this documentation in the test directory? If any change required, couldn't this be done in the readme file or even as PR comment if applicable?

test/docker-outside-of-docker/scenarios.json
```diff
@@ -180,5 +180,72 @@
...
+        "mounts": [
+            {
+                "source": "/var/run/docker.sock",
+                "target": "/custom/docker/rootless.sock",
+                "type": "bind"
+            }
+        ],
+        "containerUser": "vscode"
+    },
+    "xdg_runtime_dir_socket": {
```
v-Kaniska244: Why do we really need this test when it appears to be quite similar to `rootless_docker_socket` scenario?

test/docker-outside-of-docker/scenarios.json
```diff
@@ -180,5 +180,72 @@
...
+        "containerUser": "vscode"
+    },
+    "root_docker_socket": {
+        "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+        "features": {
+            "docker-outside-of-docker": {
+                "moby": false
+            }
+        },
+        "mounts": [
```
v-Kaniska244: Do we really need the explicit mounts for this case as I guess this should be the default behavior. isn't it?

src/docker-outside-of-docker/devcontainer-feature.json
```diff
@@ -1,51 +1,56 @@
...
+            "description": "Compose version to use for docker-compose (v1 or v2 or none)"
+        },
+        "installDockerBuildx": {
+            "type": "boolean",
+            "default": true,
+            "description": "Install Docker Buildx"
+        },
+        "installDockerComposeSwitch": {
+            "type": "boolean",
+            "default": true,
```
v-Kaniska244: Although not quite related to this PR, can we make this flag as false by default to align the behavior with the d-in-d feature? This is because the compose-switch component which is installed using this flag is obsolete and not really required by default as the docker compose v2 is the default configuration for the feature.

test/docker-outside-of-docker/ROOTLESS_DOCKER.md
sireeshajonnalagadda: Sorry, the file was misplaced. The rootless socket information has already been added to README.md. ROOTLESS_DOCKER.md provides clear examples on how to configure d-o-d feature with rootless docker setup. I will remove it from here and add a summary in the PR comment. 
Thanks!

test/docker-outside-of-docker/scenarios.json
sireeshajonnalagadda: This just another variant of using custom socket path. If it appears redundant I will remove this.

src/docker-outside-of-docker/devcontainer-feature.json
```diff
@@ -1,51 +1,56 @@
...
-      ],
-      "default": "v2",
-      "description": "Compose version to use for docker-compose (v1 or v2 or none)"
-    },
-    "installDockerBuildx": {
-      "type": "boolean",
-      "default": true,
-      "description": "Install Docker Buildx"
+    "id": "docker-outside-of-docker",
+    "version": "1.9.1",
```
v-Kaniska244, line 4: Shouldn’t this be `1.9.0` instead of `1.9.1`. The current version is `1.8.0`.

sireeshajonnalagadda: Downgraded the version to 1.9.0

src/docker-outside-of-docker/install.sh
abdurriq: +1 @sireeshajonnalagadda Did we not need this before; if so, how? Also, since they are both very similiar, can we do:
```
install cli package + buildx
if not Docker v1
   install docker-compose-plugin
```

test/docker-outside-of-docker/scenarios.json
```diff
@@ -180,5 +180,65 @@
...
+    "root_docker_socket": {
+        "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+        "features": {
+            "docker-outside-of-docker": {
+                "moby": false
+            }
+        },
+        "containerUser": "vscode"
+    },
+    "custom_rootless_socket_path": {
```
abdurriq: Isn't this identical to `rootless_docker_socket` (just with different socketPaths)? Both are custom.

test/docker-outside-of-docker/scenarios.json
abdurriq: Agreed, could you please remove this too?

test/docker-outside-of-docker/scenarios.json
```diff
@@ -202,3 +228,5 @@
         "containerUser": "vscode"
     }
 }
+
```
abdurriq, line 231: Nit: Too many newlines

test/docker-outside-of-docker/root_docker_socket.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Test script to detect Docker type
+
+if [ -S "/var/run/docker.sock" ]; then
+    echo "Root Docker detected"
+    export DOCKER_HOST="unix:///var/run/docker-host.sock"
+elif [ -S "/var/run/docker-rootless.sock" ]; then
+    echo "Rootless Docker detected"
```
abdurriq: Shouldn't this assert that we are using the root socket, not a rootless one?

src/docker-outside-of-docker/install.sh
```diff
@@ -318,23 +319,21 @@ else
...
+        fi
     else
         buildx=()
         if [ "${INSTALL_DOCKER_BUILDX}" = "true" ]; then
             buildx=(docker-buildx-plugin)
         fi
-        apt-get -y install --no-install-recommends ${cli_package_name}${cli_version_suffix} "${buildx[@]}" docker-compose-plugin
-        buildx_path="/usr/libexec/docker/cli-plugins/docker-buildx"
-        # Older versions of Docker CE installs buildx as part of the CLI package
-        if [ "${INSTALL_DOCKER_BUILDX}" = "false" ] && [ -f "${buildx_path}" ]; then
```
abdurriq, line 330: Do we not want to keep this for backwark compatibility?

PR #1550, opened 2026-01-29
[node] - Removal of default installation of yarn v1(classic)

Kaniska244:
**Ref:** [issue](https://github.com/devcontainers/images/issues/1752) 

**Description:** This yarn v1(classic) is out of active development cycle for quite sometime. And yarn v2(berry) is more flexible and available for installation via `corepack` when `installYarnUsingApt` flag is set to false.

**Changelog:** 

- Update install.sh and devcontainer-feature.json to make sure that `installYarnUsingApt` flag is set to true by default. Also ensure that yarn installation function doesn't proceed when node version is `none`
- Version bump.
- Added separate tests yarn installation via corepack.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.6.5",
+    "version": "1.7.0",
     "name": "Node.js (via nvm), yarn and pnpm.",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",


src/node/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "1.6.4",
+    "version": "1.6.5",
```
abdurriq: Shall we increment minor version since this may change assumptions being made by downstream users (Yarn may not be installed anymore)?

abdurriq (dismissed):

src/node/devcontainer-feature.json
v-Kaniska244: Hi @abdurriq 

Thank you for pointing this out. I have done the same now.

Issue #1560, opened 2026-02-04
PowerShell Feature


SQLCanuck:
It would be great if the "version" parameter supported "lts" as an option, which would use the latest available LTS version of PowerShell.
The following url points to the Latest LTS version: https://aka.ms/powershell-release?tag=lts

On another note, some of the older version numbers should be dropped as they are out of the support lifecycle.
https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.5#powershell-end-of-support-dates

v-Kaniska244:
Hi @SQLCanuck 

Thank you for reporting this issue. The `lts` version should already be supported as part of feature and if we specify `"version": "lts"` in the feature configuration part of the dev container configuration, it is expected to work.
However as you have rightly pointed out, we will add it as proposal and remove the obsolete one's.

SQLCanuck:
@Kaniska244 when I tested using "version": "lts" I ended up with the latest v7.5.4 version rather than the expected LTS v7.4.13.  Note that for the latest LTS, there is no specific tag for lts in the PowerShell GitHub repository.  It instead relies on a redirect link provided by Microsoft https://aka.ms/powershell-release?tag=lts .  Microsoft provides redirect links for the latest LTS, Stable, and Preview as can be seen in the Note here https://learn.microsoft.com/en-us/powershell/scripting/install/install-debian

If possible, I would suggest following a similar pattern for the version value, by using "lts", "stable", "preview", "latest", "exact version" as possible values.

SQLCanuck:
Screenshot of the issue.

<img width="552" height="242" alt="Image" src="https://github.com/user-attachments/assets/a15f30cd-f351-4274-b71d-767f2c04c283" />

v-Kaniska244:
PR merged. Kindly use `ghcr.io/devcontainers/features/powershell:2` get back if issue persists.

SQLCanuck:
@Kaniska244 it is working great.  Thank you for your time resolving the case, it is greatly appreciated.

PR #1562, opened 2026-02-06
[powershell] - Add `lts`, `stable` and `preview`, also remove deprecated versions.

Kaniska244:
**Ref:** #1560 

**Description:** 

- Added `lts`, `stable` and `preview` version in proposal and removed deprecated versions. Also added logic to redirect from link provided by Microsoft(such as https://aka.ms/powershell-release?tag=lts) . Microsoft provides redirect links for the LTS, Stable, and Preview versions.

- Additionally the [Microsoft packages repo for debian trixie(13)](https://packages.microsoft.com/repos/microsoft-debian-trixie-prod/dists/trixie/main/binary-amd64/Packages) doesn't have the `powershell` package available unlike other editions of debian/ ubuntu distributions. So, the package has to be downloaded from GitHub repo and installed for debian trixie(13) and the download logic was broken. Fixed that as well as part of this PR. 

**Changelog:** 

- Update devcontainer-feature.json to add `lts`, `stable` and `preview`, also remove deprecated versions
- Fixing package download logic from GitHub in the install.sh. Added the redirection logic for `lts`, `stable` and `preview` versions.
- Major version bump.
- Added separate tests.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/src/powershell/install.sh
+++ b/src/powershell/install.sh
@@ -365,7 +365,6 @@ install_pwsh() {
 
 install_using_github() {
     # Fall back on direct download if no apt package exists in microsoft pool
-    # Fall back on direct download if no apt package exists in microsoft pool
     if command -v apt-get > /dev/null 2>&1; then
         # Debian/Ubuntu dependencies
         check_packages curl ca-certificates gnupg2 dirmngr libc6 libgcc1 libgssapi-krb5-2 libstdc++6 libunwind8 libuuid1 zlib1g libicu[0-9][0-9]


src/powershell/devcontainer-feature.json
```diff
@@ -37,7 +37,7 @@
         "settings": {
                 "github.copilot.chat.codeGeneration.instructions": [
                     {
-                        "text": "This dev container includes PowerShell along with needed dependencies pre-installed and available on the `PATH`, along with the PowerShell extension."
```
abdurriq, line 40: Nit: Shouldn't have two full-stops.

test/powershell/scenarios.json
```diff
@@ -51,5 +51,37 @@
...
+      }
+    }
+  },
+  "powershell_preview_version_debian": {
+    "image": "mcr.microsoft.com/devcontainers/base:debian",
+    "features": {
+      "powershell": {
+        "version": "preview"
+      }
+    }
```
abdurriq, line 93: Do we only support Debian-based distros (Debian, Ubuntu), or others? If so, can we also run these tests on one of those (e.g. Alpine)?

v-Kaniska244: Hi @abdurriq 

Thank you for pointing this out. Although, the [os-support](https://github.com/devcontainers/features/tree/main/src/powershell#os-support) says that the feature supports only debian / ubuntu, I could see existing test case of RHEL based `almalinux:9` and as per the installation script `dnf` installer is supported. 
So I have added test cases on `almalinux:9` for `lts`, `stable` and `preview`. Kindly let me know in case of further concern.

abdurriq (approved):

PR #1576, opened 2026-02-17
Fix conda installation failure due to SHA1 signature rejection

abdurriq:
Conda installation was failing due to SHA1 signature no longer being accepted. This change switches to downloading the deb directly.

Original PR (merged into another feature branch): https://github.com/devcontainers/features/pull/1565


Kaniska244 (approved):

src/conda/install.sh
```diff
@@ -83,20 +83,78 @@ if ! conda --version &> /dev/null ; then
     usermod -a -G conda "${USERNAME}"
 
     # Install dependencies
-    check_packages curl ca-certificates gnupg2
+    check_packages curl ca-certificates
 
     echo "Installing Conda..."
 
-    curl -sS https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc | gpg --dearmor > /usr/share/keyrings/conda-archive-keyring.gpg
-    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/conda-archive-keyring.gpg] https://repo.anaconda.com/pkgs/misc/debrepo/conda stable main" > /etc/apt/sources.list.d/conda.list
```
sam-byng, line 91: question: do we no longer need this gpgkeys for signing the deb download or does the new `apt-get install` handle this safely under the hood?

i.e. is this 'gpg' logic the piece that is breaking due to sha1?

sam-byng:
Overall LGTM just a question

src/conda/install.sh
abdurriq: @sam-byng Yeah, we're having to give up GPG validation for this, which isn't ideal, but there aren't any other workarounds for now, as they've deprecated the DEB upstream:
https://github.com/conda/conda/issues/14826

We should open an issue to move away from the deb to the other supported methods, but that will take additional effort.

abdurriq: https://github.com/devcontainers/features/issues/1583 created for this.

PR #1591, opened 2026-02-24
[nvidia-cuda] - Fix installation issue on debian trixie(13)

Kaniska244:
**Description:** Fixing an [installation issue](https://github.com/devcontainers/features/actions/runs/21873828799/job/63136744052) of nvidia-cuda feature on debian trixie(13)

**Changelog:** 

- Bypassed the GPG validation logic for cuda installation from nvidia repository when platform is debian trixie(13) by using `AllowInsecureRepositories` flag. 
- Updated the default cuda version to a later, more widely available version `12.5` 
- Added tests.
- Version bump.

**Checklist:**
- [x] All checks are passed.


test/nvidia-cuda/install_all_options_debian_12.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
abdurriq: Can we make these files symlinks since the tests are identical?

v-Kaniska244: Hi @abdurriq 

Thank you for pointing this out. I have corrected them now. Please let me know in case of any further concern.

v-Kaniska244:
> LGTM, can we remove the new `features` file that was created though.

Hi @abdurriq 

Thank you for pointing this out. A submodule for features repo was accidentally created due to my local VSCode setup. I have corrected it now.

faximan:
Is there an upstream issue to nvidia for removing this "temporary" workaround requiring `AllowInsecureRepositories`? If there is no follow up these things tend to stay not so temporary.

v-Kaniska244:
> Is there an upstream issue to nvidia for removing this "temporary" workaround requiring `AllowInsecureRepositories`? If there is no follow up these things tend to stay not so temporary.

Hello @faximan ,

Please find the [nvidia issue](https://forums.developer.nvidia.com/t/apt-update-fails-due-to-obsolete-key/359462/16) for this problem. But so far, no fix received upstream and the advent of one seems unlikely.

PR #1603, opened 2026-03-12
[oryx] - Upgrade DEBIAN_FLAVOR to `bookworm`

Kaniska244:
**Description:** Upgrading Microsoft Oryx DEBIAN_FLAVOR to `bookworm` and Java will not be supported in the new version of the feature as it won't be supported [upstream](https://github.com/microsoft/Oryx/issues/2592#issuecomment-3821949686).

**Changelog:** 

- Update the DEBIAN_FLAVOR to `bookworm`
- Modified tests.
- Major Version bump as it's a breaking change.

**Checklist:**
- [x] All checks are passed.

### DIFF
--- a/test/oryx/install_older_dotnet_and_oryx.sh
+++ b/test/oryx/install_older_dotnet_and_oryx.sh
@@ -9,18 +9,18 @@ check "Oryx version" oryx --version
 check "Dotnet is not removed if it is not installed by the Oryx Feature" dotnet --version
 
 # Install platforms with oryx build tool
-check "oryx-install-dotnet-8.0" oryx prep --skip-detection --platforms-and-versions dotnet=8.0.23
-check "dotnet-2-installed-by-oryx" ls /opt/dotnet/ | grep 8.0
+check "oryx-install-dotnet-6.0" oryx prep --skip-detection --platforms-and-versions dotnet=6.0.23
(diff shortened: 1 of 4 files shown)


test/oryx/install_older_dotnet_and_oryx.sh
```diff
@@ -9,17 +9,14 @@ check "Oryx version" oryx --version
 check "Dotnet is not removed if it is not installed by the Oryx Feature" dotnet --version
 
 # Install platforms with oryx build tool
-check "oryx-install-dotnet-2.1" oryx prep --skip-detection --platforms-and-versions dotnet=2.1.30
-check "dotnet-2-installed-by-oryx" ls /opt/dotnet/ | grep 2.1
+check "oryx-install-dotnet-8.0" oryx prep --skip-detection --platforms-and-versions dotnet=8.0.23
```
abdurriq: In the scenario we state that we should have Dotnet 7.0 but here we have 8.0; is that param still respected? If not, do we need these tests (both this and the "prev" one)?

v-Kaniska244: Hi @abdurriq 

I have modified the tests as per the comment. Would you kindly have a look again.

PR #1606, opened 2026-03-19
feat(github-cli): add github-cli auth-on-setup option

neilime:
## Summary

Add an optional `authOnSetup` flow to authenticate `gh` after container startup, and make GitHub CLI extension installation work correctly with runtime auth or a git-based fallback.

**no breaking changes:** the original behavior is the default one.

## Changes

- add `authOnSetup` and `installExtensionsFromGit` feature options
- defer extension installation until post-start auth when needed
- support token-based login via `GH_TOKEN` or `GITHUB_TOKEN`
- fail fast for unsupported extension configurations
- refactor the feature install flow into smaller auth and extension scripts

**Battle tested with automated tests and in real-world devcontainer**


Kaniska244 (approved):

neilime:
Hi @Kaniska244 any chance to have this PR merged? Thank you :)

neilime:
Hi @Kaniska244, here is a kindly reminder, hoping having this PR merged one day :) 

Thanks

abdurriq:
Hello @neilime, apologies this is my fault as I've been meaning to look at this but haven't had a chance. I'll try to review this within the week.

abdurriq:
@neilime Would you be able to break this PR up, as I'm struggling to review it given the size and number of files touched?

neilime:
@abdurriq done: 6 atomic commits (atomic as much as I can) with their own purpose

src/github-cli/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "github-cli",
-    "version": "1.1.0",
+    "version": "1.3.0",
```
abdurriq: Q: Why not 1.2.0?

abdurriq (approved):

abdurriq:
I noticed some inconsistencies in double-quotes around params and the logic between `src/github-cli/scripts/install-github-cli.sh` and `src/github-cli/install.sh` in (https://github.com/devcontainers/features/pull/1606/changes/98182e13036f8d4ef0bf3046cfab8a9f6caf3a0c). I wonder why it wasn't just a 1:1 copy, since we know the old version already works. Either way, the new one seems OK, so it shouldn't be too much of an issue.

abdurriq (dismissed):

neilime:
> I noticed some inconsistencies in double-quotes around params and the logic between `src/github-cli/scripts/install-github-cli.sh` and `src/github-cli/install.sh` in ([98182e1](https://github.com/devcontainers/features/commit/98182e13036f8d4ef0bf3046cfab8a9f6caf3a0c)). I wonder why it wasn't just a 1:1 copy, since we know the old version already works. Either way, the new one seems OK, so it shouldn't be too much of an issue.

Not a strict 1:1 copy on purpose: the top-level script now only orchestrates the install/auth/extensions scopes, so the previous logic was split into dedicated scripts. Your concern about drift was valid though, and I found one real inconsistency in the installExtensionsFromGit path and fixed it. Aside from that, the remaining differences are mostly cleanup, plus one small fallback fix in the GitHub CLI installer.

src/github-cli/devcontainer-feature.json
neilime: Solved

neilime:
@abdurriq Hi, any news for merging this PR? Thanks

PR #1616, opened 2026-04-07
Add support for configuring npm version in Node devcontainers

sireeshajonnalagadda:
Context:  
When using the Node feature in devcontainers, npm is installed globally but always defaults to the bundled version. This often leads to noisy update prompts when running npm commands, and any manual updates are lost after rebuilding the container.

Problem:

Developers frequently need to pin a specific npm version globally.

Current workaround requires adding an onCreateCommand in devcontainer.json to run npm install -g npm@<version>.

This is repetitive and not baked into the image.

Solution (Implemented in this PR):

Added support for specifying npmVersion alongside version in the Node feature configuration.

Example usage:

json
"ghcr.io/devcontainers/features/node:1": {
  "version": "24",
  "npmVersion": "11.11.1"
}
Ensures npm version is installed globally during container build, avoiding update prompts and rebuild inconsistencies.

Changes made:

- Adds a new npmVersion option to the Node devcontainer feature so users can pin npm (or set it to latest / none) directly in devcontainer.json.

- Updates the Node feature install script to install/upgrade npm globally during build, including skip logic (none), retries, proxy handling, and a compatibility fallback when the requested npm version doesn’t match the Node version.

- Adds/updates tests and scenarios to cover installing a specific npm version, skipping npm updates, upgrading to latest, and an incompatible “latest npm + Node 16” case.

Outcome:  
Developers can now configure both Node and npm versions directly in their devcontainer setup, improving reproducibility and reducing noise from npm update prompts.


src/node/devcontainer-feature.json
```diff
@@ -27,6 +27,22 @@
...
+                "latest",
+                "10.9.0",
+                "10.8.0",
+                "10.7.0",
+                "9.9.3",
+                "8.19.4",
+                "latest",
+                "none"
+            ],
+            "default": "10.9.0",
```
Kaniska244: This default npm version should be `lts` or `latest`. Better change it to `lts`

src/node/install.sh
```diff
@@ -381,6 +382,102 @@ if [ ! -z "${ADDITIONAL_VERSIONS}" ]; then
...
+                        fi
+                    elif [ "$MIN_NODE" -gt "0" ]; then
+                        echo "✓ Node.js $NODE_MAJOR.x meets npm $NPM_VERSION requirement"
+                    fi
+                else
+                    echo "Could not determine Node.js requirements for npm $NPM_VERSION, proceeding anyway..."
+                fi
+            fi
+            
+            if [ -z "$NPM_VERSION" ] || [ "$NPM_VERSION" = "none" ]; then
```
abdurriq: Could you please move these to the top condition? That would simplify the branching here a lot.

test/node/install_npm_none.sh
```diff
@@ -0,0 +1,29 @@
...
+source dev-container-features-test-lib
+
+# When npmVersion is "none", npm should not be updated from node's bundled version
+check "npm_not_updated" bash -c '
+    npm --version >/dev/null
+
+    NODE_MAJOR=$(node -p "process.versions.node.split(\".\")[0]")
+    NPM_MAJOR=$(npm --version | cut -d. -f1)
+
+    case "$NODE_MAJOR" in
```
abdurriq, line 15: Since we can control which Node version is installed, can we pin to the latest (24) -- so that upgrades don't break this test, and then verify NPM 11 is installed?

abdurriq (approved):

Kaniska244 (approved):

PR #1622, opened 2026-04-16
[conda] - Accept the terms of service for root and target user and conda cleanup

Kaniska244:
**Description:** With the new conda upgrade, it's mandatory to accept conda terms of service and previously it was only done for the root user of the container which isn’t adequate. As part of this PR pre-accepting conda terms of service for the configured target user in the container. Also clean up conda cache to free up some disk space.

**Changelog:** 

- Pre-accept conda terms of reference for the target user in the container. Clean up conda cache and refrain from upgrading old vulnerable python libraries as with the change of source of installation, the libraries are already up-to-date.
- Added tests for both root and target users
- Minor version bump.

**Checklist:**
- [x] All checks are passed.


src/conda/devcontainer-feature.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "conda",
-  "version": "2.0.0",
```
abdurriq, line 3: Shouldn't we override 2.0.0 since I assume it currently won't work?

src/conda/install.sh
```diff
@@ -147,14 +160,14 @@ if ! conda --version &> /dev/null ; then
 
     # Temporary fixes
     # Due to https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23491
-    install_user_package certifi
+    #install_user_package certifi
```
abdurriq: Why were these commented out?

Kaniska244: Hi @abdurriq 

These are not required to be manually updated for now. With the installation source change, these packages are installed having latest versions only, so I commented out these steps.
Screenshot 2026-04-16 at 3 34 15 PM

src/conda/devcontainer-feature.json
Kaniska244: Hi @abdurriq 

I will change it.

src/conda/install.sh
abdurriq: @Kaniska244 Could we please remove them then?

abdurriq (dismissed):

abdurriq (approved):

PR #1624, opened 2026-04-16
Add default `copilot update` to `postStartCommand`

cmbrose:
Now every time the container restarts copilot will upgrade itself so that you're on the latest version.

Open to feedback on if this should be in `postCreateCommand` so that it's only done once when the container is created.

### DIFF
--- a/src/copilot-cli/devcontainer-feature.json
+++ b/src/copilot-cli/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "copilot-cli",
-    "version": "1.1.1",
+    "version": "1.1.0",
     "name": "GitHub Copilot CLI",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/copilot-cli",
     "description": "Installs the GitHub Copilot CLI.",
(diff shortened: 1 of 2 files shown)


src/copilot-cli/devcontainer-feature.json
```diff
@@ -15,6 +15,7 @@
             "description": "Select version of the GitHub Copilot CLI, if not latest."
         }
     },
+    "postStartCommand": "copilot update",
```
eljog: will this discard the feature option that allows the user to select a specific version?

abdurriq: @eljog Good point, it will force an update. Perhaps we can make it conditional on if the version is latest or not; i.e. by persisting the version that was selected at install time.

cmbrose: Fair enough, I updated the logic so that `install.sh` creates a marker file if `latest` is used and the update only triggers in that case

eljog (dismissed):

abdurriq (dismissed):

src/copilot-cli/devcontainer-feature.json
```diff
@@ -15,6 +15,7 @@
             "description": "Select version of the GitHub Copilot CLI, if not latest."
         }
     },
+    "postStartCommand": "[ -f /etc/devcontainer-copilot-cli/use-latest ] && copilot update || true",
```
abdurriq: As suggested by Devraj, we may want to do this for prerelease too.

abdurriq (approved):

PR #1625, opened 2026-04-16
[node] - Install pnpm as non-root user to prevent root-owned npm cache

brianhelba:
Currently, the pnpm installation block runs `npm install -g pnpm` in a bare subshell as root, unlike every other npm/nvm operation in the script which uses `su ${USERNAME}`. This causes the npm cache directory to be created owned by `root:root`, leading to `EACCES` errors for the non-root user on subsequent npm operations.

This is particularly reproducible on macOS with Rosetta 2 emulation, where the cache directory may not already exist from prior steps.

Note, the explicit setting of proxy env vars (`http_proxy`, `https_proxy`, `no_proxy`) is likely a workaround for https://github.com/npm/cli/issues/6835. No other commands in this script use that workaround anymore, so this change uses the same `su` syntax as all other commands.

### DIFF
--- a/src/node/install.sh
+++ b/src/node/install.sh
@@ -482,13 +482,7 @@ if [ ! -z "${PNPM_VERSION}" ] && [ "${PNPM_VERSION}" = "none" ]; then
     echo "Ignoring installation of PNPM"
 else
     if bash -c ". '${NVM_DIR}/nvm.sh' && type npm >/dev/null 2>&1"; then
-        (
-            . "${NVM_DIR}/nvm.sh"
-            [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
-            [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
-            [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
-            npm install -g pnpm@$PNPM_VERSION --force
-        )
+        su ${USERNAME} -c "umask 0002 && . '${NVM_DIR}/nvm.sh' && npm install -g pnpm@${PNPM_VERSION} --force"
     else
         echo "Skip installing pnpm because npm is missing"
     fi
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "2.0.1",
+    "version": "2.1.0",
     "name": "Node.js (via nvm), yarn and pnpm.",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",
--- a/src/node/devcontainer-feature.json
+++ b/src/node/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "node",
-    "version": "2.0.0",
+    "version": "2.0.1",
     "name": "Node.js (via nvm), yarn and pnpm.",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/node",
     "description": "Installs Node.js, nvm, yarn, pnpm, and needed dependencies.",


Kaniska244 (changes requested):
Hi @brianhelba 

Thank you for the contribution. Would you kindly bump the node feature version.

brianhelba:
@Kaniska244 Done. This is simply a bugfix and doesn't add or remove features, so I bumped the patch version.

Kaniska244 (approved):

Kaniska244:
> @brianhelba please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

Hi @brianhelba 

Would you please accept the above license agreement.

brianhelba:
@Kaniska244 Sorry, I'm awaiting approval from my employer. I'm hoping this will be quick.

brianhelba:
@microsoft-github-policy-service agree company="Kitware, Inc."

brianhelba:
@Kaniska244 All done.

src/node/install.sh
```diff
@@ -386,13 +386,7 @@ if [ ! -z "${PNPM_VERSION}" ] && [ "${PNPM_VERSION}" = "none" ]; then
     echo "Ignoring installation of PNPM"
 else
     if bash -c ". '${NVM_DIR}/nvm.sh' && type npm >/dev/null 2>&1"; then
-        (
-            . "${NVM_DIR}/nvm.sh"
-            [ ! -z "$http_proxy" ] && npm set proxy="$http_proxy"
```
abdurriq, line 487: Others may need these proxies; could they be added back?

src/node/install.sh
```diff
@@ -386,13 +386,7 @@ if [ ! -z "${PNPM_VERSION}" ] && [ "${PNPM_VERSION}" = "none" ]; then
...
-            [ ! -z "$https_proxy" ] && npm set https-proxy="$https_proxy"
-            [ ! -z "$no_proxy" ] && npm set noproxy="$no_proxy"
-            npm install -g pnpm@$PNPM_VERSION --force
-        )
+        su ${USERNAME} -c "umask 0002 && . '${NVM_DIR}/nvm.sh' && npm install -g pnpm@${PNPM_VERSION} --force"
```
abdurriq, line 485: Are there any implications for switching to user-based vs root based? There may have been assumptions made by others based on this being root-based. If it could cause issues, we may want to consider adding an option and defaulting to the original approach.

abdurriq (changes requested):
Thank you for your contribution! I've left a couple comments which I would appreciate if you could review and address.

src/node/install.sh
brianhelba: From my commit message:
> Note, the explicit setting of proxy env vars (`http_proxy`, `https_proxy`, `no_proxy`) is likely a workaround for https://github.com/npm/cli/issues/6835. No other commands in this script use that workaround anymore, so this change uses the same su syntax as all other commands.

So, this doesn't change anything for users. They can still set `$http_proxy` and [npm will directly respect it](https://docs.npmjs.com/cli/v9/using-npm/config#https-proxy).

Every other install block works this way. The removed code just seems like leftovers of a workaround. If the workaround is actually necessary for some reason, than it's a bug that every other install block lacks it.

src/node/install.sh
brianhelba: Every other install block installs things as the user `USERNAME`, instead of root. I am not aware of any reason that pnpm's files should be owned by root, but everything else should be owned by `USERNAME`.

Further, this doesn't just cause pnpm-specific things to be owned by root, it can impact other common Node directories if this block runs first. From my commit message:
 >This causes the npm cache directory to be created owned by `root:root`, leading to `EACCES` errors for the non-root user on subsequent npm operations.

The unconditional ownership of some files by root is the essence of the bug. Features are (I believe) supposed to respect the active user (and every other part of this script does). A flag would be a "would you like to to fix the bug" option.

Users can continue to set `USERNAME` to root (by setting [`USER` in their Dockerfile](https://docs.docker.com/reference/dockerfile/#user) or using [other dev container features](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user)) if they want everything installed by the Node feature to be owned by root.

brianhelba:
@abdurriq I've left responses to all feedback. Could you please review?

brianhelba:
@abdurriq I've rebased this to fix merge conflicts.

I've also replied to all of your feedback. If you disagree with any of my responses and want me to make specific changes, please let me know. I'd like to get this bug fixed any way possible.

abdurriq (approved):
LGTM, I've just bumped the minor version to make it more clear that this could be a change that breaks existing config.

Issue #847, opened 2024-02-13
PHP feature should optionally install xdebug


razvanphp:
I personally want to use the (production) docker image for my app where I install PHP (and even composer) already, so I just need the option to install xdebug via the features flags, instead of baking this into my Dockerfile.

Would greatly expand the usability for PHP developers of devcontainers.

Thank you!
R

samruddhikhandale:
Hello 👋 

From 👇  looks like we already install `xdebug` with the PHP Feature. This is also baked into the `mcr.microsoft.com/devcontainers/php` image hosted from https://github.com/devcontainers/images/tree/main/src/php

https://github.com/devcontainers/features/blob/f91ff3cd8de5d33a52d21887963ba7166600cc44/src/php/install.sh#L186-L193, 

> [PHP feature should optionally install xdebug](https://github.com/devcontainers/features/issues/847#top)

@razvanphp Are you looking for an option which controls this `xdebug` installation? (as opposed to installing it only when you request a version)

Let me know if I am misunderstanding your question, thanks!

PR #1626, opened 2026-04-20
Enable support for many php extensions

### LINKED ISSUE #1332: Not possible to install PHP extensions
no options are available for choosing PHP extensions

mafiebig:
Enabling support for a number of high-value extensions. 

Resolves #1332 and #847

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 workspace.code-workspace
+.idea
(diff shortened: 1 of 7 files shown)


mafiebig:
@microsoft-github-policy-service agree [company="Highline College"]

mafiebig:
@microsoft-github-policy-service agree company="Highline College"

mafiebig:
Looking at the failing test, it doesn't seem like a valid test, but rather a copy and paste of some install code. Can I nuke that test so that all tests fire green?

PR #1632, opened 2026-05-02
Fix broken golangci-lint installation in go feature

max06:
Currently using the go-feature in devcontainers without specifying an older version to install breaks building devcontainers.

After investigating, 2 reasons were identified for breaking:
- golangci-lint now releases sbom-files with their binary releases. The install-script now finds 2 hashes matching the filename, which was corrected by the maintainers yesterday (and released today, 3 hours ago in https://github.com/golangci/golangci-lint/pull/6539).
- The golangci-lint project switched their primary branch from master to main, date irrelevant. Their fix for the first issue was properly merged into main, but the installation script for this container-feature was never updated, still pointing to master - missing the fix.

This should be merged quickly.

Potential workaround: Using `v2.11.4` as version for golangci-lint should work since sbom-files were added in v2.12 for the first time.

### DIFF
--- a/src/go/install.sh
+++ b/src/go/install.sh
@@ -325,11 +325,11 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
     # Install golangci-lint from precompiled binares
     if [ "$GOLANGCILINT_VERSION" = "latest" ] || [ "$GOLANGCILINT_VERSION" = "" ]; then
         echo "Installing golangci-lint latest..."
-        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/main/install.sh | \
             sh -s -- -b "${TARGET_GOPATH}/bin"
     else
         echo "Installing golangci-lint ${GOLANGCILINT_VERSION}..."
-        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/main/install.sh | \
             sh -s -- -b "${TARGET_GOPATH}/bin" "v${GOLANGCILINT_VERSION}"
     fi
 
--- a/src/go/install.sh
+++ b/src/go/install.sh
@@ -325,11 +325,11 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
     # Install golangci-lint from precompiled binares
     if [ "$GOLANGCILINT_VERSION" = "latest" ] || [ "$GOLANGCILINT_VERSION" = "" ]; then
         echo "Installing golangci-lint latest..."
-        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/main/install.sh | \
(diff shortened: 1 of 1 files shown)


src/go/install.sh
```diff
@@ -325,11 +325,11 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
     # Install golangci-lint from precompiled binares
     if [ "$GOLANGCILINT_VERSION" = "latest" ] || [ "$GOLANGCILINT_VERSION" = "" ]; then
         echo "Installing golangci-lint latest..."
-        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/main/install.sh | \
```
ldez:
```suggestion
        curl -fsSL https://golangci-lint.run/install.sh | \
```

src/go/install.sh
```diff
@@ -325,11 +325,11 @@ if [ "${INSTALL_GO_TOOLS}" = "true" ]; then
...
             sh -s -- -b "${TARGET_GOPATH}/bin"
     else
         echo "Installing golangci-lint ${GOLANGCILINT_VERSION}..."
-        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | \
+        curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/main/install.sh | \
```
ldez:
```suggestion
        curl -fsSL https://golangci-lint.run/install.sh | \
```

ldez (commented):
We (golangci-lint maintainers) recommend using `https://golangci-lint.run/install.sh` instead of an URL based of a branch.

https://golangci-lint.run/docs/welcome/install/local/#binaries

This is for several reasons:
- GitHub applies new [rate limits](https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/)
- Referencing a branch can be brittle

rjchien728:
Confirming this hits us on `linux/arm64` via `ghcr.io/devcontainers/features/go` with `golangciLintVersion: latest` (resolves to v2.12.1). Same `hash_sha256_verify` failure caused by the SBOM line in checksums.txt being matched alongside the tarball.
I tested the workaround curl -sSfL https://golangci-lint.run/install.sh | sh directly in our devcontainer — it works and produces the correct binary. So this PR's approach LGTM from a downstream user's perspective.

b-bremer (approved):
Thank you.

Sorry for the unnecessary approval.

alperyazar:
I can confirm that fixing linter to older version fixes the problem until this PR is merged.

```json
{
  "features": {
    "ghcr.io/devcontainers/features/go:1": {
      "golangciLintVersion": "2.11.4"
    }
  }
}
```

max06:
@bamurtaugh Good morning and sorry for the ping - could you please try to speed this up a bit? It shouldn't take days to merge a simple fix.

v-Kaniska244:
Thank you for the contribution. This has been already fixed by PR #1636. Closing the PR.

max06:
┻━┻︵ \(°□°)/ ︵ ┻━┻

Issue #1639, opened 2026-05-05
docker-in-docker: Feature request - Ability to pin containerd version


glennschmidt:
On apt-based systems, the feature script installs the `moby-engine` package, which depends on `moby-containerd >= 1.7.0`, so it will end up just installing the latest version of containerd available in the repo.

This week, containerd 2.3.0 was released, and it broke my dev containers (https://github.com/containerd/containerd/issues/13346). It would be good if there was a variable we could set to specify what version of `moby-containerd` to install, as a workaround when these types of issues occur.

Falco20019:
This is what I did to downgrade (on Ubuntu 24.04), just in case someone else also needs it:
`sudo apt-get update && sudo apt-get install -y --allow-downgrades moby-containerd=2.2.3-ubuntu24.04u1`

You can use `apt list --all-versions moby-containerd | grep 2.2.3` to find your variant.

mafuba:
+1, hitting this on `docker-in-docker` Feature with `moby-containerd 2.3.0-debian12u1` + `moby-engine 29.4.2-debian12u1`. Root cause is upstream containerd/containerd#13346 (EROFS-differ init failure cascades into dockerd refusing to start). Workaround we ended up with - a `postCreateCommand` that idempotently downgrades `moby-containerd` to `2.2.3-debian12u1`:

\`\`\`bash
PINNED="2.2.3-debian12u1"
if [[ "$(dpkg-query -W -f='${Version}' moby-containerd 2>/dev/null)" != "$PINNED" ]]; then
  sudo apt-get update -qq
  sudo apt-get install -y --allow-downgrades "moby-containerd=$PINNED"
  sudo pkill -9 dockerd containerd 2>/dev/null || true
  sudo rm -f /run/docker.pid
  sudo /usr/local/share/docker-init.sh &
fi
\`\`\`

Works, but a `containerdVersion` knob on the Feature would let us drop the workaround once containerd/containerd#13328 ships.

silva-renan:
+1 facing this issue. Setting "moby": false is also an workaround.

v-Kaniska244:
This [PR](https://github.com/devcontainers/features/pull/1645) has been merged to fix the issue with the `moby-containerd` package and erofs file system module. Please try the latest version of the `docker-in-docker` feature `ghcr.io/devcontainers/features/docker-in-docker:3` and get back in case of any issues.

JuJup:
[...]

PR #1645, opened 2026-05-11
fix(docker-in-docker): disable containerd erofs snapshotter to fix dockerd startup

### LINKED ISSUE #1642: docker-in-docker feature fails on macOS ARM64 with containerd 2.3.x unless erofs-utils is installed
## Summary

The `ghcr.io/devcontainers/features/docker-in-docker:2` feature failed to start the inner Docker daemon on macOS Apple Silicon with the `mcr.microsoft.com/devcontainers/dotnet:1-10.0` base image.

Inside the devcontainer, `docker ps` failed with:

```text
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
dockerd was launched, and containerd started, but dockerd timed out waiting for containerd:

failed to start containerd: timeout waiting for containerd to start
```
The relevant containerd log included:

```text
failed to check mkfs.erofs availability: failed to run mkfs.erofs --help: exec: "mkfs.erofs": executable file not found in $PATH: skip plugin
no plugins registered for io.containerd.transfer.v1: plugin: not found
```

Installing erofs-utils in the devcontainer image fixed the issue:
```text
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        protobuf-compiler \
        erofs-utils \
    && rm -rf /var/lib/apt/lists/*
```

## Environment
Host: macOS Apple Silicon
Docker Desktop: 4.72.0
Docker Engine: 29.4.2
Dev Containers extension: 0.459.0
VS Code: 1.119.0
Base image: mcr.microsoft.com/devcontainers/dotnet:1-10.0
Feature: ghcr.io/devcontainers/features/docker-in-docker:2
Architecture inside container: aarch64 / arm64

## Expected

The Docker-in-Docker feature should either install erofs-utils, avoid enabling a containerd path that requires mkfs.erofs, or document that base images must include erofs-utils for containerd 2.3.x.

## Actual

The inner Docker daemon does not become usable; docker ps reports that it cannot connect to /var/run/docker.sock.

## Workaround

Install erofs-utils in the Dockerfile before the feature starts dockerd.

Kaniska244:
## Summary

Fixes #1642 and #1639.

This PR fixes `docker-in-docker` startup failures caused by `containerd >= 2.3` probing the `erofs` snapshotter on hosts where the kernel exposes the `erofs` filesystem.

The main fix is to disable the `io.containerd.snapshotter.v1.erofs` plugin in `containerd` config and ensure `dockerd` actually uses that config by starting `containerd` explicitly and passing its socket to `dockerd`.

In addition, this PR includes a few related improvements and test updates in the `docker-in-docker` feature.

---

## Problem

On some hosts, `dockerd` fails to start because bundled `containerd` probes the `erofs` snapshotter plugin during startup.

That plugin requires `mkfs.erofs` support that is not available in older distro versions of `erofs-utils`, especially on:

- Debian 12 (`bookworm`)
- Ubuntu 22.04 (`jammy`)

When the host kernel exposes `erofs`, plugin initialization fails, `containerd` never becomes ready, and `dockerd` times out.

This is especially visible on newer environments such as:

- GitHub Actions arm64 runners
- Codespaces hosts
- systems where `erofs` has been loaded explicitly

Since the feature uses `overlayfs` anyway, the safest fix is to disable the `erofs` snapshotter entirely.

---

## Why the previous config-only approach was insufficient

A config-only change under `/etc/containerd/config.toml` was not enough because `dockerd`, when started normally, launches its own child `containerd` instance using an auto-generated config under `/var/run/docker/containerd/`.

That means `/etc/containerd/config.toml` is ignored unless we explicitly start `containerd` ourselves and point `dockerd` to it.

This PR therefore applies a two-part fix:

1. Write the `disabled_plugins` entry into `/etc/containerd/config.toml`
2. Start `containerd` explicitly and run `dockerd` with `--containerd /run/containerd/containerd.sock`

---

### 1. Disable the `erofs` snapshotter in `containerd`

In `src/docker-in-docker/install.sh`:

- ensure `/etc/containerd/config.toml` exists
- if missing or empty, generate it from `containerd config default`
- add `io.containerd.snapshotter.v1.erofs` to top-level `disabled_plugins`
- handle all config states safely:
  - existing empty list
  - existing populated list
  - no `disabled_plugins` key yet
- append an idempotency marker comment so repeated runs do not duplicate entries

This makes the configuration safe for re-runs and image-layer reuse.

---

### 2. Start `containerd` explicitly before `dockerd`

Also in `src/docker-in-docker/install.sh`, the generated init flow now:

- probes common locations for the `containerd` binary
- starts `containerd --config /etc/containerd/config.toml`
- waits for `/run/containerd/containerd.sock`
- if available, passes `--containerd /run/containerd/containerd.sock` to `dockerd`
- otherwise falls back to the previous behavior where `dockerd` spawns its own `containerd`

This keeps the fix additive and avoids breaking hosts where explicit `containerd` startup is not available.

The retry/cleanup flow also continues to clean up `containerd` alongside `dockerd`.

---

### 3. Add `erofs-utils` on Debian installs

In the Debian package path of `src/docker-in-docker/install.sh`, this PR adds:

- `erofs-utils`

This ensures `mkfs.erofs` is available where relevant and documents the dependency clearly, even though the primary fix is still disabling the snapshotter.

---

### 4. Fix Docker CE package download architecture handling for RHEL/tdnf path

In `src/docker-in-docker/install.sh`, the Docker CE RPM download logic now derives the correct repository architecture dynamically instead of hardcoding `x86_64`.

This adds support for architectures such as:

- `x86_64`
- `aarch64`

and updates the RPM lookup/download patterns accordingly.

This improves portability for non-x86 environments.

---

### 5. Install Docker Compose v1 in an isolated virtual environment

In `src/docker-in-docker/install.sh`, the Compose v1 installation path now:

- creates a dedicated Python virtual environment
- installs `pip`, `setuptools`, and `wheel` into that venv
- installs `docker-compose` and dependencies there
- symlinks the resulting `docker-compose` binary into the expected path

This avoids PEP 668 / externally-managed-environment issues on newer distros such as:

- Debian trixie
- Ubuntu noble

and avoids modifying distro-managed Python site-packages.

---

### Workflow update

In `.github/workflows/test-pr-arm64.yaml`:

- include `src/docker-in-docker/**` and `test/docker-in-docker/**` in workflow path triggers
- add a `docker-in-docker` filter in changed-path detection
- add a step to explicitly load and verify the `erofs` kernel module:

```sh
sudo modprobe erofs && grep erofs /proc/filesystems
```

This makes the failure scenario reproducible in CI.

There is also a matrix exclusion added for:

- `docker-in-docker` + `mcr.microsoft.com/devcontainers/base:debian`

---

### Test script updates

The following test scripts now explicitly verify that Docker is actually usable by adding `docker ps` checks:

- `test/docker-in-docker/dockerIp6tablesDisabledTest.sh`
- `test/docker-in-docker/docker_build_older.sh`
- `test/docker-in-docker/pin_docker-ce_version_moby_false.sh`

This strengthens validation by confirming that `dockerd` not only starts, but is usable.

---

## Docs / metadata updates

This PR also updates `docker-in-docker` feature metadata and documentation:

### `src/docker-in-docker/devcontainer-feature.json`
- bump feature version from `2.17.0` to `3.0.0`

### `src/docker-in-docker/README.md`
- update example reference from:
  - `ghcr.io/devcontainers/features/docker-in-docker:2`
  - to `ghcr.io/devcontainers/features/docker-in-docker:3`

### `src/docker-in-docker/NOTES.md`
- update the sample image tag from:
  - `mcr.microsoft.com/devcontainers/typescript-node:16`
  - to `mcr.microsoft.com/devcontainers/typescript-node:24`

---

## Files changed

- `.github/workflows/test-pr-arm64.yaml`
- `src/docker-in-docker/NOTES.md`
- `src/docker-in-docker/README.md`
- `src/docker-in-docker/devcontainer-feature.json`
- `src/docker-in-docker/install.sh`
- `test/docker-in-docker/dockerIp6tablesDisabledTest.sh`
- `test/docker-in-docker/docker_build_older.sh`
- `test/docker-in-docker/pin_docker-ce_version_moby_false.sh`

---

## Compatibility / risk

- No behavior change when the `erofs` snapshotter is not relevant
- Fallback behavior remains unchanged if explicit `containerd` startup is unavailable
- Config modification is idempotent and safe across repeated runs
- No new external dependency flow is introduced beyond packages already used by the feature
- Docker continues to use the expected `overlayfs` path

---

## Result

This PR makes `docker-in-docker` startup more reliable across affected Debian/Ubuntu environments, especially on arm64 and newer hosts where `erofs` is exposed, while also improving test coverage, compose installation robustness, and architecture handling.

---

## Related

- Closes #1642
- Related to #1639

### DIFF
--- a/.github/workflows/test-pr-arm64.yaml
+++ b/.github/workflows/test-pr-arm64.yaml
@@ -45,6 +45,9 @@ jobs:
             "mcr.microsoft.com/devcontainers/base:debian",
             "mcr.microsoft.com/devcontainers/base:noble"
           ]
+        exclude:
+          - features: docker-in-docker
+            baseImage: mcr.microsoft.com/devcontainers/base:debian           
     steps:
       - uses: actions/checkout@v6
 
--- a/.github/workflows/test-pr-arm64.yaml
+++ b/.github/workflows/test-pr-arm64.yaml
@@ -51,6 +51,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: "Load erofs module and verify"
+        run: sudo modprobe erofs && grep erofs /proc/filesystems
+
       - name: "Install latest devcontainer CLI"
         run: npm install -g @devcontainers/cli
 
--- a/.github/workflows/test-pr-arm64.yaml
+++ b/.github/workflows/test-pr-arm64.yaml
@@ -8,6 +8,8 @@ on:
     paths:
       - "src/powershell/**"
       - "test/powershell/**"
+      - "src/docker-in-docker/**"
+      - "test/docker-in-docker/**"      
 
 jobs:
   detect-changes:
@@ -23,6 +25,7 @@ jobs:
           #   <feature-name>: ./**/<feature-name>/**
           filters: |
             powershell: ./**/powershell/**
+            docker-in-docker: ./**/docker-in-docker/**
 
   test:
     needs: [detect-changes]
--- a/src/docker-in-docker/NOTES.md
+++ b/src/docker-in-docker/NOTES.md
@@ -4,7 +4,7 @@ This docker-in-docker Dev Container Feature is roughly based on the [official do
 * As the name implies, the Feature is expected to work when the host is running Docker (or the OSS Moby container engine it is built on). It may be possible to get running in other container engines, but it has not been tested with them.
 * The host and the container must be running on the same chip architecture. You will not be able to use it with an emulated x86 image with Docker Desktop on an Apple Silicon Mac, like in this example:
   ```
-  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:16
+  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:24
   ```
   See [Issue #219](https://github.com/devcontainers/features/issues/219) for more details.
 
--- a/src/docker-in-docker/NOTES.md
+++ b/src/docker-in-docker/NOTES.md
@@ -4,7 +4,7 @@ This docker-in-docker Dev Container Feature is roughly based on the [official do
 * As the name implies, the Feature is expected to work when the host is running Docker (or the OSS Moby container engine it is built on). It may be possible to get running in other container engines, but it has not been tested with them.
 * The host and the container must be running on the same chip architecture. You will not be able to use it with an emulated x86 image with Docker Desktop on an Apple Silicon Mac, like in this example:
   ```
-  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:16
+  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:22
   ```
   See [Issue #219](https://github.com/devcontainers/features/issues/219) for more details.
 
--- a/src/docker-in-docker/NOTES.md
+++ b/src/docker-in-docker/NOTES.md
@@ -4,7 +4,7 @@ This docker-in-docker Dev Container Feature is roughly based on the [official do
 * As the name implies, the Feature is expected to work when the host is running Docker (or the OSS Moby container engine it is built on). It may be possible to get running in other container engines, but it has not been tested with them.
 * The host and the container must be running on the same chip architecture. You will not be able to use it with an emulated x86 image with Docker Desktop on an Apple Silicon Mac, like in this example:
   ```
-  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:22
+  FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/typescript-node:16
   ```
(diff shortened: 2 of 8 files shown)


.github/workflows/test-pr-arm64.yaml
```diff
@@ -42,9 +45,15 @@ jobs:
             "mcr.microsoft.com/devcontainers/base:debian",
             "mcr.microsoft.com/devcontainers/base:noble"
           ]
+        exclude:
+          - features: docker-in-docker
+            baseImage: mcr.microsoft.com/devcontainers/base:debian
```
abdurriq, line 50: Nit: trailing whitespace

abdurriq (approved):

PR #1649, opened 2026-05-14
fix(ruby): accept both "_" and "." tag separators from ruby/ruby

rhiroe:
ruby/ruby switched its release tag separator from "\_" (e.g. v3_4_9) to "." (e.g. v4.0.4) starting at Ruby 4.0.0. The current regex in find_version_from_git_tags only matches the "\_" form, so VERSION=latest and ADDITIONAL_VERSIONS resolution silently skip every 4.x release.

Normalize the tag list with `tr "_" "."` before matching and use "." canonically in the regex. The `separator` parameter is no longer consulted for matching, so drop the now-redundant `"_"` argument from the three ruby call sites.

### DIFF
--- a/src/ruby/devcontainer-feature.json
+++ b/src/ruby/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "ruby",
-    "version": "1.3.2",
+    "version": "1.3.3",
     "name": "Ruby (via rvm)",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/ruby",
     "description": "Installs Ruby, rvm, rbenv, common Ruby utilities, and needed dependencies.",
--- a/src/ruby/install.sh
+++ b/src/ruby/install.sh
@@ -147,17 +147,17 @@ find_version_from_git_tags() {
     local repository=$2
     local prefix=${3:-"tags/v"}
     local separator=${4:-"."}
-    local last_part_optional=${5:-"false"}    
+    local last_part_optional=${5:-"false"}
     if [ "$(echo "${requested_version}" | grep -o "." | wc -l)" != "2" ]; then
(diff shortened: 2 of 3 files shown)


rhiroe:

Kaniska244 (changes requested):
Hi @rhiroe 

Thank you for the contribution. Would you kindly bump the feature version and also add relevant test cases if applicable for this change. Also please check the failing tests.

rhiroe:
The CI failure is caused by RVM's stable branch not including Ruby 4.x in its known rubies list (it currently tops out at `ruby-3.0.0`). Ruby 4.x is already present in RVM's master branch but has not yet been merged into stable.
 
Because `install.sh` uses `curl -sSL https://get.rvm.io | bash -s stable`, the installed RVM does not recognize `4.0.4` as a known MRI ruby. As a result, the default alias is created as `4.0.4` instead of `ruby-4.0.4`, producing a dangling symlink at `/usr/local/rvm/rubies/default`. This causes `which gem` to return nothing, and `${ROOT_GEM} install rake` expands to `install rake` — invoking the system install command instead of gem install.
 
I'm not sure what the best path forward is here. A few options come to mind:
 
- Add `rvm use default` — resolves the issue immediately, though this doesn't address the underlying cause.
- Wait for RVM's stable branch to include Ruby 4.x — would address the root cause, though the timeline depends on the RVM project.
- Switch from `stable` to `master` in the RVM install command — would bring in Ruby 4.x support now, though it's worth considering whether using master is appropriate for this environment.

Happy to hear any thoughts on which direction makes sense.

rhiroe:
Once #1654 is merged, the issue will be resolved and this PR can be closed.

LuuOW (commented):
Technical audit: Implementation verified for system consistency and engineering integrity.

PR #1654, opened 2026-05-19
feat(ruby)!: rewrite to use ruby-build with optional rbenv/rvm (v2.0.0)

rhiroe:
## Summary

This PR is a continuation of #764 (now closed), which itself was opened in response to #757 and #603. It rewrites the Ruby Feature on top of **ruby-build** and bumps the version to **2.0.0**.

### 1. RVM's `stable` channel does not support recent Ruby versions

RVM's `stable` branch lags behind the upstream Ruby release cycle. As a result, users who need a recent Ruby version (e.g. 3.3.x or later) often cannot install it through the current Feature without workarounds.

### 2. Both RVM and rbenv are installed unconditionally

The current Feature installs both RVM and rbenv regardless of what the user actually needs. This causes conflicts — in particular, rbenv is effectively unusable when RVM is present because `GEM_HOME`, `GEM_PATH`, and `MY_RUBY_HOME` are all hard-coded to RVM paths in `containerEnv`. Any version manager the user adds on top will be broken by those static environment variables.

### 3. A version manager is not always necessary

For the common case where a project targets a single, fixed Ruby version, there is no need for a version manager at all. Installing RVM by default adds unnecessary complexity, startup overhead, and a frequent source of user confusion.

## What this PR does

- **Replaces RVM/rbenv with ruby-build as the sole installer.** Ruby is installed to `/usr/local/rubies/<version>/` and exposed via a `/usr/local/rubies/current` symlink on the `PATH`.
- **Removes `GEM_HOME`, `GEM_PATH`, and `MY_RUBY_HOME` from `containerEnv`.** These variables interfere with any version manager or gem tooling the user may add. RubyGems' own dynamic defaults work correctly when these are left unset.
- **Adds a `versionManager` option** (`none` / `rbenv` / `rvm`, default `none`). Users who need a version manager can opt in explicitly. When `rbenv` or `rvm` is selected, ruby-build is wired in as its build back-end.
- **Broadens OS support** from Debian/Ubuntu (`apt`) to also cover RHEL/Fedora (`dnf`/`yum`), Alpine (`apk`), openSUSE (`zypper`), and Arch (`pacman`).
- **Adds a shared `ruby` group** owning `/usr/local/rubies/` so the non-root container user can run `gem install` without `sudo`.
- Updates test scenarios to cover the `rbenv` and `rvm` paths, and removes the now-unnecessary `ruby_fallback_test`.

## Relationship to previous work

This PR picks up directly from #764. The approach here aligns with the consensus reached in #757: ruby-build as the default, version managers as opt-in, and no static `GEM_HOME`/`GEM_PATH`.

## Breaking changes

- `GEM_HOME`, `GEM_PATH`, and `MY_RUBY_HOME` are no longer set by the Feature. Projects that relied on these values being set to RVM paths will need to update their configuration.
- The Feature version is bumped to **2.0.0** to signal the backward-incompatible change.
- RVM is no longer installed by default. Use `"versionManager": "rvm"` to restore the previous behaviour.


src/ruby/install.sh
```diff
@@ -56,389 +57,341 @@ elif [ "${USERNAME}" = "none" ] || ! id -u ${USERNAME} > /dev/null 2>&1; then
...
+# Called before ruby versions are installed so that install_ruby_version()
+# can delegate to 'rvm install'.
+install_rvm() {
+    echo "Installing rvm..."
+
+    # Best-effort GPG key import; some networks block keyservers.
+    gpg --keyserver hkp://keyserver.ubuntu.com \
+        --recv-keys \
+        409B6B1796C275462A1703113804BB82D39DC0E3 \
+        7D2BAF1CF37B13E2069D6956105BD0E739499BDB 2>/dev/null \
```
Kaniska244: Hi @rhiroe 

Thank you for the contribution. Would it be possible for you to put the keys in a constant like the old code had,  instead of being literals inside the function. Also wouldn't it have been better to retain the retry logic and the keyserver logic around the GPG key import? 
Kindly let me know in case of any concern.

test/ruby/scenarios.json
```diff
@@ -44,12 +44,22 @@
...
+        "image": "ubuntu:noble",
+        "features": {
+            "ruby": {
+                "version": "3.4.2",
+                "versionManager": "rbenv"
+            }
+        }
+    },
+    "ruby_rvm": {
         "image": "mcr.microsoft.com/devcontainers/base:bullseye",
```
Kaniska244: Could this be upgraded to bookworm please?

test/ruby/scenarios.json
```diff
@@ -44,12 +44,22 @@
             "ruby": {}
         }
     },
-    "ruby_fallback_test": {
+    "ruby_rbenv": {
+        "image": "ubuntu:noble",
```
Kaniska244: Could we use `mcr.microsoft.com/devcontainers/base:noble` as base image instead?

Kaniska244 (changes requested):

src/ruby/install.sh
rhiroe: Addressed in 07e07de75dda34c364ceddcf9efa7dc78811fed1

test/ruby/scenarios.json
rhiroe: Addressed in 07e07de75dda34c364ceddcf9efa7dc78811fed1

test/ruby/scenarios.json
rhiroe: Addressed in 07e07de75dda34c364ceddcf9efa7dc78811fed1

rhiroe:
@Kaniska244 Thanks for the review! I've addressed your feedback in 07e07de75dda34c364ceddcf9efa7dc78811fed1. Note that the commit was squashed via fixup to keep the commit log clean.

Kaniska244 (dismissed):

src/ruby/NOTES.md
```diff
@@ -2,6 +2,13 @@
...
+This Feature supports Linux images that ship one of the following package managers: `apt`, `dnf`/`yum`, `apk`, `zypper`, or `pacman`. The script detects the available package manager and installs the build dependencies that ruby-build needs.
 
 `bash` is required to execute the `install.sh` script.
+
+## Layout
+
+- Ruby is installed under `/usr/local/rubies/<version>` by ruby-build.
+- The default Ruby is exposed via the `/usr/local/rubies/current` symlink, which is placed on the `PATH` through `containerEnv`.
+- `ruby-build` itself is cloned to `/usr/local/share/ruby-build` and symlinked into `/usr/local/bin/ruby-build` so additional versions can be installed later.
+- A shared `ruby` group owns `/usr/local/rubies`; the configured non-root user is added to it so `gem install` can write into the active Ruby tree without `sudo`.
```
abdurriq, line 14: Is this consistent with what we had before? If not, I'm curious if this is a standard, as it seems a bit risky to have all users write to the same gems path vs. having their own local ones.

test/ruby/install_additional_ruby_trixie.sh
abdurriq: Nit: This file could be a symlink to test/ruby/install_additional_ruby.sh

src/ruby/NOTES.md
abdurriq: Ah nvm, I see this is similar to
```
chown -R "${USERNAME}:rvm" "/usr/local/rvm/"
chmod -R g+r+w "/usr/local/rvm/"
find "/usr/local/rvm/" -type d | xargs -n 1 chmod g+s
```

abdurriq (changes requested):

test/ruby/install_additional_ruby_trixie.sh
rhiroe: Done — replaced `install_additional_ruby_trixie.sh` with a symlink to `install_additional_ruby.sh`.
01a95a97b6fe6f8f4a5689d2636546e35368badf

abdurriq (approved):

PR #1666, opened 2026-05-27
[docker-in-docker] - Move the iptables switching logic in the docker-init script and isolated tests for specific cases

### LINKED ISSUE #1659: [docker-in-docker] `docker-in-docker` fails on Ubuntu 24.04 containers on RHEL/AlmaLinux hosts: codename-based iptables fallback in v2.17.0 is too narrow
The fix introduced in PR #1637 (v2.17.0) for iptables-legacy failures on Ubuntu 26.04 uses a hardcoded `VERSION_CODENAME` check to decide whether to use `iptables-nft` vs `iptables-legacy`. This only handles Ubuntu 26.04 (`resolute`), but the same failure occurs on any Ubuntu/Debian container image running on a host kernel where the `ip_tables` module is not loaded, for example, AlmaLinux 9 or Fedora 41+ hosts.

### Steps to Reproduce

- Host OS: **AlmaLinux 9** (or Fedora 41+): kernel does not load `ip_tables` by default
- Devcontainer base image: `mcr.microsoft.com/devcontainers/base:ubuntu-24.04` (codename: `noble`)
- Feature: `ghcr.io/devcontainers/features/docker-in-docker:2` (v2.17.0 / v3.x)

Starting the devcontainer results in:

```
failed to start daemon: Error initializing network controller: error obtaining controller instance:
failed to register "bridge" driver: failed to create NAT chain DOCKER: iptables failed:
iptables --wait -t nat -N DOCKER: iptables v1.8.10 (legacy): can't initialize iptables table `nat':
Table does not exist (do you need to insmod?)
```

Confirming manually inside the container:

```sh
$ iptables-legacy -nL
iptables v1.8.10 (legacy): can't initialize iptables table `filter': Table does not exist (do you need to insmod?)
# exits with code 3

$ type iptables-legacy
iptables-legacy is hashed (/usr/sbin/iptables-legacy)
# exits with code 0  <- binary exists, but kernel module is absent
```

### Root Cause

In PR #1637, the initial fix correctly used a runtime `-nL` probe to verify whether `iptables-legacy` actually works before switching alternatives. However, this probe was removed because `install.sh` runs at image build time, when kernel modules are not accessible, so the probe would always fail during build.

The replacement logic uses a hardcoded codename allowlist:

```sh
case "${VERSION_CODENAME}" in
    resolute) use_nft=true ;;
esac
```

This only catches Ubuntu 26.04 (`resolute`). On Ubuntu 24.04 (`noble`) and any other non-`resolute` image the script falls through to:

```sh
elif type iptables-legacy > /dev/null 2>&1; then
    update-alternatives --set iptables /usr/sbin/iptables-legacy
```

Since `type iptables-legacy` succeeds (the binary exists), `iptables-legacy` is set as the default, even though the host kernel does not have `ip_tables` loaded. The failure only manifests at container runtime when `dockerd` tries to create the NAT chain.

This is the same underlying problem described in issue #1235 (Fedora/RHEL hosts), which also remains unresolved by the v2.17.0 fix.

### Proposed Fix

The iptables alternatives selection logic should be moved from `install.sh` (build-time) to `docker-init.sh` (the runtime entrypoint script). At runtime, the `-nL` probe works correctly and can detect the actual kernel capability:

```sh
# In docker-init.sh, before starting dockerd:
if type iptables-legacy > /dev/null 2>&1 && iptables-legacy -nL > /dev/null 2>&1; then
    update-alternatives --set iptables /usr/sbin/iptables-legacy
    update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
elif type iptables-nft > /dev/null 2>&1; then
    update-alternatives --set iptables /usr/sbin/iptables-nft
    update-alternatives --set ip6tables /usr/sbin/ip6tables-nft
fi
```

This approach is host-kernel-agnostic and does not require maintaining a hardcoded list of codenames.

### Environment

|                  |                                                                      |
|------------------|----------------------------------------------------------------------|
| Host OS          | AlmaLinux 9                                                          |
| Container OS     | Ubuntu 24.04.3 LTS (`noble`)                                         |
| Feature version  | `ghcr.io/devcontainers/features/docker-in-docker:2` (v2.17.0 / v3.x) |
| Docker version   | 29.4.3-1                                                             |
| iptables version | v1.8.10 (legacy)                                                     |

### Related Issues / PRs

- PR #1637: introduced the v2.17.0 fix (codename-based approach), which only covers Ubuntu 26.04
- Issue #1235: same failure on Fedora 41+ hosts; also not fixed by v2.17.0
- Issue #1633: original Ubuntu 26.04 report that triggered #1637

### LINKED ISSUE #1662: [docker-in-docker] v2 and v3 feature fails now for previously perfectly working devcontainer configurations
## base system:
   
- Ubuntu 26.04 LTS
- Host Docker Engine Community 29.5.0, API 1.54 (min API 1.40)

## devcontainer configuration

```json
{
    "name": "morbyd (docker-in-docker)",
    "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
    "features": {
        "ghcr.io/devcontainers/features/docker-in-docker:3": {
            "version": "latest",
            "moby": false // go for the upstream Docker-CE
        },
        //"ghcr.io/thediveo/devcontainer-features/local-pkgsite:0": {},
        //"ghcr.io/thediveo/devcontainer-features/goreportcard:0": {},
        //"ghcr.io/thediveo/devcontainer-features/go-mod-upgrade:0": {},
        //"ghcr.io/thediveo/devcontainer-features/lazygit:0": {},
        //"ghcr.io/thediveo/devcontainer-features/gocover:0": {
        //    "tags": "matchers",
        //    "uncovered-packages": "github.com/thediveo/morbyd/v2/run/dockercli",
        //    "num-programs": "1",
        //    "race": true,
        //    "verbose": true,
        //    "html": true
        //},
        //"ghcr.io/thediveo/devcontainer-features/pin-github-action:0": {}
    },
    "remoteEnv": {
        "GOPATH": "/home/vscode/go",
        "PATH": "/home/vscode/go/bin:/go/bin:/usr/local/go/bin:${localEnv:PATH}"
    },
    "postCreateCommand": "go install go.uber.org/mock/mockgen@latest",
    "customizations": {
        "vscode": {
            "extensions": [
                "stkb.rewrap",
                "brunnerh.insert-unicode",
                "mhutchie.git-graph",
                "ms-vscode.makefile-tools",
                "redhat.vscode-yaml"
            ]
        }
    }
}
```

As you can see from the above configuration, still using the 24.04 base image here. Inside the devcontainer, `docker ps` fails with `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`

`ps faux` shows this interesting part:

```
root           1  0.0  0.0   1168   780 ?        Ss   16:34   0:00 /sbin/docker-init -- /bin/sh -c echo Container started trap "exit 0" 15 /usr/local/share/docker-init.sh exec "$@" while sleep 1 & wait $!; do :; done -
root           7  0.0  0.0   2808  1908 ?        S    16:34   0:00 /bin/sh -c echo Container started trap "exit 0" 15 /usr/local/share/docker-init.sh exec "$@" while sleep 1 & wait $!; do :; done -
root        8290  0.0  0.0   2704  1616 ?        S    16:39   0:00  \_ sleep 1
root        7216  0.0  0.0 2754320 46048 ?       Sl   16:35   0:00 /usr/bin/containerd --config /etc/containerd/config.toml
```

So, `containerd` runs and responds to `sudo ctr version`, but `dockerd` is missing.

I've removed all containers, images and volumes, but to no avail. Always same problem.

Disabling all features except for `docker-in-docker` still shows the failure.

Kaniska244:
## Summary

Fixes #1659 and fixes #1662.

This PR updates the `docker-in-docker` feature to support selecting the `iptables` backend at container runtime instead of only during image build. This is needed for hosts where the correct backend depends on the host kernel state at runtime, such as systems where the `ip_tables` kernel module is not loaded and `iptables-legacy` exists but does not work.

It also updates Docker Compose handling so `dockerDashComposeVersion` supports `latest` and uses it as the default.

## Changes

- Adds a new `docker-in-docker` option:
  - `iptablesSwitchAtRuntime`:
    - Default: `false`
    - When enabled, the feature writes the `iptables` alternatives selection logic into `docker-init.sh` so it runs before `dockerd` starts.
- Keeps the existing install-time behavior when `iptablesSwitchAtRuntime` is not enabled.
- Updates the runtime selection logic to:
  - Prefer `iptables-legacy` only when the legacy backend is actually available through the host kernel.
  - Fall back to `iptables-nft` when legacy support is unavailable.
- Updates Docker Compose version support:
  - Adds `latest` as a valid `dockerDashComposeVersion` value.
  - Changes the default `dockerDashComposeVersion` from `v2` to `latest`.
  - Updates `install.sh`, `devcontainer-feature.json`, and the README option table accordingly.
- Updates the `docker-in-docker` README with the new/updated options.
- Bumps the `docker-in-docker` feature version to `3.1.0`.

## Test Coverage

Adds isolated test scenarios for:

- Build-time iptables switching behavior.
- Runtime iptables switching behavior.
- Debian-based containers with `ip_tables` loaded.
- Debian-based containers without `ip_tables` loaded.
- Ubuntu-based containers with `ip_tables` loaded.
- Ubuntu-based containers without `ip_tables` loaded.
- Stress test coverage for `iptablesSwitchAtRuntime=true`.
- Docker Compose `latest` behavior with Moby enabled.
- Docker Compose `latest` behavior with Moby disabled.

Also updates CI to isolate the iptables-specific scenarios so they do not interfere with the standard `docker-in-docker` scenario test matrix.

## Validation

The new tests verify that:

- `docker-init.sh` contains the runtime switching block only when `iptablesSwitchAtRuntime=true`.
- The selected `iptables` alternative is either `iptables-legacy` or `iptables-nft` as appropriate.
- `iptables` commands work.
- `dockerd` starts successfully.
- `docker ps` works inside the dev container.
- Docker daemon logs show successful initialization.
- `docker compose version` returns a valid version when `dockerDashComposeVersion=latest`.
- `docker-compose --version` returns a valid version when `dockerDashComposeVersion=latest`.
- `/usr/local/bin/docker-compose` is installed for the `latest` Docker Compose scenario.


SrzStephen:
LGTM @ `7f2a5ed77143a90f735a7b6c799ac0d9cb44b55f` but you pushed while I was typing ;) 

I don't see anything wrong with [7360873](https://github.com/devcontainers/features/pull/1666/commits/7360873729d711c85bc3932b8a854ffd0f6a9e91) if the tests pass

There are cases where `iptables` can be loaded but `iptables nat` (same case for iptables-legacy) isn't... I can't think of a decent reason why this would come into play.

If you can think of one then it might be worth checking for iptables nat - I cant remember what case I ran into since it was ~ a month ago, entirely possible that I was doing something silly.


Example:
Both available
```
docker run --rm -it --cap-add=NET_ADMIN debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test ==' && iptables -t nat -L"
```
iptable-nat not available.

```
docker run --rm -it debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test (expected to fail) ==' && iptables -t nat -L || echo 'NAT not available or cannot access kernel tables'"
```

v-Kaniska244:
> LGTM @ `7f2a5ed77143a90f735a7b6c799ac0d9cb44b55f` but you pushed while I was typing ;)
> 
> I don't see anything wrong with [7360873](https://github.com/devcontainers/features/pull/1666/commits/7360873729d711c85bc3932b8a854ffd0f6a9e91) if the tests pass
> 
> There are cases where `iptables` can be loaded but `iptables nat` (same case for iptables-legacy) isn't... I can't think of a decent reason why this would come into play.... If you can think of one then it might be worth checking for iptables nat - I cant remember what case I ran into since it was ~ a month ago, entirely possible that I was doing something silly.
> 
> Example: Both available
> 
> ```
> docker run --rm -it --cap-add=NET_ADMIN debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test ==' && iptables -t nat -L"
> ```
> 
> iptable-nat not available.
> 
> ```
> docker run --rm -it debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test (expected to fail) ==' && iptables -t nat -L || echo 'NAT not available or cannot access kernel tables'"
> ```

Hi @SrzStephen 

Thank you for the review. Let me have a look for that scenario.

src/docker-in-docker/devcontainer-feature.json
```diff
@@ -60,6 +61,11 @@
       "type": "boolean",
       "default": false,
       "description": "Disable ip6tables (this option is only applicable for Docker versions 27 and greater)"
+    },
+    "iptablesSwitchAtRuntime": {
```
SrzStephen, line 65: Probably the only thing I'd try to make a case for, but I'm not overly familiar with this repos general policy here so if it tends to more conservative then that's ok:

If you're moving to a v4 release it might be worth making this a `default: true` since you're already cutting a major rev.

Reason: The only time it matters what iptables is present is at runtime so it's probably the correct approach, someone can always opt into the old behaviour if they wish (by setting it to false) but you'll probably see some "why doesn't docker run" issues similar to https://github.com/devcontainers/features/issues/1659.

Kaniska244: Makes sense. I would much rather put it back to version 3 as the flag could help doing this seamlessly.

thediveo: I'm wondering since 3.0.0 and 3.0.1 break on Ubuntu 24.04 devcontainer base images on a nUbuntu 26.04 host anyway and my testing indicates that the 3.1.0/4.0.0 feature works also on an Ubuntu 22.04 why not considering this a fix with a new compatible feature with the default of `iptablesSwitchAtRuntime:true`? This makes the 3.x feature working; are there any side effects of the iptablesSwitchAtRuntime?

thediveo:
can this please be reviewed and moved forward in production, as it is an important fix?

test/docker-in-docker/docker_with_default_iptables.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
abdurriq, line 1: Isn't this identical to test/docker-in-docker/docker_without_iptables_ubuntu.sh except for log check logic? Can we make one a symlink of the other?

test/docker-in-docker/docker_without_iptables.sh
```diff
@@ -0,0 +1,34 @@
...
+    if ! grep -q "Daemon has completed initialization" /tmp/dockerd.log; then
+        echo "❌ Docker daemon failed to start. Last errors from /tmp/dockerd.log:"
+        echo "----- dockerd.log (tail) -----"
+        tail -n 100 /tmp/dockerd.log
+        echo "----- error/fatal lines -----"
+        grep -iE "error|fatal|failed|panic" /tmp/dockerd.log || true
+        exit 1
+    fi
+'
+check "log-for-completion" bash -c "cat /tmp/dockerd.log | grep 'Daemon has completed initialization'"
```
abdurriq: Nit: Duplicate of line 29

test/docker-in-docker/scenarios.json
```diff
@@ -1,4 +1,63 @@
...
+        "image": "mcr.microsoft.com/devcontainers/base:debian",
+        "features": {
+            "docker-in-docker": {
+                "moby": "false",
+                "iptablesSwitchAtRuntime": true
+            }
+        },
+        "initializeCommand": "sudo modprobe ip_tables"
+    },
+    "docker_without_iptables": {
```
abdurriq: Nit: should be something like `with_legacy_iptables` or `with_default_iptables`; as both still have iptables

test/docker-in-docker/docker_with_legacy_iptables.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
abdurriq, line 1: Same here, identical to the ubuntu version; can this be a symlink?

test/docker-in-docker/docker_stress_iptables_runtime.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
abdurriq: "Stress" doesn't seem applicable here, could we call this `docker_iptables_switch_at_runtime.sh` or something more relevant instead?

src/docker-in-docker/install.sh
```diff
@@ -970,6 +976,29 @@ DOCKER_DEFAULT_ADDRESS_POOL=${DOCKER_DEFAULT_ADDRESS_POOL}
 DOCKER_DEFAULT_IP6_TABLES=${DOCKER_DEFAULT_IP6_TABLES}
 EOF
 
+# On Debian-based images, re-assert the iptables alternative at container start
+# (only when the user opted into runtime switching via iptablesSwitchAtRuntime=true).
+if [ "${IPTABLES_SWITCH_AT_RUNTIME}" = "true" ] && [ "${ADJUSTED_ID}" = "debian" ]; then
```
abdurriq, line 981: Nit: Can this be an else branch of the above if block? It would be nice to have these be together rather than hundreds of lines apart.

.github/workflows/test-pr.yaml
```diff
@@ -92,5 +92,42 @@ jobs:
       - name: "Install latest devcontainer CLI"
         run: npm install -g @devcontainers/cli
 
+      - name: "Exclude iptables-isolation scenarios from docker-in-docker"
```
abdurriq: Nit: Can we make it clear that these will run under a separate job?

src/docker-in-docker/install.sh
Kaniska244: Hi @abdurriq 

This logic is separated from the existing logic because here its not executing the commands rather writing them into the `docker-init.sh` which is written towards the end of the installation script in a block as per the existing logic. I could technically move this part alone above much above near the old `iptables` switching logic and later append the remaining part as -is towards the end of the `install.sh`. My only concern is that it could impact the readability of the `docker-init.sh` snippet which right now is being formed almost as a block as per the current code.
Kindly let me know if we should still change this. I will go ahead accordingly.

.github/workflows/test-pr.yaml
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/docker_stress_iptables_runtime.sh
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/docker_with_legacy_iptables.sh
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/scenarios.json
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/docker_without_iptables.sh
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/docker_with_default_iptables.sh
Kaniska244: Hi @abdurriq 

This has been implemented. Kindly let me know in case of further concern.

.github/workflows/docker-in-docker-stress-test.yaml
abdurriq: Could we please also change the "stress" part here as it's not accurate?

test/docker-in-docker/docker_with_legacy_iptables.sh
abdurriq, line 1: Nit: `docker_with_legacy_iptables` would be a more accurate name

src/docker-in-docker/install.sh
abdurriq: Ah I see, in that case we can leave it as-is, thanks for clarifying.

v-Kaniska244:
> LGTM @ `7f2a5ed77143a90f735a7b6c799ac0d9cb44b55f` but you pushed while I was typing ;)
> 
> I don't see anything wrong with [7360873](https://github.com/devcontainers/features/pull/1666/commits/7360873729d711c85bc3932b8a854ffd0f6a9e91) if the tests pass
> 
> There are cases where `iptables` can be loaded but `iptables nat` (same case for iptables-legacy) isn't... I can't think of a decent reason why this would come into play.
> 
> If you can think of one then it might be worth checking for iptables nat - I cant remember what case I ran into since it was ~ a month ago, entirely possible that I was doing something silly.
> 
> Example: Both available
> 
> ```
> docker run --rm -it --cap-add=NET_ADMIN debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test ==' && iptables -t nat -L"
> ```
> 
> iptable-nat not available.
> 
> ```
> docker run --rm -it debian:bookworm-slim bash -c "apt-get update -qq && apt-get install -y iptables >/dev/null && echo '== iptables version ==' && iptables --version && echo '== NAT test (expected to fail) ==' && iptables -t nat -L || echo 'NAT not available or cannot access kernel tables'"
> ```

Hi @SrzStephen 

Thank you again for this comment. The [runtime block](https://github.com/Kaniska244/features/blob/add-test-d-in-d/src/docker-in-docker/install.sh#L984-L998) intentionally avoids `iptables-* -L` as a probe because the CLI `auto-modprobes` and would mask the very condition we're testing for. I can extend the same [kernel module check](https://github.com/Kaniska244/features/blob/add-test-d-in-d/src/docker-in-docker/install.sh#L988-L990) check to look for `iptable_nat` module for the legacy branch, but I'm hesitant to land it without a reproduction because I am not entirely sure about it's presence in the latest debian/ ubuntu based distros and if there should be a similar/ equivalent check for `iptables-nft` as well.  
Instead of using separate, rigid table modules like `iptables-legacy` uses `iptable_nat` sub-module, `iptables-nft` translates the old command syntax into modern Netfilter expressions. It automatically tells `nf_tables` to build a chain with a type nat attribute, [ref](https://ewpratten.com/blog/learning-nftables). 
Also, so far, I haven't able to come up with a simulation of the test case where `iptables` can be loaded but `iptables nat` isn't. In case if you have any advice on the reproduction of the similar problem you has faced in past, anything you could possibly recall, could help a great deal if we look further to safeguard the new implementation against the problem where `iptables nat` is not loading specifically.

.github/workflows/docker-in-docker-stress-test.yaml
Kaniska244: Hi @abdurriq 

It has been implemented. Kindly let me know in case of further concern.

test/docker-in-docker/docker_with_legacy_iptables.sh
Kaniska244: Hi @abdurriq

It has been implemented. Kindly let me know in case of further concern.

src/docker-in-docker/devcontainer-feature.json
Kaniska244: Hi @thediveo 

Implemented in this line.

abdurriq (approved):

PR #1671, opened 2026-06-09
[php] - Install xdebug from source if installation fails with PECL

Kaniska244:
## Summary

This PR updates the PHP feature’s Xdebug installation flow to make it more resilient when PECL installation fails. It keeps PECL as the first installation path, but falls back to building Xdebug from the official source tarball if PECL is unavailable, has stale channel metadata, or does not advertise compatibility with the selected PHP version.

## Changes

- Added `install_xdebug_from_source()` as a fallback installer for Xdebug.
- Runs `pecl channel-update pecl.php.net` before attempting to install Xdebug.
- Falls back to source-based Xdebug installation when `pecl install xdebug` fails.
- Restores `IFS` after parsing the PHP version to avoid leaking shell state.
- Bumps the PHP feature version from `1.1.4` to `1.1.5`.
- Updates PHP version proposals to include `8.5` / `8.5.0`.
- Updates PHP tests to cover PHP `8.5.0` as the default and `8.4.15` as an additional installed version.
- Enables Composer installation in the PHP test scenario.

## Testing

Updated the PHP feature test scenario and install checks to validate:

- PHP `8.5.0` is installed as the default version.
- PHP `8.4.15` is installed as an additional version.
- Composer is installed and available.

### DIFF
--- a/src/php/devcontainer-feature.json
+++ b/src/php/devcontainer-feature.json
@@ -1,6 +1,6 @@
 {
     "id": "php",
-    "version": "1.1.4",
+    "version": "1.1.5",
     "name": "PHP",
     "documentationURL": "https://github.com/devcontainers/features/tree/main/src/php",
     "options": {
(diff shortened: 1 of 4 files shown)


abdurriq (approved):

kisecrow9334 (approved):
Please.

PR #1675, opened 2026-06-24
Fix installing git-core PPA in Ubuntu 26.04

trxcllnt:
Apt in Ubuntu 26.04 is no longer compatible with armored keys, so this PR dearmors the git-core PPA key after it is fetched.


test/git/install_git_from_ppa_noble.sh
```diff
@@ -0,0 +1,17 @@
...
+# Definition specific tests
+check "version" git  --version
+
+cd /tmp && git clone https://github.com/devcontainers/feature-starter.git
+cd feature-starter
+check "perl" bash -c "git -c grep.patternType=perl grep -q 'a.+b'"
+
+# Report result
+reportResults
+
```
Kaniska244: Hi @trxcllnt 
Thank you for the contribution. If all the three new test scripts contain the exact same steps. Could you please symlink two of them one script?

Kaniska244 (changes requested):

test/git/install_git_from_ppa_noble.sh
trxcllnt: I followed the existing pattern to add these tests. Do you want me to do this for all the duplicated scenario tests for this feature, or just the new ones?

Kaniska244: Hi @trxcllnt 

Indeed the repository has lots of such duplicated test cases. I requested for the new scripts alone. Will take care of the existing duplicates in future.

trxcllnt: Done :+1:

Kaniska244 (approved):

abdurriq (approved):

PR #1679, opened 2026-06-30
[git] - Fixing issue for alpine installation from source and updating tests

Kaniska244:
## Summary

Fixes [CI failure](https://github.com/devcontainers/images/actions/runs/28441270640/job/84279702305?pr=1922#step:3:922). This PR fixes a build failure when installing Git from source on **Alpine** by adding the missing `linux-headers` dependency, and hardens the install script to **fail loudly** if the build doesn't succeed (instead of silently reporting "Done!"). It also overhauls the `git` feature's test matrix — adding a "version-is-latest" assertion, replacing the obsolete CentOS 7 test with Debian Bookworm/Trixie scenarios, and bumping the feature version to `1.3.7`.

## Key changes

**1. The core fix — `src/git/install.sh`**
- Adds `linux-headers` to the Alpine build dependencies (the missing package that caused the source build to fail).
- Adds `make` to the RHEL dependency list.
- Captures the `make` build result and exits with an error + message if the build fails, rather than always printing `Done!`:

```bash
make -s "${git_options[@]}" all && make -s "${git_options[@]}" install 2>&1
build_result=$?
rm -rf /tmp/git-${GIT_VERSION}
clean_up
if [ "${build_result}" -ne 0 ]; then
    echo "(!) Failed to build and install git ${GIT_VERSION}." >&2
    exit 1
fi
echo "Done!"
```

**2. Version bump — `src/git/devcontainer-feature.json`**
- `1.3.6` → `1.3.7`.

**3. Test improvements** (`install_git_from_src*.sh` for default, alpine, noble)
- Adds two helper functions: `get_latest_git_version` (queries the GitHub tags API) and `check_git_is_latest_version`, plus a new `version-is-latest` check to assert the installed Git matches the latest stable release.

**4. Test scenario matrix — `test/git/scenarios.json`**
- Removes the obsolete `install_git_from_src_centos-7` scenario (and deletes `install_git_from_src_centos-7.sh`).
- Adds `install_git_from_src_bookworm` (`debian:bookworm`) and `install_git_from_src_trixie` (`debian:trixie`) scenarios, with the new test files symlinking to `install_git_from_src.sh`.
- Changes the default scenario image from `ubuntu:noble` to `ubuntu:resolute`.

## Test Coverage

**New assertions added**
- A `version-is-latest` check is added to the default, **alpine**, and **noble** source-install tests. It resolves the latest stable Git tag from the GitHub API and asserts the installed binary matches:

```bash
# Resolves the latest stable git version from GitHub
get_latest_git_version() {
    curl -sSL -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/git/git/tags" \
        | grep -oP '"name":\s*"v\K[0-9]+\.[0-9]+\.[0-9]+(?=")' \
        | sort -rV \
        | head -n 1
}

# Verifies the installed git version matches the latest stable version on GitHub
check_git_is_latest_version() {
    local latest_version installed_version
    latest_version="$(get_latest_git_version)"
    installed_version="$(git --version | awk '{print $3}')"
    [ -n "$latest_version" ] && [ "$installed_version" = "$latest_version" ]
}

check "version-is-latest" check_git_is_latest_version
```

**Scenario / platform coverage changes**

| Scenario | Before | After | Change |
|---|---|---|---|
| `install_git_from_src` (default) | `ubuntu:noble` | `ubuntu:resolute` | Base image updated |
| `install_git_from_src_centos-7` | `centos:centos7` | — | ❌ Removed (EOL) |
| `install_git_from_src_bookworm` | — | `debian:bookworm` | ✅ Added |
| `install_git_from_src_trixie` | — | `debian:trixie` | ✅ Added |
| `install_git_from_src_alpine` | unchanged | unchanged | + `version-is-latest` check |
| `install_git_from_src_noble` | unchanged | unchanged | + `latest version` check |

**Coverage assessment**
- ✅ **Directly validates the fix path** — the Alpine source-build scenario now also asserts the resulting Git version, so a broken/incomplete Alpine build would surface.
- ✅ **Modernizes the Debian matrix** — Bookworm + Trixie replace EOL CentOS 7, keeping tested distros current.
- ⚠️ **Network-dependent** — `check_git_is_latest_version` makes an unauthenticated GitHub API call, which can hit rate limits in CI and cause flaky failures.


test/git/install_git_from_src.sh
```diff
@@ -5,8 +5,25 @@ set -e
 # Optional: Import test library
 source dev-container-features-test-lib
 
+# Resolves the latest stable git version from GitHub
+get_latest_git_version() {
```
abdurriq: Nit: Can we share these two functions across the scripts rather than duplicating it in all of them?

abdurriq (dismissed):

test/git/install_git_from_src.sh
Kaniska244: Hi @abdurriq 

Thank you for pointing this out. Implemented the same.

abdurriq (approved):

PR #1685, opened 2026-07-12
fix(rust): support Azure Linux base images

reynoldsbd:
Explicitly support Azure Linux base images in Rust feature.

Required two minor changes:

* Explicitly recognize distro ID `azurelinux` and map it to the same behavior as `mariner`
* Explicitly install `passwd`/`shadow-utils` package dependency for the `groupadd` command (which is not present by default in Azure Linux base images)

Also add a `rust_with_azurelinux` test scenario to validate expected behavior.

Changes validated by running `devcontainer features test -f rust` locally.

### DIFF
--- a/src/rust/NOTES.md
+++ b/src/rust/NOTES.md
@@ -2,8 +2,8 @@
 
 ## OS Support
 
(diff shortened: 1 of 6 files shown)


abdurriq (approved):

abdurriq:
LGTM, thanks for the contribution. There is some trailing whitespace, but this can be fixed post-merge.

PR #1705, opened 2026-08-06
feat(github-cli): support private extension installs

tylerkrop:
- Support installing private extensions with  `GH_TOKEN`/`GITHUB_TOKEN`
- Forward authentication to the selected non-root user and use `gh auth git-credential` for cloning
- Prevent temporary build authentication from incorrectly disabling the extension-list fallback

Testing

- Passed the existing public extension scenario
- Built with a private repository using a BuildKit secret and verified the token was not persisted in the image


tylerkrop:
@microsoft-github-policy-service agree company="GitHub"

Kaniska244 (approved):

abdurriq (approved):

PR #1708, opened 2026-08-11
feat: Add support for preserving pwsh history

nebula-it:


Kaniska244 (changes requested):
Hi @nebula-it 

Thank you for the contribution. The tests are failing to get sudo access while executing the onCreate script you have added for non dev container origin base images. 
This is happening because the lifecycle hooks such as `onCreateCommand` use the default container users to execute the commands and  the dev container origin base images have a default user `vscode` configured with sudo access when isn't the case for other base images. Would you kindly optimize the solution accordingly?

nebula-it:
Hey @Kaniska244 

Do you have an example `devcontainer.json` that I can use to verify that setup? This was working when I tested it locally with a `devcontainer.json` that just had powershell feature.

v-Kaniska244:
> Hey Kaniska (@Kaniska244)
> 
> Do you have an example `devcontainer.json` that I can use to verify that setup? This was working when I tested it locally with a `devcontainer.json` that just had powershell feature.

Hi @nebula-it 

The sample configuration is already present in `/workspaces/features/test/powershell/scenarios.json` file. The following test case can be used to reproduce this.
```
  "powershell_alma_linux": {
    "image": "almalinux:9",
    "features": {
      "powershell": {} 
    }
  }
```
If you have devcontainers/cli installed, you can use the following command for this with your changes.
```
devcontainer features test . -f powershell --filter "powershell_alma_linux" --skip-autogenerated --preserve-test-containers
``` 

Even better, you can run it with debian trixie(13) basic `debian:trixie` base image and it should reproduce the same error.
```
devcontainer features test . -f powershell -i "debian:trixie" --skip-scenarios --preserve-test-containers
```
Then if you try the dev container debian trixie(13) image `mcr.microsoft.com/devcontainers/base:trixie`, the error won't come and installation will be successful.
```
devcontainer features test . -f powershell -i "mcr.microsoft.com/devcontainers/base:trixie" --skip-scenarios --preserve-test-containers
```

nebula-it:
@Kaniska244 The two failures above seems to be unrelated reasons. Can you please double check

v-Kaniska244:
> Kaniska (@Kaniska244) The two failures above seems to be unrelated reasons. Can you please double check

Hi @nebula-it 

If you scan [the log](https://github.com/devcontainers/features/actions/runs/32793860694/job/97708072944?pr=1708), it's full of "permission denied" errors during test executions such as [this](https://github.com/devcontainers/features/actions/runs/32793860694/job/97708072944?pr=1708#step:5:309). Could you please check?

Kaniska244 (changes requested):
Replied on previous comment

nebula-it:
@Kaniska244 Fixed. 

Also fixed the profile test, which would have always passed no matter if the profile was loaded or not.

Kaniska244 (approved):

abdurriq (approved):

PR #1720, opened 2026-08-31
[github-cli] - Rotating the PGP key fingerprint

v-Kaniska244:
## Summary

- Updates the GitHub CLI repository signing key to the new full PGP fingerprint:
  `7F38 BBB5 9D06 4DBC B3D8 4D72 5612 B364 6231 3325`
- Exports the downloaded key as a portable OpenPGP keyring instead of storing it in GnuPG's keybox format.
- Adds APT-based installation coverage for Debian Trixie and Ubuntu Resolute.
- Bumps the Feature patch version from `1.1.1` to `1.1.2`.

## Background

GitHub CLI is rotating the PGP key used to sign its Linux package repositories. The current key expires on September 5. The APT repository is currently signed with both the old and new keys, but it will only be signed with the new key on or after that date.

The Feature's default installation path downloads packages directly from GitHub Releases and is unaffected. Users who set `installDirectlyFromGitHubRelease` to `false`, however, install `gh` through APT and need the new key in their local keyring. Without it, they may be unable to install or upgrade `gh` after the old key expires, even when a repository signature was created before the expiration date.

New key: https://keyserver.ubuntu.com/pks/lookup?search=7f38bbb59d064dbcb3d84d725612b36462313325&fingerprint=on&op=index

## Keyring compatibility

The existing implementation passed the APT keyring path directly to `gpg --recv-keys`. Modern GnuPG versions can create that file using the GnuPG-specific keybox format, which newer APT verification tooling in Debian Trixie and Ubuntu Resolute rejects as an unsupported file type.

This change receives the key into a temporary GnuPG home and then exports and dearmors it into a portable OpenPGP keyring. The resulting keyring is supported by both newer distributions and older Debian and Ubuntu releases.

## Testing

Added APT-based installation scenarios using `version: latest` and `installDirectlyFromGitHubRelease: false` for:

- Ubuntu Noble
- Ubuntu Resolute
- Debian Trixie

Validated with:

```text
devcontainer features test . -f github-cli --skip-autogenerated --preserve-test-containers
```

All tests passed.


babakks (commented):
Thanks for this, @Kaniska244! 🙌

abdurriq (approved):

PR #1732, opened 2026-09-09
improve(github-cli): retry gh extension install before falling back to git clone

rhiroe:
gh extension install failures fell through silently to the git-clone fallback. Retry it up to 3 times with backoff before falling back, since failures are most often transient (network/API rate limiting) rather than the extension being fundamentally unusable, and log each failed attempt's exit code so provisioning logs show why the fallback was hit.

Extensions that require a build step (e.g. github/gh-stack) are not usable after a plain git clone. Improving the fallback itself (e.g. downloading a prebuilt release binary) would add significant logic and ongoing maintenance cost, so that path is not pursued here; instead, after cloning, check for the expected gh- executable and exit 1 if it's missing, so a build-requiring extension fails provisioning loudly instead of leaving a broken extension in place.


v-Kaniska244 (changes requested):
Hi @rhiroe 

Would you please investigate the test failure?

rhiroe:
Apparently, as of gh v2.101.0, GitHub Copilot is a built-in command (gh copilot) instead of an extension, so gh extension install github/gh-copilot fails.

rhiroe:
I created a pull request: https://github.com/devcontainers/features/pull/1741

PR #1733, opened 2026-09-09
fix(github-cli): fix ensure_gh_extension_list_wrapper always creating gh wrapper

rhiroe:
The check stripped GH_TOKEN along with the other auth env vars before running `gh extension list`, but gh CLI requires some token to be present just to run the extension subcommands, regardless of whether listing local extensions actually needs auth. With no token and a fresh empty GH_CONFIG_DIR, the check could never succeed, so the /usr/local/bin/gh wrapper was installed unconditionally instead of only when real gh actually fails to see the extensions.


v-Kaniska244 (changes requested):
Hi @rhiroe 

Thank you for the contribution. Would you please bump the feature version for this change.

PR #1734, opened 2026-09-09
feat(github-cli): support pinning extensions to a tag via owner/repo:tag

rhiroe:
Allows `extensions` entries like `github/gh-stack:v1.0.0` to install a pinned version, passing `--pin` to `gh extension install` and `--branch` to the git clone fallback.


v-Kaniska244 (changes requested):
Hi @rhiroe 

Thank you for the contribution. Would you please bump the feature version as well.

Issue #890, opened 2024-03-03
PowerShell Feature - Support for Alpine


ili101:
Appreciate it if the PowerShell feature supported Alpine.

```
 > [dev_containers_target_stage 6/6] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=powershell_1,target=/tmp/build-features-src/powershell_1     cp -ar /tmp/build-features-src/powershell_1 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/powershell_1  && cd /tmp/dev-container-features/powershell_1  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/powershell_1:
0.341 ===========================================================================
0.341 Feature       : PowerShell
0.341 Description   : Installs PowerShell along with needed dependencies. Useful for base Dockerfiles that often are missing required install dependencies like gpg.
0.341 Id            : ghcr.io/devcontainers/features/powershell
0.341 Version       : 1.3.2
0.341 Documentation : https://github.com/devcontainers/features/tree/main/src/powershell
0.341 Options       :
0.341     VERSION="latest"
0.341     MODULES=""
0.341     POWERSHELLPROFILEURL_=""
0.341 ===========================================================================
0.347 ./install.sh: line 144: dpkg: command not found
0.347 ERROR: Feature "PowerShell" (ghcr.io/devcontainers/features/powershell) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/powershell for help troubleshooting this error.
------
Dockerfile.extended:33
--------------------
  32 |     
  33 | >>> RUN --mount=type=bind,from=dev_containers_feature_content_source,source=powershell_1,target=/tmp/build-features-src/powershell_1 \
  34 | >>>     cp -ar /tmp/build-features-src/powershell_1 /tmp/dev-container-features \
  35 | >>>  && chmod -R 0755 /tmp/dev-container-features/powershell_1 \
  36 | >>>  && cd /tmp/dev-container-features/powershell_1 \
  37 | >>>  && chmod +x ./devcontainer-features-install.sh \
  38 | >>>  && ./devcontainer-features-install.sh \
  39 | >>>  && rm -rf /tmp/dev-container-features/powershell_1
  40 |     
--------------------
[...]

Issue #824, opened 2024-01-30
Git Feature - Add OS support for Alpine


LavissaWoW:
#782 added distro checks that excludes the Alpine distro. This has effectively broken building images with git for Alpine.

```
------
 > [dev_containers_target_stage 4/4] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=git_0,target=/tmp/build-features-src/git_0     cp -ar /tmp/build-features-src/git_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/git_0  && cd /tmp/dev-container-features/git_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/git_0:
0.310 ===========================================================================
[2024-01-30T21:00:27.731Z] 
0.310 Feature       : Git (from source)
0.310 Description   : Install an up-to-date version of Git, built from source as needed. Useful for when you want the latest and greatest features. Auto-detects latest stable version and installs needed dependencies.
[2024-01-30T21:00:27.731Z] 0.310 Id            : git
0.310 Version       : 1.2.0
0.310 Documentation : https://github.com/devcontainers/features/tree/main/src/git
0.310 Options       :
0.310     VERSION="os-provided"
0.310     PPA="true"
0.310 ===========================================================================
0.316 Linux distro alpine not supported.
[2024-01-30T21:00:27.732Z] 0.316 ERROR: Feature "Git (from source)" (git) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/git for help troubleshooting this error.
------
```

samruddhikhandale:
Hi 👋 

According to https://github.com/devcontainers/features/tree/main/src/git#os-support, we officially never added support for `Alpine` to the `Git` Feature.

@LavissaWoW Does `v1.1.6` of the Git Feature with the `alpine` image work fine for you? If so, we can start officially supporting it.

Feel free to contribute a PR if needed, we appreciate community support. Thanks!

jdputschadi:
Because I authored the offending PR, I wanted to do some due diligence about the `git` feature version `1.1.6` (pre-PR [#782](https://github.com/devcontainers/features/pull/782) ) and trying to use it with an Alpine base image...

With respect to the `git` feature at version `1.1.6` and Alpine images:

## Using `devcontainer features test` to test with an `alpine:latest` base image:

```
[...]

Issue #751, opened 2023-11-13
DinD and DoutD does not support Alpine Base Images


sr229:
Official devcontainer features should be at least be able to support their own base images, however, in the case for Docker in Docker and Docker outside Docker, and a couple of few more features, they do not support the Alpine image.

joshspicer:
Thanks for the feedback.  I agree that would be ideal. With the engineering capacity we have on the maintainers team, we're only able to officially support a subset of what we'd like to - in this case it's that our Features are compatible with debian-like distros.  Features like `common-utils` we maintain for multiple distros due to its convenience.

I'd encourage members of the community that would find this useful to [build their own Features and share them with the community](https://containers.dev/guide/author-a-feature) (and post here for visibility if folks search in the future!).

samruddhikhandale:
+1 to Josh's comment above.

Also, if you are interested, feel free to contribute a PR for `docker-in-docker` and `docker-outside-of-docker` Features with added support for `Alpine`. We'd appreciate that, and happy to help in the process. Thank you!

cirolosapio:
i've created a [list](https://github.com/cirolosapio/devcontainers-features) of alpine features
maybe it could help

sr229:
> +1 to Josh's comment above.
> 
> Also, if you are interested, feel free to contribute a PR for `docker-in-docker` and `docker-outside-of-docker` Features with added support for `Alpine`. We'd appreciate that, and happy to help in the process. Thank you!

I'd be happy to help if you can guide me to the right direction, I'll give it a gander and see what I can fix!

Issue #498, opened 2023-03-21
python feature not compatible with alpine - better error message


hediet:
This is my devcontainer.json:
```json
{
	"name": "Alpine",
	"image": "mcr.microsoft.com/devcontainers/base:alpine-3.17",
	"features": {
		"jupyterlab": "latest"
	}
}
```

Opening the dev container produces the following error:

```
#0 0.453 Feature       : Python
#0 0.453 Description   : Installs the provided version of Python, as well as PIP
X, and other common Python utilities.  JupyterLab is conditionally installed wit
h the python feature. Note: May require source code compilation.
#0 0.453 Id            : ghcr.io/devcontainers/features/python
#0 0.453 Version       : 1.0.22
#0 0.453 Documentation : https://github.com/devcontainers/features/tree/main/src
/python
#0 0.453 Options       :
#0 0.453     VERSION="os-provided"
#0 0.453     INSTALLTOOLS="true"
#0 0.453     OPTIMIZE="false"
#0 0.453     INSTALLPATH="/usr/local/python"
#0 0.453     INSTALLJUPYTERLAB="true"
#0 0.453     CONFIGUREJUPYTERLABALLOWORIGIN=""
#0 0.453 =======================================================================
====
#0 0.461 find: /var/lib/apt/lists/*: No such file or directory
#0 0.461 Running apt-get update...
#0 0.462 ./install.sh: line 190: apt-get: command not found
#0 0.462 ERROR: Feature "Python" (ghcr.io/devcontainers/features/python) failed
to install! Look at the documentation at https://github.com/devcontainers/featur
es/tree/main/src/python for help troubleshooting this error.
------
ERROR: failed to solve: executor failed running [/bin/sh -c cp -ar /tmp/build-fe
atures-src/python_1 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-conta
iner-features/python_1  && cd /tmp/dev-container-features/python_1  && chmod +x
./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm
 -rf /tmp/dev-container-features/python_1]: exit code: 127
[...]

PR #1739, opened 2026-09-16
feat/github-cli: support Alpine based images

the-wondersmith:
PR adds support for using the `github-cli` feature with Alpine based images. 

> [!NOTE]
> PR *technically* implements an extensible pattern for implementing support for *other* currently unsupported linux flavors. That's more of a side effect / happy accident than actual intention. That said, if the `devcontainers/features` maintainers don't have objections, I will happily extend it to the other single-distro (read: debian) features if this PR is accepted. 😅


the-wondersmith:

the-wondersmith:
As noted in the PR description, if this is accepted I'm happy to tackle:
- #498 
- #751 
- #890 
- #824 

and generally adding Alpine support across the board (incrementally).

PR #1741, opened 2026-09-18
fix(github-cli): remove github/gh-copilot from extension install test

rhiroe:
gh v2.101.0 bundles Copilot as a built-in command, so `gh extension install github/gh-copilot` now always fails with
 "copilot" matches the name of a built-in command or alias.

https://github.com/devcontainers/features/actions/runs/35105341774/job/104825172213


v-Kaniska244 (approved):

